PR #23, opened 2015-11-20
Matt poc

mattstratton:


mattstratton:
The map still doesn't quite center properly. See #24

PR #61, opened 2015-11-29
Add legacy template and port stuff

m1keil:
This needs to be more pretty (a lot of duplication is going on)
port all of the blogs. blogs needs to be checked that the links are the same. Also fix broken images in blog posts.

css/js/images needs to be cleaned up.

events still missing.


mattstratton:
I don't quite understand what this is for. We cannot run more than one theme in Hugo at a time. Can you describe a bit more of what the approach is for this?

m1keil:
I've created extra theme in order to avoid any merge conflicts, or any current work we have on the "devopsdays" site.

As we discussed previously, it would be much easier for us to merge the old site to use Hugo, and only then start making edits to the layout.

The devops-legacy theme does most of it (not including /events, since we want to use data files there).
If you run hugo and instruct it to use this theme, you should get the exact replica of today's site, only instead of webby, it's powered by hugo.
I've ported the blog posts and did some minimal pagination stuff under the /blog section. 

Tell me what you think.

mattstratton:
Where does all the event data come from? I don't see any of the old events?

m1keil:
yep, I didn't ported any events yet.
Also the future and past events in the footer are still not auto generated but static. Work in progress.

m1keil:
Just let me know if I should invest time in this

mattstratton:
My only concern with this is that we will import all the pages over, and then have to touch them all with the new layout. I think the layout is actually almost done; the thing that is the holdup right now is there is no point in importing the events until that is done since the data file format is not finalized right now.

I would suggest we do the following:

1. Finalize the event data file format
2. Finalize the event detail page template (this is actually pretty much done I think)
3. Finalize the sponsor data file format and integration (I think this is done, just need to validate it)
4. Import all the events to the new structure
5. Import all the existing pages
6. THEN work on making any changes to the theme. I think that the existing theme should work for importing stuff now?

In this approach, we should make NO more stylistic changes to layout, etc...just get the data over so it all displays. Then we can make changes to the templates which will be much more of a DRY approach.

m1keil:
But the current layout is entirely different from the site we have at devopsdays.org so I guess I'm not really understanding what is the goal.

mattstratton:
Which parts are completely different? The menu nav? I pretty much replicated how it looked when I built this theme. The only real difference is in the menu bar/social links. The page layouts can/should be the same, and those are just templates anyway...my concern is building a structure with partials that doesn't reflect the overall?

I might be missing something so if you can be specific about what is fundamentally different in the new theme it will help me a lot :)

m1keil:
Well, not only. Right now the front page is different in many things (alignment, sizes, missing elements). I thought it would be easier just to take what we currently have and convert it to use Hugo, instead of building everything from scratch and trying to make it look identical.

Did you run the legacy theme?

mattstratton:
I did. I pulled it down and ran it. I guess as long as we feel like any content imported using the "legacy" theme wouldn't require touching later, then it makes sense. What I don't want to do is important dozens of events, and then have to touch their files again to make them work with the final theme. The layout stuff you've mentioned seems to be more chrome than functional, and I don't think it affects the import of data one way or another. 

I see the point of having two themes so you can switch between them as we go. The biggest thing is to make sure they are both compatible with the data/content and the content files don't need to be changed for the new setup. 

The only thing I would suggest is to not over-write the content/page/organizing.md file...I put a ton of content in it which will get blown away by this merge (see #52 for reference)

Otherwise I think it is good :)

m1keil:
From my checks the `organizing.md` is still there, and working.

mattstratton:
I probably just panicked from the github diff. Then I say plus one on the merge. I'll fire it up. It won't hurt anything :)

PR #78, opened 2016-03-07
Refactor for event path

mattstratton:


mattstratton:
This was a particularly large refactor - event data files now can (and should) have spaces and match to the folder of the event.

Also, it now determines the event data file based upon the filepath of the content file. Sorta.

PR #89, opened 2016-03-31
Migrate Minneapolis 2016

tduffield:
Here's an initial migration for minneapolis. The logos are borderless (which can easily be changed).


bridgetkromhout:
LGTM.

PR #100, opened 2016-04-11
SIN-16 Update event details

ctrabold:


data/events/2016-singapore.yml
```diff
@@ -5,29 +5,27 @@ friendly: "2016-singapore" # Four digit year and the city name in lower-case. Do
...
+# List of pages you want to show up in the navigation of your page. Remove any that aren't relevent yet.
+navigationelements: [
+  "welcome",
+  "location",
+  "propose",
+  "sponsor",
+  "contact",
+  "conduct"
+]
+team: ["Christian Trabold", "Satyam", "Maris"]
```
bridgetkromhout: Do they use just one name?

Issue #108, opened 2016-04-11
Sponsor logos are two wide instead of three


kmugrage:
It makes for a pretty long list.

bridgetkromhout:
Agreed that three-wide looks better.

bridgetkromhout:
With the fixed merged, do we want this issue so we revisit it for fix-optimizing later? /cc @mattstratton

mattstratton:
I will tag this for the "future enhancements" milestone and move it out of "ready" state. This will likely get revisited when we redesign the UX anyway.

bridgetkromhout:
This goes away if we switch to a sponsor footer (https://github.com/devopsdays/devopsdays-web/issues/295).

PR #109, opened 2016-04-11
reduced total logo width from 100px to 95px

kmugrage:
For #108 

I'm not sure this is the "right" way to do this compared to changing the width of the div, but reducing the total size of the logos by 5px makes them 3 wide on desktop browsers.


bridgetkromhout:
I am going to defer this one to @mattstratton. Agree with the notion, unsure how he'd prefer it implemented.

mattstratton:
This seems like the less elegant manner of doing it, but it also works, which is better than debating the CSS. I think it's fine for now. I'll accept it.

PR #111, opened 2016-04-11
MSP program updates.

bridgetkromhout:
Added speaker info and updated program grid.


themes/devopsdays-legacy/layouts/talk/single.html
```diff
@@ -0,0 +1,80 @@
...
+ <!-- speaker code begin -->
+
+      {{ $speakers := $e.speakers }}
+        {{ range $e.speaker_levels }}
+          {{ range where $speakers "level" .id }}
+          {{ $filename := (print .id ".md") }}
+            {{if eq $filename (index $path 3)}}
+              <h2>Speaker</h2>
+              {{ $s :=  (index $.Site.Data.events.minneapolis2016.speakers .id) }}
+              {{ if isset $.Site.Data.events.minneapolis2016.speakers .id }}
```
bridgetkromhout: Here's where I'm running into trouble. I don't seem to be able to construct a version of this variable that can contain a variable itself. (It also barfs if the "minneapolis2016" segment starts with a number or contains a hyphen, but that's a lesser concern.)

In order to make this truly general-purpose, I need to be able to make this generic.

themes/devopsdays-legacy/layouts/talk/single.html
```diff
@@ -0,0 +1,80 @@
...
+
+      {{ .Content }}
+
+ <!-- speaker code begin -->
+
+      {{ $speakers := $e.speakers }}
+        {{ range $e.speaker_levels }}
+          {{ range where $speakers "level" .id }}
+          {{ $filename := (print .id ".md") }}
+            {{if eq $filename (index $path 3)}}
```
bridgetkromhout: This is the trick to make sure that for a given talk I display only the bio I want. Checking the "id" against the name of the file is exceedingly brittle, but when I try checking against `.Title`, that's out of scope.

themes/devopsdays-legacy/layouts/talk/single.html
bridgetkromhout: Obviously we shouldn't merge this as-is, but I wanted @mattstratton to see where I'm at.

bridgetkromhout:
Going to handle this from a branch locally.

PR #122, opened 2016-04-13
Added offsite nav links. Fixes #76

### LINKED ISSUE #76: Some events want the menu to link outside
Again, see DC as an example. Should be able to override the destination of menu items such as "register"

mattstratton:
This fundamentally changes the structure of the navigation links in the event configuration file. See the sample data file for an example. The "old" style data file is still functional for backwards compatibility.


mattstratton:
Please note that the URL set *must* have the proper "http(s)://" in front of it, or it won't work. We should probably add some error handling at some point for this.

PR #130, opened 2016-04-13
Added event_map short code

jyee:
We've already got the event lat+lon and a location name. I figured it would be nice if we used these to generate a map on each event's location page via a shortcode.

Organizers will just need to ensure their lat+lon and location (venue name) are correct in the YAML, then uncomment the event_map shortcode in location.md.

### DIFF
--- a/content/events/sample-event/location.md
+++ b/content/events/sample-event/location.md
@@ -5,7 +5,7 @@ type = "event"
 draft = true
 
 +++
-<h4>
 
 Information about the venue including address, map/direction, parking/transit, and any hotel group discount codes.
-<p>
+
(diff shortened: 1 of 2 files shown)


mattstratton:
I don't see the example use of the short code?

jyee:
`{{< event_map >}}`
It's [commented out ](https://github.com/devopsdays/devopsdays-web/pull/130/files#diff-c747bc9b052ef485b45c5cd933da810eR11)in `content/events/sample-event/location.md`

mattstratton:
Oh, it was hard to see in the mobile app I was using to review the PR :)

This looks rad! Good improvement.

Issue #75, opened 2016-03-07
Support adding more info for organizers


mattstratton:
Organizers should be able to add twitter handles, employer, and email address (all optional)

See DC for an example

bridgetkromhout:
I would prefer if we don't encourage email addresses. I don't want people emailing a specific organizer instead of the group list. That leads to dropped threads. (Especially if that organizer is me!)

mattstratton:
Yup. To be fair, I wrote this issue at 2 AM. I agree that the use of email addresses should be dropped from this feature :)

bridgetkromhout:
This is done, as far as I can tell.

PR #150, opened 2016-04-15
Support adding more info for organizers

jyee:
- New YAML key for "team_members".
- Supports name (required), twitter (optional, with or without @),
  and employer (optional).
- Maintains compatibility with simpler "team" list method.

See issue #75


bridgetkromhout:
👍

PR #181, opened 2016-05-02
Remove year from the past events.

bridgetkromhout:
The year in the past events looks terrible and makes it wrap a lot. Presumably we could replace it if we need it in a re-design, but for now it would be nice for it to not be cluttering a space we're working in.


mattstratton:
shipit.

PR #192, opened 2016-05-03
New logos

kmugrage:


mattstratton:
@kmugrage - we are starting to use the new sponsor file/logo guidelines in the README where they should be named YYYY-SPONSORNAME. This is fine for now, but anythign new, please use the year prefix.

PR #199, opened 2016-05-03
Ohio 2015 logos

kjenkins19:
All the logos involved for the 2015 Ohio.
If the image existed in the website already, but did not have a 2016 reference, I created the 2016 version.
If the image did not exist, I updated the image, and gave it a 2015 reference.


bridgetkromhout:
:thumbsup:

PR #218, opened 2016-05-05
[2016-chicago] first pass at migrating webby data

jerrycattell:
added sponsors and updated existing sponsor logos to be 200x200

still need to use config in a number of places instead of hard-coding values


mattstratton:
Well dang it, I guess I can't use Chicago as my experimental data anymore...:)

Issue #217, opened 2016-05-05
specify event logo size and dimensions


mattstratton:
I don't know that we want to be explicit about them, but we should set a max width and then wrap the img resource with it.

This all gets better once we move to responsive/bootstrap.

bridgetkromhout:
Okay. I think if we're recommending
```
 <div style="text-align:center;">
  {{< event_logo >}}
</div>
```

and giving no further guidance than that, then probably the default behavior needs to be non-giant size.

Example of badness:

![screen shot 2016-05-05 at 6 19 43 am](https://cloud.githubusercontent.com/assets/2104453/15042674/ae603356-128b-11e6-8750-1e313a6c4c11.png)

How's this band-aid?

mattstratton:
resolved in #220

PR #219, opened 2016-05-05
Keep logos from being giant.

bridgetkromhout:
Possible temporary band-aid for https://github.com/devopsdays/devopsdays-web/issues/217.


mattstratton:
Seems to work for now. I say :shipit:

PR #224, opened 2016-05-05
New proposal for Oslo 2016

steinim:


bridgetkromhout:
Can you adjust this PR so that there aren't a ton of unrelated commits in here? Thanks!

PR #229, opened 2016-05-05
Added concur logo at silver

kmugrage:


data/sponsors/2016-concur.yml
```diff
@@ -0,0 +1,2 @@
+name: "Concur"
+url: "http://www.concur.com"
```
bridgetkromhout: No newline - what kind of editor are you using, man?! :)

kmugrage: Dreamweaver! (It's actually sublime. It's the day for unexplained stuff I guess)

PR #232, opened 2016-05-05
Corrections to initial Nashville site

bridgetkromhout:
We require a code of conduct. You can modify it, but you must have one.

There are also some slight fiddly bits that made this not show up correctly.

FYI:
@jcockhren 
@jfryman


jfryman:
Ah. This part was unclear to me... whether or not the hugo default pages would be automatically rendered or if we had to be explicit.

Thanks. We may end up changing it, but I appreciate the help with cleanup. :bow:

PR #255, opened 2016-05-06
Refactor future footer

### LINKED ISSUE #233: Refactor Future sidebar
The Upcoming Events box above the map and the Future Events sidebar do not line up for events that do not have dates attached to them yet.

For example:  http://devopsdays.github.io/events/2016-madison/ does not have dates attached, and shows up in the grey box above the map, but not in the Future Events sidebar.

image
vs. 
image

### LINKED ISSUE #246: past events with conflicting dates don't all display on events page

### LINKED ISSUE #247: different behavior of "year" key for future and past events on events page
If the event is "past" then it needs to be in quotes, if it's "current" it needs to not be in quotes. This is a problem :)

mattstratton:
It's somewhat sloppy, and has the year 2016 hardcoded into it for now, but it works. This removes the issue with the "events on the same date" (only for future footer; the issue still exists for past events for now) and it also takes care of the "year in quotes" issue.

Resolves #247
Resolves #246
Resolves #233

### DIFF
--- a/data/events/2016-ohio.yml
+++ b/data/events/2016-ohio.yml
@@ -1,5 +1,5 @@
 name: "2016-ohio" # The name of the event. Four digit year with the city name in lower-case, with no spaces.
-year: "2016" # The year of the event. Make sure it is in quotes.
+year: 2016 # The year of the event. Make sure it is in quotes.
 city: "Ohio" # The city name of the event. Capitalize it.
 friendly: "2016-ohio" # Four digit year and the city name in lower-case. Don't forget the dash!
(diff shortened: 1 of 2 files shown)


mattstratton:
@bridgetkromhout - can you take a look at this and double-test it for me to make sure it does fix the issues? I think I got it handled.

bridgetkromhout:
This works to band-aid the listed issues, hackish as it may be. It doesn't display 2017 events well (it places them right after the last of the 2016 events with dates and doesn't display the year) but that's something we can deal with later.

PR #269, opened 2016-05-07
Migrate 2016-philly w/ updates.

peterjshan:
This should be everything we need for the migration.

cc: @bridgetkromhout


bridgetkromhout:
👍

PR #288, opened 2016-05-07
Charlotte - static

bridgetkromhout:
Hey, @mattstratton - 

Would this trigger the problems you saw with static stuff earlier? Not positive I fully understand them.

This is my attempt at an MVP to get an old site back online: 

* a minimal data yaml file that just has the date stuff necessary to make the event show up.
* An import/modify of the index.html into the welcome.md (I couldn't easily see how to just redirect)
* everything else is static.

Drawbacks: sponsors missing on front page, possible build issues, too many logo files eventually, will need to revisit, slightly manual
Upsides: gets the site back up much faster, preserves all links

What do you think? If you think this is okay I'd do more like this.


bridgetkromhout:
Also, I still would like to actually migrate everything. It's just too slow, imo, though. I'd like stuff to be not-missing as soon as possible.

mattstratton:
I think this should be safe. I had toyed with this idea myself and never did it. The downside is we will end up with a lot of sponsor image duplication but I think we can deal with it.

Issue #93, opened 2016-04-11
Ability to control "Become a <level> sponsor" links 


WhatsChrisDoing:
Extend the sponsor_levels array to include the number of sponsor slots at each level so that the "Become a <level> sponsor" can be automatically excluded for levels that are full.

bridgetkromhout:
Thanks for working on this, @yvovandoorn. I think this might be best configured like this:

sponsor_levels:
    - id: platinum
      label: Platinum
      max: 3 # since Minneapolis already has 3, this would automatically remove the "become a..."
    - id: gold
      label: Gold
      max: 12
    - id: silver
      label: Silver
      number:   # left blank which makes the number unlimited

And here's an optional but related possibility:

    - id: host
      label: Host
      number: 1 # If one host sponsor is allowed, change "Host Sponsors" to instead display as "Host Sponsor".

Your option for turning off all "become a sponsor" messages on a city-wide basis is fine, but I suspect that the main uses of that would be "before the prospectus is done" and "when the conference is almost happening". So, an event-wide override like this might work:

sponsors_accepted: "yes"

I think it's fine for it to be a required field and could default to "no". Then people could turn it on when they're ready.

bridgetkromhout:
Moving the discussion from https://github.com/devopsdays/devopsdays-web/pull/294#issuecomment-217731362, @yvovandoorn says:

> The biggest challenge I've been having with Hugo (and thus my limits) is how the logic around how to compare sponsor.level vs sponsor_levels.max works.

> Given this is my 4th day with Hugo (or Go in general) so far I've come this far:
```
      {{ else if ge var .max }}
         <div class = "sponsor-cta">
          <br>
       </div>
```

> My struggles:

> Is how to properly add up the levels set under sponsors.level against sponsor_levels.max.

> Example

> I have 4 sponsors:
```
 sponsors:
  - id: solvinity
    level: gold
  - id: stackstate
    level: gold
  - id: thoughtworks-products
    level: gold
 sponsor_levels:
  - id: gold
    label: Gold
    max: 2
```

> How do I count up the id 'gold' being used 3x under sponsors to fire off against (greater than or equal to) under sponsor_levels.max:2. Feels like there is a loop in involved and the sysadmin in me to use unix commands to determine the comparison. Obviously not appropriate here.

bridgetkromhout:
@pmoosh and @ksatirli are also probably interested in this.

bridgetkromhout:
[...]

PR #294, opened 2016-05-08
Ability to remove sponsor link form "Become a XXX Sponsor!"

yvovandoorn:
Introduces a new option: `sponsor_link_ask' to have the ability to turn off the "Become a XXX Sponsor!" feature. 

As this is an "opt_in" feature, as "it doesn't change default behavior", only those events that want to use the option will see its benefit.

### DIFF
--- a/data/events/2016-amsterdam.yml
+++ b/data/events/2016-amsterdam.yml
@@ -5,7 +5,7 @@ friendly: "2016-amsterdam" # Four digit year and the city name in lower-case. Do
 
 # Datus and status
 status: "current" # Options are "past" or "current". Use "current" for upcoming.
-sponsor_status: "closed"
+sponsor_link_ask: "closed"
 startdate: 2016-06-29  # The start date of your event. Leave blank if you don't have a venue reserved yet.
 enddate: 2016-07-01  # The end date of your event. Leave blank if you don't have a venue reserved yet.
 cfp_date_start: 2016-01-15 # start accepting talk proposals.
--- a/data/events/2016-amsterdam.yml
+++ b/data/events/2016-amsterdam.yml
@@ -22,7 +22,7 @@ team: ["Harm Boertien", "Thomas Krag", "Kerim Satirli", "Arjan Wolfs", "Wilco No
 organizer_email: "organizers-amsterdam-2016@devopsdays.org" # Put your organizer email address here
 proposal_email: "organizers-amsterdam-2016@devopsdays.org" # Put your proposal email address here
 
-sponsor_link_ask: "closed"
+sponsors_accepted: "yes"
 sponsors:
   # gold
   - id: chef
--- a/data/events/2016-amsterdam.yml
+++ b/data/events/2016-amsterdam.yml
@@ -22,7 +22,7 @@ team: ["Harm Boertien", "Thomas Krag", "Kerim Satirli", "Arjan Wolfs", "Wilco No
 organizer_email: "organizers-amsterdam-2016@devopsdays.org" # Put your organizer email address here
 proposal_email: "organizers-amsterdam-2016@devopsdays.org" # Put your proposal email address here
 
-sponsors_accepted: "yes"
+sponsors_accepted: "no"
 sponsors:
   # gold
   - id: chef
@@ -69,16 +69,22 @@ sponsors:
 sponsor_levels:
   - id: bbq
     label: BBQ
+    max: 1
   - id: beer
     label: Beer
+    max: 1
   - id: gold
     label: Gold
+    max: 8
   - id: silver
     label: Silver
+    max: 4
   - id: bronze
     label: Bronze
+    max: 4
   - id: community
     label: Community
+
 
 nav_elements: # List of pages you want to show up in the navigation of your page.
   - name: welcome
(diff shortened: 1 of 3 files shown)


yvovandoorn:
@pmoosh and @ksatirli - open for feedback on it.

data/events/2016-amsterdam.yml
```diff
@@ -22,7 +22,7 @@ team: ["Harm Boertien", "Thomas Krag", "Kerim Satirli", "Arjan Wolfs", "Wilco No
 organizer_email: "organizers-amsterdam-2016@devopsdays.org" # Put your organizer email address here
 proposal_email: "organizers-amsterdam-2016@devopsdays.org" # Put your proposal email address here
 
-# Sponsors 2016-sendachi
```
yvovandoorn: Removed this stray comment from the amsterdam event data as they aren't a sponsor and it was a remnant of another yml file I looked at for inspiration.

bridgetkromhout:
Thanks! It's common to sell out of one type of sponsorship and still be signing others up. This would ideally be an optional flag on each specific sponsorship level. See also https://github.com/devopsdays/devopsdays-web/issues/93.

yvovandoorn:
Let me take a stab at it on my plane ride in the coming hours!

themes/devopsdays-legacy/layouts/partials/event_footer.html
```diff
@@ -19,7 +19,7 @@
       {{ end }}
     {{ end }}
     {{ if eq $e.status "current" }}
-      {{ if eq $e.sponsor_link_ask "closed" }}
+      {{ if eq $e.sponsors_accepted "closed" }}
```
bridgetkromhout: I think this will be easier to understand if instead of "closed"/"open" it's "yes"/"no" (as seen in the yaml file above).

yvovandoorn: Yup. You're right. I missed that in my branch.

On Sunday, 8 May 2016, Bridget Kromhout <notifications@github.com> wrote:

> In themes/devopsdays-legacy/layouts/partials/event_footer.html
> <https://github.com/devopsdays/devopsdays-web/pull/294#discussion_r62431741>
> :
>
> > @@ -19,7 +19,7 @@
> >        {{ end }}
> >      {{ end }}
> >      {{ if eq $e.status "current" }}
> > -      {{ if eq $e.sponsor_link_ask "closed" }}
> > +      {{ if eq $e.sponsors_accepted "closed" }}
>
> I think this will be easier to understand if instead of "closed"/"open"
> it's "yes"/"no" (as seen in the yaml file above).
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/devopsdays/devopsdays-web/pull/294/files/298fb56f8b9e7ebfa4a7daf67db5b9017f0f5228..19b2cba09de8fbaf11a50fb0c68cca5594eb5ac3#r62431741>
>


-- 

*Yvo van Doorn* — Manager, Technical Services (EMEA)

+31 6 38 74 07 46 <+316740746> – yvo@chef.io – *my: *Linkedin
<http://www.linkedin.com/in/yvovandoorn/>  Twitter
<https://twitter.com/yvov>

CHEF

CHEF.IO <http://www.chef.io/>

TM

*chef.io <http://www.chef.io/>*   Blog <http://www.chef.io/blog/>   Facebook
<https://www.facebook.com/getchefdotcom>   Twitter
<https://twitter.com/chef>   Youtube <https://www.youtube.com/getchef>

yyyy-city.yml
```diff
@@ -47,7 +47,7 @@ sponsors:
   - id: arresteddevops
     level: community
 
-sponsor_link_ask : open # Whether you want "Become a XXX Sponsor!" link
+sponsors_accepted : open # Whether you want "Become a XXX Sponsor!" link
```
bridgetkromhout: We're quoting most strings in this file, I think? (Also, let's make it "yes".)

yvovandoorn:
airport commits merging to branch.

struggling with how to do the logic on when not present link when `max` is met.

yvovandoorn:
@bridgetkromhout 

Ok so the logic / idea behind turning it off city/event wide makes sense and I've updated the PR as such. 

The biggest challenge I've been having with Hugo (and thus my limits) is how the logic around `max` works.

Given this is my 4th day with Hugo (or Go in general) so far I've come this far:
`
      {{ else if ge var .max }}
        <div class = "sponsor-cta">
          <br>
        </div>
`

My struggles:

Is how to properly add up the levels set under sponsors.level against sponsor_levels.max. 

Example

I have 4 sponsors:
`sponsors:
  - id: solvinity
    level: gold
  - id: stackstate
    level: gold
  - id: thoughtworks-products
    level: gold
`

and in 
`sponsor_levels:
  - id: gold
    label: Gold
    max: 2`

How do I count up the id 'gold' being used 3x under sponsors to fire off against (greater than or equal to) under sponsor_levels.max:2. Feels like there is a loop in involved and the sysadmin in me to use unix commands to determine the comparison. Obviously not appropriate here.

bridgetkromhout:
This PR looks great, @yvovandoorn - thanks! I'll move the discussion of the additional work into https://github.com/devopsdays/devopsdays-web/issues/93.

PR #308, opened 2016-05-10
devopsdays AMS: Add devoteam as Bronze Sponsor

yvovandoorn:
Agreement signed. Adding sponsor to website.


bridgetkromhout:
You're being sponsored by https://en.wikipedia.org/wiki/Devo? :)

PR #312, opened 2016-05-10
2016-boston updates

masteinhauser:
Finally adding sponsors to boston.

Massive thanks to @bridgetkromhout for actually migrating this over from the webby based site while I was running around putting out tire fires!


bridgetkromhout:
👍

PR #314, opened 2016-05-10
[DC 2016] Small post-migration changes

markcornick:
* "Washington, D.C." -> "Washington, DC"
* Sponsorship edits

### DIFF
--- a/data/events/2016-washington-dc.yml
+++ b/data/events/2016-washington-dc.yml
@@ -1,6 +1,6 @@
 name: 2016-washington-dc # The name of the event. Four digit year with the city name in lower-case, with no spaces.
 year: "2016" # The year of the event. Make sure it is in quotes.
-city: "Washington, D.C. " # The city name of the event. Capitalize it.
+city: "Washington, DC" # The city name of the event. Capitalize it.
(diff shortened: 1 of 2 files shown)


data/events/2016-washington-dc.yml
```diff
@@ -68,6 +68,7 @@ team_members: # Name is the only required field for team members.
 
 organizer_email: "organizers-washington-dc-2016@devopsdays.org" # Put your organizer email address here
 proposal_email: "proposals-washington-dc-2016@devopsdays.org" # Put your proposal email address here
+sponsors_accepted: "no"
```
bridgetkromhout: Right now this toggle turns off _all_ the "become a whatever sponsor" links. There's an open issue (https://github.com/devopsdays/devopsdays-web/issues/93) around toggling them per-level. Are you sure you want them all gone just to get rid of the Platinum one?

markcornick: For now, yes, I'd like them all gone. The "Become A Platinum & Venue Sponsor" link should never have been there in the first place, as there was only one (for the USPTO) and this is the easiest way to avoid confusion. I'll keep an eye on #93 and perhaps we will re-enable the Au/Ag level sponsor links when that lands. Thanks!

Issue #259, opened 2016-05-06
decouple setting the browser title bar from setting the header text displayed in page


yvovandoorn:
Current behavior:
When you modify a `title =` attribute in an event it influences the browser bar title (yay!) but it also influences header text in the body of the page. 

Example: 
<img width="1690" alt="title-issue" src="https://cloud.githubusercontent.com/assets/2694655/15088526/b872cf96-13f5-11e6-9231-2d1f0db8e0b3.png">

Expected behavior:
The `title =` attribute should modify `#1` which is the browser bar title. There should be a separate attribute to modify / alter the header text in the body. Perhaps a new attribute of `header_title` or `page_title` (I'm not marketing) could be created to customize / override.

mattstratton:
So the theory is making an optional bit of frontmatter - if it's not set on a page, it will default to the `title`, but you can override it with the frontmatter?

I think `header_title` seems to be principle of least surprise on that one.

bridgetkromhout:
This feels like a future enhancement to me. Do we still consider it part of a specific milestone that I feel like we've now passed?

yvovandoorn:
Yes. Future enhancement. Once your change is live I'd like to take stab at
it.

On Saturday, 4 June 2016, Bridget Kromhout <notifications@github.com> wrote:

> This feels like a future enhancement to me. Do we still consider it part
> of a specific milestone that I feel like we've now passed?
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/devopsdays/devopsdays-web/issues/259#issuecomment-223777022>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/ACkd_9F-dD7Mi8PdGlILsLNFx5Yk68GIks5qIeItgaJpZM4IZTBc>
> .
>


-- 

*Yvo van Doorn* — Manager, Technical Services (EMEA)

+31 6 38 74 07 46 <+316740746> – yvo@chef.io – *my: *Linkedin
<http://www.linkedin.com/in/yvovandoorn/>  Twitter
<https://twitter.com/yvov>

CHEF

CHEF.IO <http://www.chef.io/>

TM

*chef.io <http://www.chef.io/>*   Blog <http://www.chef.io/blog/>   Facebook
<https://www.facebook.com/getchefdotcom>   Twitter
<https://twitter.com/chef>   Youtube <https://www.youtube.com/getchef>

bridgetkromhout:
[...]

PR #315, opened 2016-05-10
Fixes display issue

bridgetkromhout:
Heads up to @barkerd427 - there's an open issue about this (https://github.com/devopsdays/devopsdays-web/issues/259) but for now, you don't want it to look like this (Kansas City twice in one line):

screen shot 2016-05-10 at 10 12 53 am


barkerd427:
Thanks!

PR #321, opened 2016-05-11
New proposal for Oslo 2016

steinim:


phrawzty:
The merge in a09c925 makes me nervous. I've prepared PR #322 with just the proposal.

PR #326, opened 2016-05-11
[DC 2016] Replace Capital One logo

markcornick:
The existing logo is for Capital One Investing; however, our sponsorship
is from Capital One corporate, so replacing the logo at their request.


phrawzty:
Hello,

Event-specific sponsor profiles are frowned upon (see the README); might I suggest modifying this PR such that this sponsor is referred to as "Capital One Corporate"?  In practice this would mean `s/-dc/-corp/;` I suppose, but it means that the profile can be used by other events more intuitively.

data/sponsors/capitalone-dc.yml
```diff
@@ -0,0 +1,2 @@
+name: "Capital One"
```
phrawzty: Maybe add "Corporate" in here so as to differentiate it from the existing Capital One profile?

markcornick:
I'll change from -DC to -corp when I get to my desk, but my position as well as the sponsor's is that the proper logo was in place before the migration, and unless we're going to break old events that use the Capital One Investing logo, a new profile is needed. And the sponsor wishes to be known simply as "Capital One", not "Capital One corporate."

data/sponsors/capitalone-dc.yml
markcornick: I chose to rename the existing capitalone.yml to "Capital One Investing" to match the image, and to leave capitalone-corp.yml (former capitalone-dc.yml) as "Capital One" to match its image and the sponsor's wishes.

kmugrage:
Seattle was the conflict. It turns out they had asked us to change it as well, and we missed it. In this case they should be standardized on the same logo.

phrawzty:
Hello again! I've spoken with an organiser of the only other even that Capital One is sponsoring, and they want to have the logo changed as well.  I will henceforth prepare a PR to replace the extant logo with the one that you've proposed here, thus satisfying both of your use cases neatly. PR incoming.

markcornick:
Thanks Dan and Ken!

PR #330, opened 2016-05-11
Adding a test.

bridgetkromhout:
I don't actually want to merge this. But weirdly, it doesn't build. It's confusing because it's just like Minneapolis (which builds).

Bridgets-MacBook-Air:devopsdays-web bridget$ !hugo
hugo server -w --baseUrl="http://localhost:1313"
ERROR: 2016/05/11 Error while rendering page events/2016-london/program/test-user.md: template: theme/talk/single.html:20:24: executing "theme/talk/single.html" at <index .Site.Data.spe...>: error calling index: int is not index type for map[string]interface {}
0 of 11 drafts rendered
0 future content
572 pages created
2 paginator pages created
17 tags created
0 categories created
in 1280 ms
Watching for changes in /Users/bridget/github/devopsdays/devopsdays-web/{data,content,layouts,static,themes}
Serving pages from memory
Web Server is available at http://localhost:1313/ (bind address 127.0.0.1)
Press Ctrl+C to stop

### DIFF
--- a/content/events/2016-london/program/test-user.md
+++ b/content/events/2016-london/program/test-user.md
@@ -1,0 +1,11 @@
++++
+City = "London"
+Year = "2016"
+date = "2016-03-06T21:28:07-06:00"
+title = "Test User"
+type = "talk"
+
++++
+
+This is a test.
+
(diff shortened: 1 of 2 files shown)


bridgetkromhout:
Fixed by @mattstratton in https://github.com/devopsdays/devopsdays-web/pull/332.

PR #332, opened 2016-05-11
Fix talk page

mattstratton:
There's an issue right now where "past" events need their year set as a non-string, but "current" events need it as a string. This raises hell on the template for a talk. This fixes that.


bridgetkromhout:
Thanks, Matt!

PR #340, opened 2016-05-13
Add Curtis Cockerham to organizers

barkerd427:
Added Curtis Cockerham to organizers list.


bridgetkromhout:
Send email to info@devopsdays.org - this should be merged at the same time they are added to the organizers email and to Slack.

PR #350, opened 2016-05-14
devopsdays Amsterdam 2016 - Initial Draft of program

yvovandoorn:
Initial draft of program.

This is a WIP hence why nav elements are still disabled.


yvovandoorn:
Self committing per @bridgetkromhout if it only effects your events data. 

I know self committing PRs is a no-no, but this is a hidden WIP, no nav elements activated yet.

Issue #295, opened 2016-05-08
redesign how sponsors are displayed


bridgetkromhout:
We have a number of open issues around the sponsor logos. This meta-issue will help us plan that work. I just read through all the sponsor-related issues, and there's one obvious choice that helps almost every issue.

I assert that the sponsors belong at the bottom of each page, not on the right-hand side. This would allow a lot more flexibility in the responsiveness of the pages without constraining us to make room for a column. I'm picturing something pretty much exactly like what Velocity does: http://conferences.oreilly.com/velocity/devops-web-performance-ca (scroll down).

Here are all the issues around how we display sponsors:

https://github.com/devopsdays/devopsdays-web/issues/227 - consider allowing different sizes of sponsor images
https://github.com/devopsdays/devopsdays-web/issues/216 - consider supporting sponsor links with no image
https://github.com/devopsdays/devopsdays-web/issues/197 - Make sponsor border links optional
https://github.com/devopsdays/devopsdays-web/issues/195 - Add ability to disable sponsor sidebar
https://github.com/devopsdays/devopsdays-web/issues/108 - Sponsor logos are two wide instead of three
https://github.com/devopsdays/devopsdays-web/issues/87 - Sponsor sidebar looks weird on iOS
https://github.com/devopsdays/devopsdays-web/issues/93 - Ability to control "Become a <level> sponsor" links
https://github.com/devopsdays/devopsdays-web/issues/26 - add an enhancement to show "become the FIRST x sponsor"
 
Discussion welcome.

mattstratton:
Bear in mind with a responsive design the sponsors could display "stacked" (not a sidebar) on narrower view ports and at a really wide viewport they could be on the right as usual.

mattstratton:
That being said, putting the sponsors at the bottom makes for a slightly easier bit of front end coding :)

WhatsChrisDoing:
I think we need to move ahead with this. The sponsor logos are wrapping to the bottom of the page on most mobile devices in any case.

mattstratton:
I'm leery about jumping into "redesign"efforts piecemeal. There needs to be an overall approach. Which we should be working towards as soon as possible.

WhatsChrisDoing:
[...]

PR #355, opened 2016-05-14
Moves sponsors to bottom

bridgetkromhout:
I think @mattstratton is right in https://github.com/devopsdays/devopsdays-web/issues/295 and we should approach the re-design in a considered fashion. When I started looking at what it would take to move the sponsors to the bottom, though, it was so trivial that I figured I'd throw a PR together for consideration.

I ran through all pages on all 2016 events to make sure this change wouldn't make them unusable. There are a few cosmetic issues caused by how non-responsive the site is in general, but https://github.com/devopsdays/devopsdays-web/commit/a1709011e286bf547ed7267a51d29ad3163b700b fixes the only major issue.

Because moving sponsors to the bottom means there is a lot more space for them, I went with what I thought was more aesthetically pleasing of no borders (https://github.com/devopsdays/devopsdays-web/commit/d6b82c510a980b62d0fa7e40a92f3a6f971fdc5d). Here's what that looks like in practice:

screen shot 2016-05-14 at 5 03 32 pm

There are improvements we could make in the future (perhaps centering all this) that should definitely wait for a redesign.

I also widened the 2016 programs as seen in these before & after screencaps. While it could be argued that should wait for the redesign, it looked really weird without the sponsor sidebar since it didn't move to fill the space.

Before:
screen shot 2016-05-14 at 5 12 49 pm

After:
screen shot 2016-05-14 at 5 13 12 pm

Let the bikeshedding commence. :) (Do not merge without @mattstratton's 👍 .)

### DIFF
--- a/content/events/2016-amsterdam/program.md
+++ b/content/events/2016-amsterdam/program.md
@@ -14,116 +14,116 @@ tags = ["amsterdam","amsterdam-2016"]
 
 
 <center><b><h2>Schedule</h2></b></center>
-<div class="span-7 append-bottom border">
+<div class="span-11 append-bottom border">
 
-<div class="span-7 last">
+<div class="span-11 last">
 <h4>30 June - Thursday</h4>
 </div>
 
-<div class="span-2">08:00-09:00</div><div class="span-4 box last">Registration, Breakfast, and Sponsor Booths Open</div>
-<div class="span-2">09:00-09:15</div><div class="span-4 box last"><strong></strong>Opening Welcome</div>
-<div class="span-2">09:15-09:45</div><div class="span-4 box last">
+<div class="span-2">08:00-09:00</div><div class="span-8 box last">Registration, Breakfast, and Sponsor Booths Open</div>
+<div class="span-2">09:00-09:15</div><div class="span-8 box last"><strong></strong>Opening Welcome</div>
+<div class="span-2">09:15-09:45</div><div class="span-8 box last">
 <a href="/events/2016-amsterdam/program/erica-baker">Erica Baker (Slack) - Diversity in Technology</a> (Opening keynote)
     <br />
 </div>
-<div class="span-2">09:45-10:00</div><div class="span-4 box last">
+<div class="span-2">09:45-10:00</div><div class="span-8 box last">
   Sponsors
 </div>
-<div class="span-2">10:00-10:30</div><div class="span-4 box last">
+<div class="span-2">10:00-10:30</div><div class="span-8 box last">
 <a href="/events/2016-amsterdam/program/daniel-van-gils">Daniël van Gils (Cloud66) - How the hell do I run my microservices in production, and will it scale?</a>
     <br />
 
 </div>
-<div class="span-2">10:30-10:45</div><div class="span-4 box last">
+<div class="span-2">10:30-10:45</div><div class="span-8 box last">
   Break
 </div>
 
-<div class="span-2">10:45-11:15</div><div class="span-4 box last">
+<div class="span-2">10:45-11:15</div><div class="span-8 box last">
 <a href="/events/2016-amsterdam/program/warner-more">Warner More (CoverMyMeds) - DevOps has Always Been About Security</a>
     <br />
 </div>
 
-<div class="span-2">11:15-11:30</div><div class="span-4 box last">
+<div class="span-2">11:15-11:30</div><div class="span-8 box last">
   Sponsors
 </div>
 
-<div class="span-2">11:30-12:00</div><div class="span-4 box last">
+<div class="span-2">11:30-12:00</div><div class="span-8 box last">
 <a href="/events/2016-amsterdam/program/avishai-ish-shalom">Avishai Ish-Shalom (Fewbytes) - The Mathematics of Reliability</a>
     <br />
 </div>
 
-<div class="span-2">12:00-13:00</div><div class="span-4 append-bottom last">Lunch (catered)</div>
+<div class="span-2">12:00-13:00</div><div class="span-8 append-bottom last">Lunch (catered)</div>
 
-<div class="span-2">13:00-14:00</div><div class="span-4 box last"><strong>Open Space</strong> (Open Space) <br /> Open Space Opening</div>
+<div class="span-2">13:00-14:00</div><div class="span-8 box last"><strong>Open Space</strong> (Open Space) <br /> Open Space Opening</div>
 
-<div class="span-2">14:00-14:45</div><div class="span-4 box last"><strong>attendee-suggested <a href="/pages/open-space-format">Open Space</a> breakout discussions</strong> Open Space #1</div>
+<div class="span-2">14:00-14:45</div><div class="span-8 box last"><strong>attendee-suggested <a href="/pages/open-space-format">Open Space</a> breakout discussions</strong> Open Space #1</div>
 
-<div class="span-2">14:45-15:00</div><div class="span-4 box last"><strong>Break</strong> <br /> </div>
+<div class="span-2">14:45-15:00</div><div class="span-8 box last"><strong>Break</strong> <br /> </div>
 
-<div class="span-2">15:00-15:45</div><div class="span-4 box last"><strong>attendee-suggested <a href="/pages/open-space-format">Open Space</a> breakout discussions</strong> Open Space #2</div>
+<div class="span-2">15:00-15:45</div><div class="span-8 box last"><strong>attendee-suggested <a href="/pages/open-space-format">Open Space</a> breakout discussions</strong> Open Space #2</div>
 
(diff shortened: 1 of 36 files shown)


kmugrage:
This looks much better. 

(Sponsor hat on) When sponsoring, I generally try to get my AP to pay as quickly as possible so that our logo is "above the fold". In this layout, that's never going to happen, so the exposure is a bit lower. I'm personally ok with the negligible impact, but others may have larger concerns.

MikeRosTX:
I personally like the esthetics of the pages because there's more real estate to the right to accommodate the schedule and other items.

What I'm afraid of are the complaints coming from sponsors stating that people might not scroll all the way down the page to see their logo and who is supporting/sponsoring the particular event at each city.

yvovandoorn:
So screenshot for after doesn't show what happens to the actual sponsor. How does prioritizing sponsor levels work.

mattstratton:
@yvovandoorn - see the earlier screenshot 

https://cloud.githubusercontent.com/assets/2104453/15270924/edcce328-19f6-11e6-988e-d421d03933d3.png

bridgetkromhout:
To address @yvovandoorn's question:

A view of the "after" Minneapolis main page "above the fold" when zoomed all the way out:

screen shot 2016-05-14 at 8 29 08 pm

Scrolled down to show all the Minneapolis sponsors:

screen shot 2016-05-14 at 8 28 58 pm

The Amsterdam front page (zoomed all the way out):

screen shot 2016-05-14 at 8 31 04 pm

Scrolled down, the rest of the Amsterdam sponsors:

screen shot 2016-05-14 at 8 31 37 pm

For both Minneapolis and Amsterdam I'd probably recommend a div to rearrange the welcome page a bit, take better advantage of the extra space horizontally.

I understand sponsors wanting visibility - I work for a sponsor too - and I also know the the reality of most conference websites is that the sponsors are in a footer on every page (at best) or tucked away on a sponsor page (at worst). And as we saw in the complaints about how the sidebar would scroll off to the bottom, we're building a confusing expectation with the current layout (see @whatschrisdoing's comment in https://github.com/devopsdays/devopsdays-web/issues/295#issuecomment-219218290).

Also an argument could be made that being in a fixed-width sidebar on the right is dreadful for mobile, since you'll never be seen at all if people are zoomed in enough to read.

@kmugrage is right that we may have created an artificial economy of rushing for placement on the page - but recall, the old site also would shuffle the icon placement inside a level! which we're no longer doing.

mattstratton:
Honestly the right solution is to make it responsive. At a larger breakpoint the sponsors will be on the right and a smaller breakpoint they go in the bottom.

bridgetkromhout:
> Honestly the right solution is to make it responsive. At a larger breakpoint the sponsors will be on the right and a smaller breakpoint they go in the bottom.

Agree. And that's not something we'll solve today (as nice as that would be), since we do need to work with a designer.

The problems I'd like to solve right now are:
* difficult to read the narrow-column program
* sponsor icons look cramped
* difficult to see all the sponsors of a given level together because only two (or three if we change that) fit per row
* difficult to do bigger sponsor icons for higher levels (this doesn't solve it but it sets us up to solve it)

And the reason this is currently bugging me more than it did before:
* video embeds of a fixed size like those generated by vimeo spill over onto the sponsors (ignore the lack of permissions to embed - London knows and says they'll fix it)

screen shot 2016-05-14 at 9 20 15 pm

MikeRosTX:
I am definitely OK with having sponsors at the bottom. As long as we all agree to inform sponsor that complain, that we're just aligning up with how other conference sites publish sponsors listed at the footer.

Agree with Bridget regarding video embedding so close to the sponsor logos would bug me too. Definitely having sponsors at the footer is better to view schedules, embedded videos, Eventbrite, etc.

jedi4ever:
I agree with @kmugrage - logo above the fold is what the logo on the site is worth - if you're a smalller event it's one of the few extras you get. Moving it below the fold is less valuable.

Given screens are wider now I'd suggest keeping it right and only moving it below the fold if on a mobile device or smaller screen estate

yvovandoorn:
:+1: @kmugrage and @jedi4ever.

Keeping it on the right in order of level is something a majority of our sponsors expect. In most situations we have the screen real estate these days to have it in the right.

MikeRosTX:
I completely understand what @kmugrage, @jedi4ever & @yvovandoorn are stating about keeping it on the right side visible and valuable to the sponsors.

But if we clearly explain to those sponsors who complain that we're just trying to make DevOpsDays website look similar to the way other major conferences look that have the sponsors at the bottom, I'm sure they'll understand.

It's better to ask for forgiveness than their permission up front.  If push came to shove and we had a MAJORITY of the sponsors raise a lot of complaints, we can always change it back.  But we should at least give it a try and to those sponsors who complain, we can point the to these big conference to show them that all sponsors are at the bottom of the screen.

http://events.itrevolution.com/us/sponsors/

https://ignite.microsoft.com/#fbid=CidXhYwSAu1

http://conferences.oreilly.com/oscon/open-source-us

mattstratton:
Also please see #359 where I will be working to make this responsive.

yvovandoorn:
TL;DR - The sponsors (for a significant part) fund the event, the community & organizers makes the program happen. Shoving sponsors to the bottom where screen real estate allows for them to co-exist seems unnecessary devalues for a sponsor to sponsor (and thus could impact how we reach the community).

@MikeRosTX so this is my 1st year part of devopsdays, but one of the most refreshing things is that devopsdays *isn't* like other conferences. Personally, and this is my 2 cents, I don't want devopsdays to be compared to ignite, velocity, oscon, chefconf or others. 

The right side to me has always felt natural. The way we often position sponsors at these events is that their sponsorship keeps ticket costs down, thus accessibility high towards new attendees.

yvovandoorn:
Thinking more about it -- this should be a layout choice that each event can make as they setup their site versus a global enforced change. 

The default staying how it is (at least for 2016) and if certain cities decide to go with the new layout, they can declare that in `data/layout/events/year-city.yml` file as `sponsors: bottom`

mattstratton:
The challenge is that on a smaller screen you do no have the real estate for a sidebar. Which is why on smaller screen it should break to being at the bottom but stay in the side for larger break points. See the linked reposive issue.

MikeRosTX:
@yvovandoorn, this is also my 1st year as local organizer of devopsdays.  But I like your idea of "_this should be a layout choice that each event can make as they setup their site versus a global enforced change._"

yvovandoorn:
@mattstratton I completely get the mobile impact and completely agree that on a mobile device, content (program info) wins over trying to share it with sponsors.

But any device doing a native resolution of 1024x768 will properly render the page.

mattstratton:
Yes that is my design idea.

mattstratton:
Also, with regard to the video overflow, by enforcing responsive stuff that I have in mind, these things can be protected against.

mattstratton:
Replying to @bridgetkromhout statement that we cannot implement responsive until we have a designer - not necessarily. I can implement the same look and feel (no change to UX) but make it responsive without a designer. We need a designer to refactor the overall look and feel. I think I can convert what we have now to a responsive layout which will solve for these issues but not change the user experiece overall.

WhatsChrisDoing:
@mattstratton Would the responsive layout allow the sponsors to fit on the right more often on the home page? Would it make more sense to just have the sponsors on the home page and not on all of the other pages?

mattstratton:
I don't understand what "more often on the right" means? Can you clarify?

WhatsChrisDoing:
That the sponsor logos would display on the right side of the page at narrower page widths. Currently the sponsor logos wrap to the bottom when the window is about 990 pixels on my Mac. It's the wrong question. I think our event logo would drive the minimum column size in any case.

barkerd427:
I'd vote for reactive. We'll always have new events, and part of their value to sponsors is page presence. I'd also be ok with sponsors only being on the event homepage.

mattstratton:
I think that having the sponsors display on the right at a narrow viewport makes for a really bad user experience. It makes everything squished.

WhatsChrisDoing:
Likely true. I think where I'm landing after this discussion is to stay with the status quo for now and execute a home page redesign that better accommodates the sponsor logos at typical tablet portrait and landscape screen width.

iennae:
I think having sponsors appear on every page on the bottom is great. On the sponsors page making them front and center is awesome too. Not every conference puts sponsors on every single page. I think the schedule (as well as embedded videos) much nicer if things weren't crammed all to the left. Long titles, and times where multiple things are happening get smushed so it's not as nice. 

Based on the fold that Bridget shows, it looks like platinum (or whichever top sponsor) will get premium space as well which also will help folks in selecting for that visibility.

iennae:
Also.. is this a bug??
When the page is bigger the sponsors move up a level and squish the menu. 

Smaller Window - Doesn't Squish menu
Larger Window - Squishes local menu

Originally I wanted to point out from a user point of view, this is a lot of content before we actually get to the content around the schedule.

bridgetkromhout:
I'm closing this unmerged for now. We can keep discussing what it should look like in https://github.com/devopsdays/devopsdays-web/issues/295. Thanks, all.

PR #376, opened 2016-05-17
devopsdays AMS 2016 - Add Ignite details, block out unconfirmed talks

yvovandoorn:
* Added speaker bios to all ignite speakers
* Added photos from those that have sent those in (will continue to receive more)
* Deactivated a talk until confirmed
* Initial workshop schedule.


yvovandoorn:
- Removed a talker that isn't able to speak
- Worked around the issue @bridgetkromhout pointed out on missing headshots.

PR #383, opened 2016-05-18
Clean-up DC Sponsorship

nathenharvey:
* Markdown instead of HTML
* Gold Sponsorships are sold out, too.


phrawzty:
> Markdown instead of HTML

💯 

> Gold Sponsorships are sold out, too.

Congrats!

PR #384, opened 2016-05-18
Refactoring language around keynote talk

yvovandoorn:


yvovandoorn:
self merging the last change as it impacts language on our keynote.

PR #409, opened 2016-05-20
added brocade and update shomi

philippetremblay:


mattstratton:
I fixed the merge issue on this; should be good.

PR #435, opened 2016-05-23
[DC 2016] move Solidfire to silver sponsors

markcornick:


markcornick:
Disregard - stuff got seriously messed up during GitHub's outage this morning.

PR #439, opened 2016-05-23
[DC 2016] move Solidfire to silver sponsor

markcornick:


markcornick:
Ugh. Trying a fourth time. Thanks GitHub.

PR #455, opened 2016-05-24
Sponsor prices should increase, not decrease

nathenharvey:
The headers in the sponsor table were incorrect


yvovandoorn:
👍

PR #479, opened 2016-05-30
Remind Ignite presenters that there are no presenter notes

jasonious:
For my first Ignite, I asked the organizer 15 minutes before stage time
where I'd be able to see my notes.  I was then informed that Ignite talks
don't get Presenter Notes, which was quite a surprise.

Letting the Ignite presenters know this ahead of time, as well as reminding
them on the day of, should help eliminate future surprises.

Also mentioned that animations and videos do not work either

### DIFF
--- a/content/page/organizing.md
+++ b/content/page/organizing.md
@@ -499,6 +499,8 @@ If you're going to livestream, rehearse ahead of time, and then assign at least 
 Running Ignites can be a challenge: people tend to submit them last minute, or they didn't understand the format too well. Here are few tips on making this process run more smoothly:
 
 - Ask the Ignite presenters to send the slides ahead of time
+- Inform the Ignite presenters that there are no presenter notes during their talks
+- Animations (gifs or slide transitions) and videos do not work due to the conversion process
  - An option is to use a Dropbox shared directory
 - Be prepared to run around with a USB stick to get the 'on the spot' deciders.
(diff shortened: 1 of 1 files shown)


mattstratton:
The bit about animations etc depend on how the event does it. Not sure we want to assume all events will be removing them.

mattstratton:
Likewise, it's possible for an event to construct the ignites in a way where presenter notes are supported.

phrawzty:
> Likewise, it's possible for an event to construct the ignites in a way where presenter notes are supported.

To be fair, the doc specifically suggests *impressive*, which doesn't support speaker notes. Perhaps some clarification around that particular topic might be helpful?

KrisBuytaert:
This is what I put in most of my ignite speaker prep mails 

"We need PDF files , with 20 slides maximum , our tool does the
autoforwarding. 

You can test run them with 
impressive -a 15 yourslides.pdf

"

mattstratton:
I think given that the direction of the doc is assuming use of `impressive` this probably makes sense. We have never used it at Chicago so I didn't even notice it was mentioned :)

I'm okay with merging this given that.

bridgetkromhout:
Right. We can do some clarification around _how_ to run the Ignites (Minneapolis doesn't use impressive, for example), but in the vast majority of cases, notes and animations aren't going to work.

Thanks, @jasonious!

Issue #436, opened 2016-05-23
twitter link in footer doesn't actually go to @devopsdays


mattstratton:

bridgetkromhout:
Fixed in https://github.com/devopsdays/devopsdays-web/pull/484.

mattstratton:
Ahhh...see, this is why I did the "fixed in branch" label, because I was keeping the issues open until we merged this branch in. That made sense to me at the time for some reason.

bridgetkromhout:
Ah, okay. I figure we're close enough to merging now. But as you wish!

PR #484, opened 2016-05-31
A few suggestions

bridgetkromhout:
Working through the redesign, I wanted to make these changes:

* Don't we need to be on the responsive theme in this branch? I can remove this commit if in fact we shouldn't, but I need this change to see any updates of this theme locally.

* Fixed https://github.com/devopsdays/devopsdays-web/issues/436 though that doesn't address the namespacing to params-social.

* Cleaned up a bit by removing the old "/presentations" link and putting the useful content from it in "/events" for now (probably to be refactored when that page is redone). Used the lightbulb for the ideas around organizing.

* Brought in the less-complicated program changes. More to come.


mattstratton:
You can add `-t devopsdays-responsive` to your `hugo server` command in order to tell it to use that theme :)

mattstratton:
That being said, changing the config file works too :)

PR #495, opened 2016-06-01
2016 boston logos

masteinhauser:
Add new sponsors.
Fix the brainshark logo as it seemed to be rendering poorly.


phrawzty:
Hi there! It's not a huge deal, but logo files should be 200x200 - the brainshark logo in 9216874 is 250x250. If it's convenient for you to scale it, that'd be preferable. Thanks!

masteinhauser:
@phrawzty fixed! I wasn't sure if that would be cool, and an easy fix for me after re-figuring out how to do it all this morning.

PR #500, opened 2016-06-02
DevOpsDays Porto Alegre - Adding another sponsor and open the registration

gomex:


data/events/2016-portoalegre.yml
```diff
@@ -45,11 +47,15 @@ proposal_email: "proposals-portoalegre-2016@devopsdays.org" # Put your proposal
...
     level: gold
+  - id: instruct
+    level: gold
   - id: arresteddevops
     level: community
 
 sponsors_accepted : "yes" # Whether you want "Become a XXX Sponsor!" link
 sponsor_levels: # In this section, list the level of sponsorships and the label to use.
+  - id: platinum
+    label: Platinum
```
bridgetkromhout: Just so you know, the "label" is for display purposes. You can put those in Portuguese if you like.

PR #501, opened 2016-06-02
Replace Solidfire logo with NetApp logo

markcornick:
Requested by Jessica Haze, Integrated Marketing Manager at SolidFire, who says:

> As of 6:00 AM MTN tomorrow, 6/2, the SolidFire logo goes away and we’ll be using the NetApp logo moving forward. The @solidfire twitter handle should remain the same.

Leaving the link as www.solidfire.com based on this advice and on a check of the site this morning, which has a SolidFire page with a NetApp logo.

cc @nathenharvey


bridgetkromhout:
This affects only Minneapolis and DC. CCing @zruty so she can make sure our signage comes out correctly too.

PR #503, opened 2016-06-02
new thoughtworks products logo

kmugrage:


bridgetkromhout:
If I merge this, it will be effective now for _all_ devopsdays sites using this logo. This will force a retroactive change for events that are already past. I'd prefer a PR that moves `static/img/sponsors/thoughtworks-products.png` and copies `data/sponsors/thoughtworks-products.yml` to a clearly-namespaced-in-the-past name like `thoughtworks-product-before-20160602.whatever` and updates the events on this list that already happened. Ideally a tool that does just that, which I recognize is a pain to do manually. Then `static/img/sponsors/thoughtworks-products.png` can be this new image and `data/sponsors/thoughtworks-products.yml` can be whatever you want (possibly exactly what it is now).

$ grep -r thoughtworks-products data/events/*
data/events/2016-amsterdam.yml:  - id: thoughtworks-products
data/events/2016-austin.yml:  - id: thoughtworks-products
data/events/2016-chicago.yml:  - id: thoughtworks-products
data/events/2016-denver.yml:  - id: thoughtworks-products
data/events/2016-minneapolis.yml:  - id: thoughtworks-products
data/events/2016-portland.yml:  - id: thoughtworks-products
data/events/2016-saltlakecity.yml:  - id: thoughtworks-products
data/events/2016-seattle.yml:  - id: thoughtworks-products
data/events/2016-toronto.yml:  - id: thoughtworks-products
data/events/2016-vancouver.yml:  - id: thoughtworks-products
data/events/2016-washington-dc.yml:  - id: thoughtworks-products

I really like the simplicity of "every new event can just use thoughtworks-product because it's assumed current" but we should solve the wider problem too in this case. (Didn't have to worry about this for the solidfire->netapp switch because it only affected two events that were still in the future.)

Flagging @mattstratton for insights.

kmugrage:
Should the decision about "use the one which was accurate at the time" vs "change them all for consistency" be decided by the DevOpsDays events or the organization represented by the logo? 

For this specific case, it's my desire (or more accurately my marketing team's desire) as the logo owner to have them all change, as we've been told many times the previous logo is confusing.

mattstratton:
IMHO if the sponsor is the one who wants them all to change, then it's fine to change retroactively. Pending the following points:

1) it's a little bit like "rewriting history"
2) i wouldn't want to set the precedent of a sponsor being able to request this (as its effort to be done). In this case since Ken is also an organizer and website contributor, there wasn't work created from the global change for the sponsor. If a sponsor had just asked for this to be done it creates a work item for someone to do. 

The tricky part is that it does change a bunch of sites out from underneath them, probably without their awareness.

mattstratton:
I guess the question is - if not handled this way, would ThoughtWorks be asking all the existing events to update the logo anyway?

kmugrage:
In this specific case ThoughtWorks would ask for the change, because people think we have a product called "go snap", and the bad logo was created specifically for the square devopsdays requirement. That said, the desire doesn't necessarily make it a reasonable request in the webby world of needing multiple changes. 

My thought on creating the PR was that with Hugo it's actually easier to do a bulk change than only future looking. I hadn't considered the fact that I would be changing finished events without talking to those cities. I think I would have done it the same way if a company other than my own was making the request. 

If changing another city's event is the primary concern I'm not sure it shouldn't mean that getting buy in from future cities is needed as well. That seems pretty hard.

kmugrage:
I've sent email to the organizer's list for the past events to see if anyone objects.

bridgetkromhout:
I don't like the idea of rewriting history for completed events. For good or for ill, that was the logo that was in use at the time of the event. [Reconning](http://tvtropes.org/pmwiki/pmwiki.php/Main/Retcon) it is inaccurate.

I am fine with a company wanting to change all logos on all events that haven't happened yet. That doesn't require consent from every affected city (though probably they'll want to know anyhow, for purposes of updating banners and the like).

So in this case, I'm fine with doing the work if @kmugrage doesn't have time, and I mostly want @mattstratton's sanity check on whether my proposed name-spacing is the best way to preserve history while also making the "default" easy to use.

I object to any sponsor getting the green light to rewrite history.

kmugrage:
Ah, ok, didn't understand that you didn't want it changed regardless. I can do the work.

Thanks

kmugrage:
I believe this now preserves all past events who used that logo. The more I think about it the more I agree that rewriting history wasn't the greatest idea. Even if it was easier :)

bridgetkromhout:
Thanks, @kmugrage!

PR #508, opened 2016-06-03
Replace Ansible logo

markcornick:
Requested by Carrie Drummond, Marketing Coordinator at Ansible/Red Hat,
who says "We're now just using the Ansible wordmark for our community
efforts."


markcornick:
This appears to only affect 2016 DC and 2015 Mpls.

bridgetkromhout:
Thanks, @markcornick. We don't want to rewrite history, so I kept MSP 2015 on the old logo via https://github.com/devopsdays/devopsdays-web/pull/509. Now we can update this moving forward.

PR #520, opened 2016-06-04
Adds link to future & past events where expected.

bridgetkromhout:
screen shot 2016-06-04 at 10 35 11 am

I found myself accidentally scrolling to the lower right a lot and then wondering where the sorted-by-date, not-all-even-on-map event list was. This is the minimum viable "fix that". /cc @kmugrage @mattstratton


kmugrage:
Good idea

PR #527, opened 2016-06-05
Updated map zoom and labels

### LINKED ISSUE #82: Add support to localize map
The map is getting pretty cluttered, so it would be good to be able to localize the map to a region by zooming based off the user's location. If the location can't be found, then we can either zoom all the way out or zoom to a default location. In the past, Google has zoomed to their headquarters. That was annoying, so hopefully we can find a better solution.

Here is some code I've used in the past. I can't promise that the code is complete, but I'm happy to add more of my code as the site is still up and running with this functionality. I'm an awful front-end dev, so any help is appreciated.

```
function initialize() {
		  var mapOptions = {
		    zoom: 12
		  };
		  map = new google.maps.Map(document.getElementById('map-canvas'),
		      mapOptions);

		  // Try HTML5 geolocation
		  if(navigator.geolocation) {
		    navigator.geolocation.getCurrentPosition(function(position) {
		      var pos = new google.maps.LatLng(position.coords.latitude,
		                                       position.coords.longitude);

		      map.setCenter(pos);
		    }, function() {
		      handleNoGeolocation(true);
		    });
		  } else {
		    // Browser doesn't support Geolocation
		    handleNoGeolocation(false);
		  }
		}

function handleNoGeolocation(errorFlag) {
		  if (errorFlag) {
		    var content = 'Error: The Geolocation service failed.';
		  } else {
		    var content = 'Error: Your browser doesn\'t support geolocation.';
		  }

		  var options = {
		    map: map,
		    position: new google.maps.LatLng(39.106667, -94.676389),
		    content: content
		  };

		  map.setCenter(options.position);
		}

google.maps.event.addDomListener(window, 'load', initialize);
```

This code will take the user's location if allowed or set it to a default location if the user doesn't allow the location service. It also sets the default zoom level. Some amount of this was lifted from some tutorial several years ago.

barkerd427:
This closes #82 by adding support for localizing the map. I have currently set the zoom at 6, but we can play with that if it isn't optimal. If location information isn't available, then the default zoom is applied.

I also removed the labels and added info windows instead. They're still slightly buggy in that the "disappear on drag" functionality only works on the last window to appear. I've done this in the past, but it may require additional restructuring. I'll log an issue if this PR is merged.

### DIFF
--- a/themes/devopsdays-responsive/layouts/partials/map.html
+++ b/themes/devopsdays-responsive/layouts/partials/map.html
@@ -28,9 +28,6 @@ function initialize() {
       {{ end }}
     ];
 
-    // Display multiple markers on a map
-    // var infoWindow = new google.maps.InfoWindow(), marker, i;
-
     // Loop through our array of markers & place each one on the map
     for( i = 0; i < markers.length; i++ ) {
         var position = new google.maps.LatLng(markers[i][1], markers[i][2]);
@@ -38,11 +35,18 @@ function initialize() {
         marker = new MarkerWithLabel({
             position: position,
             map: map,
-            labelContent: markers[i][0] + "<br>" + markers[i][3],
-            labelAnchor: new google.maps.Point(25, 50),
-            labelClass: "labels",
-            labelStyle: { opacity: 1 }
         });
+
+        marker.info = new google.maps.InfoWindow({
+            content: markers[i][0] + "<br>" + markers[i][3]
+        });
+
+        google.maps.event.addListener(marker, 'click', function() {
+            infoWindow = this.info
+            infoWindow.open(map, this);
+            google.maps.event.addListener(map, "drag", function() {infoWindow.close();});
+        });
+
         // cityLink = markers[i][4]
         // This isn't working right now btw
         // google.maps.event.addListener(marker, "click", function (e) { location.href=cityLink});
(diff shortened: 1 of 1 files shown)


bridgetkromhout:
Thanks. I'll check it over more carefully tomorrow. Meanwhile, please put the Kansas-City-specific stuff into a different PR (per the request in the README - we don't want to mix "stuff for a specific city" with "stuff that affects the whole site" in the same PRs). Thanks!

barkerd427:
Updated.

bridgetkromhout:
Commented in #82 where we can discuss the design. Thanks!

mattstratton:
I have built this branch to http://workstation.mattstratton.com if anyone wants to see what it looks like.

mattstratton:
just a note for those of you with merge powers - this PR requires a thumbs up from both @bridgetkromhout and @mattstratton before merging.

yvovandoorn:
In general I like it. It looks less messy with the city names not shown. 

What would make it even better though if the city names it now exposes are clickable links to the actual event (that wasn't working for me on Chrome 52). 

My 2 cents.

jyee:
I like it, but similar to Bridget's comments on #82, I have a tendency to click "No" when sites ask for my location and it's unclear what providing that info will do (also for some reason, Chrome is not asking and simply showing the global, zoomed-out map).

I wonder if some sort of "See events near you" button that would trigger the location request and then zoom to an appropriate level (either a preset zoom number or encompassing the nearest X number of events near the geolocation).

mattstratton:
FYI, "do not merge" does not mean rejected, it's just a signal to other maintainers :)

barkerd427:
@yvovandoorn, that may be easy, but it's largely dependent on what information is available for that marker. Swapping in a link is easy if there's one available.

barkerd427:
@jyee, not sure why it isn't asking, but the default is that it zooms to fit all markers. It only changes if you select yes. You may have a global config switched that has turned off the location prompt. We can certainly add a separate button, but it would still prompt for access to the location information.

bridgetkromhout:
Per the discussion in #82, closing this without merging. Thanks for helping us work through this, @barkerd427!

PR #529, opened 2016-06-05
Don't display Twitter "@" when handle isn't set

### LINKED ISSUE #366: remove twitter handle link if it doesn't exist
We're just linking a `@` to twitter.com if there's no twitter handle, from:

 https://github.com/devopsdays/devopsdays-web/blob/master/themes/devopsdays-legacy/layouts/speakers/single.html#L20 

and

https://github.com/devopsdays/devopsdays-web/blob/master/themes/devopsdays-legacy/layouts/talk/single.html#L29

This needs a better fallthrough for leaving this out if not set. Related to https://github.com/devopsdays/devopsdays-web/issues/365. Heads up to @yvovandoorn.

yvovandoorn:
Add a hugo parameter check prior to displaying twitter handle. By default if the parameter is not set (0 byte content) it pass the if statement. (ref: https://gohugo.io/templates/go-templates/#conditionals)

This fixed #366

### DIFF
--- a/themes/devopsdays-responsive/layouts/speakers/single.html
+++ b/themes/devopsdays-responsive/layouts/speakers/single.html
@@ -18,9 +18,9 @@
     </div>
     <div class= "col-md-8">
        <h3><a href="/events/{{ $event_slug }}/program/{{$fname}}">{{ $s.name }}</a></h3>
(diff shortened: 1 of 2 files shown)


bridgetkromhout:
Thanks, @yvovandoorn!

PR #542, opened 2016-06-06
Add underline on hover to event headline

### LINKED ISSUE #510: Submenu links don't appear as links
Links in the submenus don't appear as links; they're just black text on a gray box. For example:

screen shot 2016-06-03 at 4 53 14 pm

screen shot 2016-06-03 at 4 53 01 pm

I was going to make a PR with a tweak but you're using Less and I got a bit lost trying to find where I was supposed to be. 😊  The compiled asset is http://www.devopsdays.org//css/devops.min.css and with this fix:

```css
 .submenu h3 a {
	color: #333333;
	/* text-decoration: none; */
}
```

You end up with this nicer submenu:

screen shot 2016-06-03 at 4 55 05 pm

cc @bridgetkromhout as you asked me to drop a line here

mattstratton:
fixes #510


yvovandoorn:
+1 makes it easier to read!

Issue #543, opened 2016-06-06
Upgrade to Hugo 0.16


mattstratton:
http://gohugo.io/meta/release-notes/

mattstratton:
Seems like there are some potential issues.

Running `hugo -v` against the existing repo using hugo 0.16.0 results in some errors [gist](https://gist.github.com/mattstratton/2da9509f62dd6a220cfd5b5091f3a842).

Many more errors show up when running `hugo server` [gist](https://gist.github.com/mattstratton/2b2c906ce167ed3cf1cffe80b3058ca1)

mattstratton:
Thread on Hugo forums on the topic:
https://discuss.gohugo.io/t/errors-with-type-template-html-should-be-string-with-hugo-0-16-0/3470

mattstratton:
This seems to be a hugo bug:

https://discuss.gohugo.io/t/errors-with-type-template-html-should-be-string-with-hugo-0-16-0/3470/4?u=mattstratton

The workaround is to wrap it in a string function:

`(string (chomp $e.year))`

mattstratton:
This is the bug in hugo to fix this issue https://github.com/spf13/hugo/issues/2187

mattstratton:
Hugo 0.16 is also causing some weirdness on event pages:


![](https://files.slack.com/files-pri/T04RN59HS-F1GBDBK5E/pasted_image_at_2016_06_13_08_34_am.png)

mattstratton:
At this point, I think that everything works, but it needs testing.

PR #546, opened 2016-06-06
Add string function to work around hugo 0.16 bug

mattstratton:
The chomp function doesn't work quite as expected in 0.16. Per https://discuss.gohugo.io/t/errors-with-type-template-html-should-be-string-with-hugo-0-16-0/3470/4?u=mattstratton, i have wrapped the chomp stuff and it seems to work fine now under 0.16.

This handles the defect mentioned in #543


mattstratton:
DO NOT MERGE this change! This only works with hugo 0.16.0 right now.

mattstratton:
This has been tested with 0.15 and it works now. So feel free to merge, but only if you've checked it out. Don't believe me :)

mattstratton:
I am merging this as it will fix an issue with the build.

PR #561, opened 2016-06-08
Add Bert Weidt to KC

barkerd427:
Adding an organizer I forgot to add.


bridgetkromhout:
Looks like email and Slack are already done.

PR #568, opened 2016-06-10
add monitis as t-shirt sponsor

pmoosh:


data/events/2016-siliconvalley.yml
```diff
@@ -37,6 +37,10 @@ sponsors: # List all of your sponsors here along with what level of sponsorship
     level: silver
   - id: extrahop
     level: silver
+  - id: 2015-pagerduty
```
bridgetkromhout: Can you put this in as pagerduty, not 2015-pagerduty? The 2015 one has the blue border hard-coded.

data/events/2016-siliconvalley.yml
```diff
@@ -37,6 +37,10 @@ sponsors: # List all of your sponsors here along with what level of sponsorship
...
+    level: silver
+  - id: 2016-monitis
```
bridgetkromhout: Please add the new sponsor as just monitis. We're trying to move away from the dates in favor of a simple one and then updating it as needed with `-before-datestamp`.

bridgetkromhout:
I would prefer monitis not be labeled as 2016, but I'm merging this in the interests of getting it done. Probably will clean up later if/when we do the rest.

pmoosh:
I do whatever needs to be done. I didn't use filenames with prefixes, until I saw them appearing. So if they are not necessary, I will stop using them. Thank you for the merge

PR #579, opened 2016-06-13
[DC 2016] Closing out DC 2016

markcornick:
This advice in content/data/2016-washington-dc.yml appears to be
incorrect with regard to the quotes:

    year: "2016" # The year of the event. Make sure it is in quotes.

With quotes there, the event does not appear in the "past" partial.
Without quotes, it does. Thus I've removed the quotes, but feel free to
tell me how to fix this another way if that breaks anything.

### DIFF
--- a/content/events/2016-washington-dc/location.md
+++ b/content/events/2016-washington-dc/location.md
@@ -5,7 +5,7 @@ type = "event"
 
 +++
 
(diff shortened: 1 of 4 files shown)


mattstratton:
Yeah, this is a defect. Sorry about that :)

PR #581, opened 2016-06-14
Ohio 2016: Setting Dates

kjenkins19:
Finally have dates for Ohio 2016! Added in location, Event, CFP dates, proposal links.


mattstratton:
Just noticed that the data file for ohio 2016 is using some of the old-style format (for example, team listing and navigation). You might want to update it (the old format IS functional, of course).

Example - https://github.com/devopsdays/devopsdays-web/blob/master/yyyy-city.yml

PR #620, opened 2016-06-20
CFP dates & Sponsor updates for Raleigh 2016

mrry550:


bridgetkromhout:
The sponsor image needs to be square to look good, and it needs to be a png to show up at all with the way the site is coded. See https://github.com/devopsdays/devopsdays-web#adding-sponsors for details.

I suggest trying to do a local build; that's how I can see if stuff works before I submit it. This is what it looks like when I build your fork locally:

screen shot 2016-06-20 at 4 20 47 pm

So, to fix: crop that sponsor image so it's a square, and export it as a png. git rm the jpg and add the png.

mattstratton:
Please also ensure that the sponsor logos are 200 x 200 px so they properly retina-fy :)

mrry550:
@mattstratton @bridgetkromhout I think I got most of the fixes. Looks good on my end, but let me know. Vendors aren't giving me 200x200, but I will press the team harder to get better graphics for OmniTI & Techtown training. I think we need to get the sponsor info up though since they have paid.

PR #627, opened 2016-06-21
[Boise 2016] CFP announcement, sponsors & committee updates

tresni:


phrawzty:
Hello,

Overall this looks fine; however, if you could kind resize the sponsor logos to [200x200](https://github.com/devopsdays/devopsdays-web/blame/master/README.md#L71), that'd be splendid. Thanks!

bridgetkromhout:
> Overall this looks fine; however, if you could kind resize the sponsor logos to 200x200, that'd be splendid. 

While we're in favor of better-looking logos, that's not a reason to prevent an otherwise-acceptable merge.

Issue #641, opened 2016-06-26
Bad file/dir names stop cloning of repo on Windows


glennsarti:
Unfortunately the directory at `static/events/2013-atlanta/proposals/The DevOps Pay Raise: Quantifying Your Value to Move Up the Ladder` contains an invalid filename character (`:`) which means the devopsdays-web repo can never be cloned on a Windows system. (or at the very least checkout the current HEAD of  29d6da2) 

Possible solution: Rename the file to something that does not include invalid Windows characters.  While this means Windows users will not be able to checkout previous commits, at least they will be able to checkout master HEAD.

Ref: Windows Naming Conventions
https://msdn.microsoft.com/en-us/library/windows/desktop/aa365247(v=vs.85).aspx

glennsarti:
It is unclear the from README whether this repo should be able to used on Windows.  Given that Hugo is cross platform I assumed Windows would be a "supported", or at least, allowed platform to use.

glennsarti:
I audit the repo for bad file names.  Result is below;

Just over 100 files are bad.

```Error: File name is incompatible to windows file systems.
/static/events/2013-atlanta/proposals/Don’t Fear the DevOps/index.html
Error: nonprintablechars= 3
Error: File name is incompatible to windows file systems.
/static/events/2013-atlanta/proposals/The DevOps Pay Raise: Quantifying Your Value to Move Up the Ladder/index.html
Error: illegalchars= 116
Error: File name is incompatible to windows file systems.
/static/events/2013-atlanta/proposals/Your Agile Performance Testing Checklist: MultiMechanize, Bees with Machine Guns, PageSpeed and AppDynamics/index.html
Error: illegalchars= 157
Error: File name is incompatible to windows file systems.
/static/events/2013-austin/proposals/Automating SQL Server Maintenance Plans with Ola’s Scripts/index.html
Error: nonprintablechars= 3
Error: File name is incompatible to windows file systems.
/static/events/2013-austin/proposals/DevStack: Learn OpenStack by running OpenStack/index.html
Error: illegalchars= 95
Error: File name is incompatible to windows file systems.
/static/events/2013-austin/proposals/Goodbye Hello World: How a Fully Provisioned Demo Application Helps All Our Rebels/index.html
Error: illegalchars= 131
Error: File name is incompatible to windows file systems.
[...]

PR #643, opened 2016-06-26
(GH-641) Rename static files which have invalid Windows names

glennsarti:
This PR renames directories, replacing characters which are invalid on Windows filesystem with underscore.  This PR also modifies references to the renamed directories so that the website links are still active.


bridgetkromhout:
> This PR also modifies references to the renamed directories

For an accepted talk, we have to assume there are external links to the talk detail page out from the wider internet. @mattstratton can weigh in on this, but IMO it's not acceptable to break a past event's talk detail page(s) just to prevent a few current event's maintainers from having to run their tests inside docker. And since we can't (I don't think) do redirects for static pages, this isn't going to solve that.

glennsarti:
@bridgetkromhout So far, the only files that have been renamed are within the proposals directories for each event, not the actually accepted talks.  At least I assume that by the name proposal, not program.  I'm not sure there'd be external links the individual talks.

glennsarti:
> ... maintainers from having to run their tests inside docker

I'm confused by the reference to docker.  There's nothing in the README about docker, and I don't see how docker would solve the issue of being able to clone this repo on a Windows filesystem.

glennsarti:
Confirmed I am now able to check out this repo on windows
```
C:\>git clone https://github.com/glennsarti/devopsdays-web.git dod -b ticket/maint/gh-641
Cloning into 'dod'...
remote: Counting objects: 26814, done.
remote: Compressing objects: 100% (352/352), done.
Receiving objects: 100% (26814/26814), 121.85 MiBemote: Total 26814 (delta 149), reused 2 (delta 2), pack-reused 26457
 | 3.25 MiB/s, done.
Resolving deltas: 100% (11661/11661), done.
Checking connectivity... done.
Checking out files: 100% (6871/6871), done.

C:\>
```

glennsarti:
It may be possible to create redirects using aliases
https://gohugo.io/extras/aliases/

For each bad file...
* Move the offending file into the content directory
* Rename it with a windows compatible name
* Inside the front matter setup an alias to its non-Windows name

This would allow the repo to be cloned, however Hugo would not be able to generate the website on disk (still bound be filesystem restrictions).  It _MAY_ be able to serve it from memory (default) though.

bridgetkromhout:
> I'm confused by the reference to docker. There's nothing in the README about docker, and I don't see how docker would solve the issue of being able to clone this repo on a Windows filesystem.

If someone's doing a docker run of an ubuntu image and then a docker clone there, they're not limited by what windows filesystems can handle, since they're on linux.

>So far, the only files that have been renamed are within the proposals directories for each event, not the actual accepted talks. At least I assume that by the name proposal, not program. I'm not sure there'd be external links to the individual talks.

Historically, when a talk was accepted it would link to the original proposal file. Here's an example:

http://bridgetkromhout.com/speaking/2015/devopsdays-newyork/

links to:

http://www.devopsdays.org/events/2015-newyork/proposals/LongDistanceDevops/

or

http://bridgetkromhout.com/speaking/2014/devopsdays-belgium/

links to a legacy static page but it's still not broken:

http://www.devopsdays.org/events/2014-belgium/proposals/ceci-nest-pas-devops/

That's the kind of behavior I don't want to lose.

bridgetkromhout:
I'll need @mattstratton to look into it but as far as I know...

> It may be possible to create redirects using aliases

...we can't do this with the static files.

mattstratton:
Correct. They would need to be migrated into markdown in the content directory. 

I have an almost-done script to do this.

glennsarti:
@mattstratton @bridgetkromhout I suppose a bigger question should perhaps be answered first...

`As a Windows User should I expect to be able to contribute to this project with Windows?`
If the answer is No, then all this is moot.  I would be happy to raise a PR to update the README to reflect that.

I feel that moving from Webby to Hugo would be the perfect opportunity to fix this issue.

glennsarti:
> I have an almost-done script to do this.

@mattstratton In that case, I can hold off on this PR and close it in preference to yours.  I'd be happy to test it for you from a Windows perspective.

mattstratton:
I don't think that Windows should be a blocker on being able to contribute. However, we have to fix the migration issue that will resolve this. I am without laptop and Internet for the next couple of days, but I expect to be able to close out my migration script by the weekend.

glennsarti:
Ok....I'll park my PR for the moment.

mattstratton:
Now, one thing that would make it work is the Docker workaround. Given that Docker on Windows is a thing, if we had a nice clean docker image for this project (like we had with webby) then we could guarantee a consistent experience for everyone. It's also a good idea because we can ensure that the docker experience uses the same version of Hugo as the prod build system.

glennsarti:
Using docker would solve some problems, unfortunately I don't feel it is an adequate workaround for this particular issue (i.e. file system).

Somewhere, somehow, this repo has to live on disk. In a Windows-on-docker environment this would be either in the container itself or on the docker host.  Being able to modify those from a Windows host is far from trivial, unless I'm missing something.

Sure we can share the filesystem between container and client, but no matter what, the invalid directory names can never be read on Windows system as they contain invalid characters.  The Windows filesystem APIs will error e.g.
Accessing files on an Ubuntu Samba server with colons in the name, with mangled names turned off
https://bugs.launchpad.net/ubuntu/+source/samba/+bug/676304

bridgetkromhout:
> Being able to modify those from a Windows client is far from trivial, unless I'm missing something.

Ah, that's a requirement that didn't present itself in my mind. The way this went in my mind was
* person has windows locally
* person runs docker, giving themselves a linux shell, and happily edits files with vim or whatever
* everything is now fine because we did an end-run around windows.

If the problem we're trying to solve is "people want to use windows tools directly" I'm skeptical (seen too many NOEOL). But, this is Stratton's area of expertise, not mine. Up to him to make it work if it's even possible.

mattstratton:
We will be able to solve for the path name problems (which we need to solve for other reasons besides just the Windows scenario) when I complete the migration of the `static` files into proper Hugo markdown files.

I know i keep promising this Real Soon Now, but I have some cycles to devote to it, and I am hopeful to complete it REALLY Soon Now. When this is done, the actual filenames on disk will NOT have spaces or specials in them, but the URL's with the specials/spaces will be set as aliases for them, so the URL's will still work.

With that in mind, I am closing this PR without merging it.

PR #659, opened 2016-06-29
Added Hack The Gap and Dynatrace logo

zruty:


bridgetkromhout:
Let's hold for the dynatrace yaml file too.

PR #686, opened 2016-07-05
PDX: Add dates to the day numbers

edunham:
I was linked straight to the schedule page and wanted to know what days the talks were on. This information was previously only available on the "welcome" page.

### DIFF
--- a/content/events/2016-portland/program.md
+++ b/content/events/2016-portland/program.md
@@ -19,7 +19,7 @@ type = "event"
   </div>
 </div>
 
-## Day 1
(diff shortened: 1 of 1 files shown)


jyee:
Thanks! We were trying to get the info up as quickly as possible. I've got plans to convert the basic schedule into a responsive grid form with dates and times... and I should have that done in the the next day or two.

jyee:
Thanks again. I've rolled out the first pass at the full schedule and included dates along with the Day 1 and 2 headings.

PR #706, opened 2016-07-09
Consolidate Chef sponsorship settings

nathenharvey:
* Use the logo without the URL subtitle
* Add the twitter handle to the sponsor config


bridgetkromhout:
Hi, Nathen -

We aren't accepting retcons of the past (see the discussion when this came up before, in https://github.com/devopsdays/devopsdays-web/pull/503). If some past events used a sub-optimal logo it's unfortunate, but should have been fixed then, and it's not up to us to change it retroactively now for those past events.

What would be fine is a PR changing chef-io to chef-before-20160709 or some such, so that people know not to use it in the future.

You could also put in separate PRs for each the future events and CC the people who've been doing the updates for those events so they know what's up. I want them to know it's going on so in case they are preparing signage, etc, they aren't caught unawares and they know what the change is for their event.

Lastly, I have no idea what adding the twitter handle to the sponsor data file is meant to accomplish, as it isn't something we're querying currently, but I guess it's fine in case you want to write some code that uses it!

nathenharvey:
Both logos are fine and almost identical. I was mainly trying to prevent confusion moving forward. "Should I use chef or chef-io?  What's the difference?"  Your suggestion of renaming should help with that so I can update this PR with that. 

I was also thinking that the Twitter handle would be helpful for other organizers. As we start getting new sponsors, we want to thank them on Twitter. If they've sponsored in the past, finding the Twitter handle is as easy as reusing their logo. There may *also* be some future programmatic use for that data. 

I happy to remove it but was thinking it might be of use to other organizers.

bridgetkromhout:
The twitter handle is fine. We definitely have that data at Minneapolis for a lot of sponsors, and perhaps @zruty would want to follow your lead and add it to the data files. (I just didn't want you adding it hoping it would currently lead to anything new being displayed, as it won't.)

bridgetkromhout:
Hi Nathen -

When you recover from chef conf, please update this PR as discussed.

nathenharvey:
@bridgetkromhout this is now a change that simply changes the files that should be deprecated in preparation for having only one active Chef configuration.

bridgetkromhout:
Great - thanks!

PR #717, opened 2016-07-12
Replaced Optum Logo

zruty:


kmugrage:
Past events have used the old logo - you may want to check out #503 for a discussion on this. 

The new logo needs to be namespaced in such a way that it won't rewrite past events. It should also be 200px x 200px.

bridgetkromhout:
Right. Let's get this changed. Closing this one without merging.

PR #731, opened 2016-07-18
These are the first Proposals for DevOpsDays Berlin 2016

range:
* Added "Proposals" to the navigation menu
* Added proposals/ directory for incoming proposals
* Updated proposals.md
* Added first three proposals.

### DIFF
--- a/content/events/2016-berlin/proposals.md
+++ b/content/events/2016-berlin/proposals.md
@@ -1,9 +1,19 @@
 +++
-date = "2016-07-08T18:12:13+02:00"
-title = "proposals"
+City = "Berlin"
+Year = "2016"
+heading = "devopsdays Berlin - Proposals"
+title = "Proposals"
 type = "event"
+tags = ["berlin","berlin-2016"]
 +++
-  {{< cfp_dates >}}
 
-<hr>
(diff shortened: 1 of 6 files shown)


phrawzty:
Hello,

In the past we supported the idea of tracking proposals during the CFP period - this is no longer the case, which is why the proposals section is no longer part of the standard template.  It caused issues in the past with people gaming the submissions system, stealing ideas, and generally generating ill will. 😢 

If you're interested in getting feedback about the proposals, be aware that this has never been successful in the DOD community - which is another reason that this element has been retired.

Finally, If you're looking for a way to track your proposals, you could use Google docs (for example), or a third-party CFP service of some type.

If you really, *really* want to do this, I won't stop you - but it is actively discouraged on our end.

phrawzty:
Oh, as an aside, if you'd like to join the official DOD Slack group, we'd be more than happy to discuss this and any other practice, policy, event, or website topic. 😄

range:
Okay, let me think about this. I'll let this pull request simmer for a day more, though. Thanks for the explanation, I was actually wondering why I couldn't find any "examples" from the last events in the tree.

range:
Okay, I'll close, we'll use Google Docs or similar.

PR #737, opened 2016-07-20
Max sponsors

jyee:
Got around to addressing the need to control the sponsor level CTA per issue #93 

Basically using $.Scratch to setup a counter, so we can count how many sponsors get iterated per level, then comparing it to the level "max" property (if it exists or is non-zero)


bridgetkromhout:
Thanks. I've not had a chance to test yet, but I appreciate going this direction.

bridgetkromhout:
I've tested this. It works like it says on the tin.

PR #743, opened 2016-07-22
allow special/international characters in speaker name

jerrycattell:
LinkTitle defaults to Title unless linktitle is specified in the md file


mattstratton:
Hey @jerrycattell - I just noticed this. Can you document it (at least in the code, if not also in the `README`) of what it does exactly?

PR #741, opened 2016-07-21
Updating Elastic logo (font change)

jyee:
Updating the Elastic logo. They've brightened their colors and changed the font.

PR #744, opened 2016-07-23
update elastic logo

jerrycattell:
I mistakenly used an old logo when updating to a 200x200


jyee:
For reference it looks like the change was made in [b68f1da2 (May 5)](https://github.com/devopsdays/devopsdays-web/commits/master/static/img/sponsors/elastic.png). I'm going to close out #741 and merge this, since this corrects a mistake (not actually updating the logo to a new version).

PR #766, opened 2016-07-26
Add displayname functionality

mattstratton:
This PR adds support for the (optional) `displayname` parameter in the event data file. If the `displayname` is set, this is what will be shown when the event is referenced (on the navigation menus, as well as on the main headers for the event page).

You should use the `displayname` parameter if your city has non-URL friendly characters, such as accented characters or spaces (i.e., "Zürich" or "Salt Lake City"). The name parameter should be set as `2016-zurich` or `2016-salt-lake-city` for example. Please also make sure that the "city" parameter is set with non-accented or space characters in this case as well.


phrawzty:
👍💯

PR #767, opened 2016-07-26
Update zurich with displayname

mattstratton:


mattstratton:
This updates the Zurich event to use the friendly pathing but the `displayname` parameter to show the city with the accent.

Issue #742, opened 2016-07-22
should remove the unused themes


mattstratton:
`themes/devopsdays` and `themes/devopsdays-legacy` are not being used anymore and we should delete them from the repo (but not from history) to minimized confusion.

bridgetkromhout:
Agreed. Every once in a while I dig things out of there, but that's all. Let's whack'm.

PR #769, opened 2016-07-26
Remove unused themes from project

mattstratton:
The `devopsdays` and `devopsdays-legacy` themes are no longer used, so I have removed them to avoid confusion.


mattstratton:
Fixes #742

PR #803, opened 2016-08-05
Add aliases for Baltimore 2016

nathenharvey:
The conference was postponed until 2017.

This change makes all of the old 2016 URLs redirect to the 2017 pages.

For example, /events/2016-baltimore/welcome/ now redirects to /events/2017-baltimore/welcome/.


kmugrage:
This 'works for me' when run locally. The event shows up with the correct dates on /events/, urls including 2016-baltimore appear to be redirected to 2017-baltimore correctly.

PR #814, opened 2016-08-07
pushing out the NYC announce date to give speakers time to confirm

kdaniels:


pcn:
:+1: if one of the organizers could merge this, we're changing our schedule per @kdaniels here.

PR #822, opened 2016-08-09
2016-boston: probably final sponsors

masteinhauser:


masteinhauser:
Thanks in advance! Getting down to the final few weeks for us finally!

PR #828, opened 2016-08-10
[Boise 2016] Add OpenShift as silver sponsor

tresni:


mattstratton:
Deploy preview ready!

Built with commit 46a67699f09ee8d315162f521ddc972493616795

http://deploy-preview-828.paratrooper-sum-25346.netlify.com

tresni:
@mattstratton I think something is up with your build machine.  It doesn't properly set the root url for the build and is missing the image (even though I;ve verified locally it's in the right place and hugo serves it correctly.)

mattstratton:
Yeah, I noticed that too. Disregard for now.

PR #833, opened 2016-08-11
Allowing multiple-word cities to work with speakers/talks

bridgetkromhout:
Still doesn't handle special characters. Should fix via event slug. But at least this band-aids the current badness with slightly less badness.

Specifically, this allows speaker data to show up on speaker pages and talk pages if the city has a multi-word name. It doesn't introduce regression for single-word names as far as my checks show.


bridgetkromhout:
Merging for now to unblock @cobusbernard although @mattstratton may want to do this differently.

Issue #545, opened 2016-06-06
figure out a solution for non-square sponsor logos


mattstratton:
A lot of sponsor logos don't really adjust well to the 1:1 dimensions. Bikeshed away as to what we can do about this...

(note: the reason we currently constrain in the CSS to 100px x 100 px and request a 200px square image is to make with fancy Retina magic. If we were to opt out of this, there would have to be some type of solution to make them better in a responsive way...I believe that in order to get the high res you have to constrain both H and W, but not sure - @jyee?)

jyee:
We don't necessarily have to constrain them, but it's the easiest way to balance std + retina without having to futz with generating multiple images and doing image-set css or srcset html5.

I guess one of the other things we could do is build out the front-end pipeline (since we've already got LESS) and have it handle the image sizing and such for us.

mattstratton:
I already started thinking about building out the pipeline, mostly because I wanted to get `html-minify` in there (although the scads of invalid HTML in the old generated webby stuff is making it break terribly).

The trick is figuring out how to have the pipeline do this stuff but have a minimum of things people need to do locally. That is to say, I don't want people to have to use `grunt` or `gulp` just to submit changes.

mattstratton:
Hmm. [This grunt plugin](https://github.com/JrSchild/grunt-retinafy) seems like it would work. The thing that would be nice would be to only have to do this when building to publish, not locally. 

I think this would work, because when working locally, you'll just have `chef-2016.png` or whatever, and the CSS is going to say "okay, no retina image here, just use this one". But then when it's run through grunt, it will make the `chef-2016@2x` image.

Also, pasting this link to keep track of it for later: https://imulus.github.io/retinajs/

bridgetkromhout:
Does https://github.com/devopsdays/devopsdays-web/blob/master/utilities/resize_sponsor_logo.sh handle any/all of this? (I've not tried/vetted it at all.)

mattstratton:
Not really. This sorta became a image processing pipeline bikeshed, and what we really need to do is figure out how to not "cheat" to get retina by constraining them to certain ratios.

mattstratton:
The retina-fication is handled in the pipeline of [devopsdays-theme](https://github.com/devopsdays/devopsdays-theme) so this issue can be closed in this repo.

PR #846, opened 2016-08-13
[2016-chicago] update sponsors

jerrycattell:
add new Slalom and Atlassian logos

update AppDynamics logo to be 200x200


bridgetkromhout:
Hi, @jerrycattell - is there a reason to add a different Atlassian and Slalom? https://github.com/devopsdays/devopsdays-web/blob/master/static/img/sponsors/atlassian.png and https://github.com/devopsdays/devopsdays-web/blob/master/static/img/sponsors/slalom.png look the same as the ones you're adding.

jerrycattell:
The atlassian logo is 500x500 with transparent background (as opposed the white that most others have). The slalom logo is 100x100, with transparency on top/bottom of the slalom with white background. I'm not opposed to just replacing those images with 200x200 logos.

bridgetkromhout:
Let's ping @mattstratton because I am not sure what he intends with the transparent vs white. (Maybe eventually we move away from logos needing to be square?)

mattstratton:
Tbh we are just lucky with the white backgrounds right now. Transparent should be fine for now. 

WRT square logos, visit this bike shed over here --> https://github.com/devopsdays/devopsdays-web/issues/545

jerrycattell:
Regardless of transparency, we're standardizing on 200x200 for now, correct? Slalom is 100x100, so we'll definitely want to update that one. The Atlassian one is 500x500, but is actually smaller in size than the 200x200 one I generated, so I'm less concerned about that one.

As I said, I'm completely fine with redoing this but updating the slalom logo in place if that is preferred.

mattstratton:
Correct. 200x200. It's actually kind of important that it's exactly that size because retina.

jerrycattell:
@bridgetkromhout @mattstratton, I switched it up to just update the logos for AppDynamics, Atlassian, and Slalom to 200x200 instead of creating new sponsors.

bridgetkromhout:
Okay. I'll merge this and then I'm going to collapse appdynamics and 2016-appdynamics into one.

PR #883, opened 2016-08-18
[india-2016] change in event date by 1 day

abhishekkr:
due to logistic reasons it's now 4-5Nov from 5-6Nov


bridgetkromhout:
I see this was changed on your site as well.

PR #885, opened 2016-08-18
Refactor past and future to use date fields

### LINKED ISSUE #882: remove dependency on current and past parameters
Based upon what we learned from creating the cfp page, we should be able to do queries based on the event dates properly now, so we can whack all `current`, etc, references.

mattstratton:
Do not merge this yet - the /events page is displaying oddly,and it’s somehow due to the two New Yorks. If you change the startdate of New York (Winter) to be in 2012, everything lays out properly.

Fixes #882


mattstratton:
This also seems to depend on Hugo 0.16 to work at all.

PR #888, opened 2016-08-19
Add redirects to legacy Ghent

mattstratton:


mattstratton:
Note: this will not take effect on production until we have fully migrated off GH pages to Netlify.

Issue #879, opened 2016-08-18
consider adding CONTRIBUTING.md file to repo


mattstratton:
It may be a duplication of some of the content in the README, but I think having a `CONTRIBUTING.md` file that explains some of the nuance of contributions to the site (especially for shared code/design related items) would be beneficial. 

This is pretty standard for a lot of projects.

bridgetkromhout:
Fine by me but let's avoid two sources of truth. Let's _move_ the contributing guidelines, not allow them to drift from each other in two places.

mattstratton:
Good call. We can just link to the file in the README under "Contributing". The README is getting large anyway :)

phrawzty:
Note that fully one third of the major headers in the README contain the string `Contribut`. 😄  I guess the "easiest" thing would be to just move those sections into CONTRIBUTE, but at first blush I'm not seeing the value add there. 

Perhaps I'm thinking about this the wrong way. If you'll indulge me, let's ask the most basic of all questions: What problem are we trying to solve❓

mattstratton:
Two things:

1) It's pretty standard on a project to have a file called CONTRIBUTING, which is where I look to learn how to follow the process of contributing. The README is for users of the software, etc.

2) The README is getting really big, and contains a lot of details that aren't necessary for someone who just wants to "use" the software, (i.e., update it with their events). When we make the thing a giant "wall of text", then nobody reads anything.

bridgetkromhout:
Renamed (https://github.com/devopsdays/devopsdays-web/pull/896) and I think this is now complete.

mattstratton:
Fixed by #890

PR #890, opened 2016-08-19
move contribution info to CONTRIBUTE.md; #879

phrawzty:
This is a first pass at addressing #879 (CONTRIBUTE.md). I *basically* just split out the various Contrib-related portions to their own document. I say basically, because I also made some improvements (hopefully) to the documentation, too. 😄


phrawzty:
I can already see that GitHub renders Markup slightly differently than my local editor does; in particular, it would appear that code blocks break the numbering. 😠 

Will fix, or failing that, find another way...

mattstratton:
I'm a little concerned that for just "content" contributions that the pull request process isn't on the README now, but I think that it's okay.

PR #892, opened 2016-08-19
Ignore a script that updates the ulimit

nathenharvey:
I always have to copy-n-paste out ulimit fixes out of the README

I've added a local bash script that will do this for me but not sure if
everyone wants this in the repo.

Here's what the script contains, I'm happy to add it if there's
interest:

```
#!/bin/bash
sysctl -w kern.maxfiles=65536
sysctl -w kern.maxfilesperproc=65536
ulimit -n 65536 65536
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,3 +48,6 @@ $RECYCLE.BIN/
 *.lnk
 node_modules
 _mdwriter.cson
+
+# A script to update the ulimt for OS X
+increase_ulimit.sh


mattstratton:
Funny thing is, in probably less than a week, this won't matter anymore (we're in process of migrating the legacy static files out of the main repo, and that's what make ulimit crash).

nathenharvey:
yay!  I look forward to that day and will remove that script then 👍

Issue #860, opened 2016-08-16
update make_new_event.sh to support dashes in city names


mattstratton:
We prefer to have the city names (for the slug, not the value of `city`) use dashes if there are spaces in the city name (i.e., "Salt Lake City" should be `2016-salt-lake-city`.

The script needs the following:

- [ ] Prompt the user to enter their city name with dashes if it has spaces
- [ ] Add the ability to set a specify city name (which will be inserted as `displayname` in the data file). Have the default for this (if the user hits "enter") be what was set before in the city name prompt. This should prompt that they only need to do this if the city has spaces or special characters.
- [ ] Update to use `$cityabbr` instead of `$city` when setting paths/data file names

bridgetkromhout:
This is complete for the creation of new cities (in the new event script). We'll need to deal with in-flight cities on a case by case basis.

I solved this by allowing "city" to have special characters, and by urlizing "name".

bridgetkromhout:
Fixed in https://github.com/devopsdays/devopsdays-web/pull/912.

Issue #861, opened 2016-08-16
reconcile code that references "displayname" parameter


mattstratton:
When trying to solve for cities with special characters, there was some work intending to create a `displayname` parameter for events. 

This seems to have turned out to not be needed, as we handled it by allowing `city` to have special characters and spaces, etc. However, there are several places in the codebase the still refer to this. It's not breaking anything, but it's going to get confusing.

Here are the places I found:

https://github.com/devopsdays/devopsdays-web/blob/master/themes/devopsdays-responsive/layouts/partials/event_header.html#L14

https://github.com/devopsdays/devopsdays-web/blob/master/themes/devopsdays-responsive/layouts/partials/past.html#L33

https://github.com/devopsdays/devopsdays-web/blob/master/themes/devopsdays-responsive/layouts/partials/future.html#L29

https://github.com/devopsdays/devopsdays-web/blob/master/themes/devopsdays-responsive/layouts/partials/upcoming_headline.html#L8

https://github.com/devopsdays/devopsdays-web/blob/master/README.md

Since the value `displayname` doesn't exist in any of the data files, including cities with special characters, this doesn't seem to be needed after all.

mattstratton:
Actually, we might want to keep this, as the intention of this would (I think) solve for #858 properly. We just need to fix the two events that were created "incorrectly" (Zurich and Brasilia).

mattstratton:
Let's discuss it some more, but I think that the right answer is to keep the `displayname` code, and retro-fix the two events (Brasilia and Zurich) that are using it incorrectly.

We do need to fix the `make_new_event.sh` script to use this properly.

bridgetkromhout:
I'm solving this by getting rid of displayname and no longer parsing "city" (it's just for display).

bridgetkromhout:
Fixed in https://github.com/devopsdays/devopsdays-web/pull/912.

PR #912, opened 2016-08-21
update for spaces/special characters

bridgetkromhout:
As far as I can tell, I've removed "displayname" from being needed at all - I came to the opposite conclusion from what @mattstratton was saying in https://github.com/devopsdays/devopsdays-web/issues/861, after digging into this. People can now just set "City" as desired, and the new event script will handle it. The few places I was parsing the city value, I'm now parsing "name" to get the city slug.

It didn't turn out to be necessary to put the city slug in the data file. When updating the new event script as requested in https://github.com/devopsdays/devopsdays-web/issues/860, I kept it simple and parsed the entered name instead of asking the user for multiple variations.

I also think I can probably remove "friendly" as a data file element from the sample data file; it's not used, although in one place a variable is also called that.

I'm leaving "current" in the data file for now until it's totally gone and https://github.com/devopsdays/devopsdays-web/issues/882 is resolved.


mattstratton:
Makes a lot of sense to me. 

`future.html` needs a tweak for this too I think. It has similar conditionals on `displayname` (although that's just hygiene; sun setting that parameter won't break the code)

bridgetkromhout:
> future.html needs a tweak for this too I think. It has similar conditionals on displayname (although that's just hygiene; sun setting that parameter won't break the code)

Yeah, I know - I was just too tired to resolve the merge conflict last night. :)

bridgetkromhout:
When merged:

Fixes https://github.com/devopsdays/devopsdays-web/issues/860

PR #917, opened 2016-08-21
[SLC] Change organizer info

davidasnider:
Removed old organizer, updated twitter/workplace for another.

### DIFF
--- a/data/events/2016-saltlakecity.yml
+++ b/data/events/2016-saltlakecity.yml
@@ -43,7 +43,8 @@ team_members:
     employer: "EMC"
   - name: "Thomas Cook"
   - name: "Mike Duffy"
-  - name: "Alex Lutz"
+    twitter: "michaeljduffy"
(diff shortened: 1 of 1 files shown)


bridgetkromhout:
Hey, @davidasnider - was Alex an active SLC organizer for 2016? If so, we shouldn't remove him from 2016. You'll submit a new event PR for 2017 if you're going to do it again, and at that point you can change out your list of people. (We don't want to rewrite history!)

mattstratton:
I agree and should have asked before i merged this - if Alex was an organizer of 2016 then the name shouldn't be removed. The Twitter/etc of a person can be retconned if they want; I don't want to rewrite history either, but I can see why a person might want that.

davidasnider:
Yep, thanks for checking. Alex ended up not participating as an organizer. Since I was adding something for Mike, I took care of that update (Alex) as well.

bridgetkromhout:
Great; thanks!

PR #924, opened 2016-08-21
Revised and moved README info.

bridgetkromhout:
This is a pretty significant rewrite of the main README with a split to place some of it in a utils-specific README. I'd like @mattstratton to read/revise/merge this because of the major changes.

My guiding principle here was automation instead of checklists. Instead of having organizers try to parse and implement the necessary steps, I'm putting them into code.

(It's debatable how readable bash is, but English is ambiguous too.)

Also, given how essential the frontmatter stuff is, I think we probably want to strongly discourage manual event creation. This means keeping an eye on new events seeded from their previous year, too.


mattstratton:
Do the scripts (now that they are in the utilities folder) assume they are run from that working directory? Do we need to tell people where to make sure they run the script from or do you think it's clear?

bridgetkromhout:
> Do the scripts (now that they are in the utilities folder) assume they are run from that working directory? Do we need to tell people where to make sure they run the script from or do you think it's clear?

Right, they need to be run from the directory they're located in, as currently written. I didn't write any logic to try to guess where on the filesystem the CWD is in relationship to the repo. I think it's pretty obvious but feel free to edit as makes sense to you.

mattstratton:
As long as they expect to be run from where they are located, then I think it's clear :)

PR #931, opened 2016-08-22
Fix shortcode to use SSL for maps

### LINKED ISSUE #899: Fix mixed-mode issues on https
when https://www.devopsdays.org is loaded, it has a mixed mode issue because of this resource:

`http://maps.gstatic.com/intl/en_us/mapfiles/drag_cross_67_16.png`

This is coming from Google, but I cannot figure out how to get it to load the SSL version of that. The maps API call seems to be configured properly for SSL. 

This is a blocker for #642

mattstratton:


mattstratton:
OK, this didn't fix the issue. It should be merged because it needs to be done, but it didn't fix it.

PR #939, opened 2016-08-23
Script cleanup

glasnt:
Some basic doc cleanup based on me trying to get this working locally. 

Note, for the `make_new_event.sh` changes, I changed the copy locations based on running the script from the top level directly, as the `cp` destinations were correct if that's the place we are assuming the script is run from.

Please let me know if you require any alterations

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -16,7 +16,7 @@ If you'd like to edit a specific devopsdays event site (and/or contribute code),
 To watch for changes and rebuild on the fly, open a new terminal, change directories to your fork of the repo, and execute the following:
 
 ```
-hugo server -w --baseUrl="http://localhost:1313"`.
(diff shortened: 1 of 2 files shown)


mattstratton:
Looking to @bridgetkromhout to weigh in on this one, since she wrote the scripts :)

bridgetkromhout:
Mostly fine although most things in that script assume that you're running it from the dir it lives in, so I'll correct that after merging. Thanks for beta-testing the stuff I just rearranged this weekend. :)

glasnt:
@bridgetkromhout thanks! I didn't do a full archeological dig, so thanks for the context. Glad it's working now!

bridgetkromhout:
Thanks. Yeah, I just moved all that stuff around and didn't update that particular script; thanks for noticing. :)

PR #953, opened 2016-08-25
Philly: Adding Speakers.

peterjshan:
Do not merge please.

Update: This should be ready.


peterjshan:
@bridgetkromhout I think this is good enough, I think we're g2g.

bridgetkromhout:
👍

PR #959, opened 2016-08-25
[Boise 2016] Add Corey Quinn to speaker

tresni:
Also remove J Paul Reed


tresni:
Realized I'd missed that very important part of having him on the speaker list :)

PR #961, opened 2016-08-26
Add event for Austin 2017

scottsbaldwin:


bridgetkromhout:
Set up the 2017 email lists.

PR #976, opened 2016-08-27
Ohio 2016: Adding 3 gold sponsors, 1 media, and updating organizer list.

kjenkins19:
Added 3 new sponsor logos: ICC, RoundTowerTechnologies, and Techlife Columbus
Created a small section for History Local Organizers per request of Warner. I would not mind refactoring this out to something more standard at a later date (Or if there is one, please point it out), but this might just be a 1 off. I used the same terms currently used for the core.


bridgetkromhout:
While you can put what you want on your contact page, most events change out their organizers as needed. If someone was on your team in 2015 and isn't on the team in 2016, then their past participation is still listed on the 2015 page, and it doesn't make much sense to list them on 2016.

The core team is the only one that uses active/not, simply because they don't have event years to delimit participation.

Issue #799, opened 2016-08-04
add structured google search data for event pages


mattstratton:
Should be fairly trivial to add the functionality which will improve the appearance of events in Google search results.

https://developers.google.com/search/docs/guides/search-gallery

mattstratton:
More background - https://developers.google.com/search/docs/data-types/events

mattstratton:
Closed in #993

PR #993, opened 2016-08-29
Add basic structured metadata to event pages for Google

esigler:
Howdy!

This change puts in the basics for JSON-LD metadata on an event page, with
a little bit of logic for if the city has been duplicated into the location.
Concatenating the image URL together took far longer than I want to admit.

Are there other types of metadata that y'all would want included?  If the
address was available in the event yaml, that might surface up better in 
search results?

Example output on what would be https://www.devopsdays.org/events/2016-chicago/welcome/:

```
<script type="application/ld+json">
{
  "@context": "http://schema.org",
  "@type": "ExhibitionEvent",
  "name": "DevOpsDays Chicago 2016",
  "startDate": "2016-08-30",
  "endDate": "2016-08-31",
  "url": "https://devopsdays.org/events/2016-chicago/welcome/",
  "image": "https://devopsdays.org/events/2016-chicago/logo.png",
  "location": {
    "@type": "Place",
    "name": "Summit West, Chicago",
    "address": "41.882219, -87.640530"
  }
}
</script>
```

Addresses https://github.com/devopsdays/devopsdays-web/issues/799

### DIFF
--- a/themes/devopsdays-responsive/layouts/event/single.html
+++ b/themes/devopsdays-responsive/layouts/event/single.html
@@ -5,5 +5,6 @@
 </div> <!-- closes the col-md-8 div from event_header -->
 {{ partial "sponsors.html" . }}
 </div> <!-- closes the row div -->
+{{- partial "event_metadata.html" . -}}
 
 {{ partial "footer.html" . }}
--- a/themes/devopsdays-responsive/layouts/partials/event_metadata.html
+++ b/themes/devopsdays-responsive/layouts/partials/event_metadata.html
@@ -1,0 +1,26 @@
+{{/* site data query copypasta from event/single.html, per mattstratton's note - esigler */}}
+{{ $path := split $.Source.File.Path "/" }}
+{{ $event_slug := index $path 1 }}
+{{ $e :=  (index $.Site.Data.events $event_slug) }}
+{{/* end site data query */}}
+
+<script type="application/ld+json">
+{
+  "@context": "http://schema.org",
+  "@type": "ExhibitionEvent",
+  "name": "DevOpsDays {{ $e.city }} {{ $e.year }}",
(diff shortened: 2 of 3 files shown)


bridgetkromhout:
Thanks, Eric! @mattstratton wants to look at this and he's right in the midst of setting up for devopsdays Chicago, so you'll hear back after that's over.

mattstratton:
When I run this through the structured testing tool, one of the warnings that comes up is:

`The description field is recommended. Please provide a value if available.`

@esigler do you think it would be easy to add that in and then re-submit the PR? I think that the description can be pretty simple for right now - or alternately, we could create a new issue to add a better event description metadata option.

esigler:
@mattstratton I was wondering what the best choice was for `description`. I'm a tiny bit worried that all events having the same `description` could be seen as spammy, but I might be paranoid.  I've pushed up a branch with a description field & default text in the template, and adds it to the example yaml.  Example output is now:

```
<script type="application/ld+json">
{
  "@context": "http://schema.org",
  "@type": "ExhibitionEvent",
  
    "description": "The conference that brings development and operations together.",
  
  "name": "DevOpsDays Chicago 2016",
  "startDate": "2016-08-30",
  "endDate": "2016-08-31",
  "url": "http://devopsdays.org/events/2016-chicago/welcome/",
  "image": "http://devopsdays.org/events/2016-chicago/logo.png",
  "location": {
    "@type": "Place",
    
      "name": "Summit West, Chicago",
    
    "address": "41.882219, -87.640530"
  }
}
</script>
```

& validated that if an event yaml doc contains `description` with a value, it populates instead of the default text.

mattstratton:
I think this should be fine - we can probably add a feature request in the future to construct the "default" to somehow include the city name, but for now it works.

PR #1048, opened 2016-09-06
confirmed singapore speakers

sergiubodiu:
Update DevOpsDays Singapore program


phrawzty:
Hello!

I'm not sure about the "About" sections in the talk descriptions - that information should be populated from the speaker bios. Why are you including it manually?

phrawzty:
To clarify, you're not obligated to use the speaker bios, but that's what they're there for. If you'd like some help with that feature, please don't hesitate to ask - we'd be more than happy to help. 😄 If you'd prefer to include the bio information in the talks directly, that's OK too.

Just let us know how you'd like to proceed and we'll go for it. 😄

bridgetkromhout:
As @phrawzty mentions, there is info in https://github.com/devopsdays/devopsdays-web/blob/master/utilities/README.md#speakers about how to use the data-driven speaker bio files which then allow for a speaker page you don't need to update manually.

phrawzty:
I understand that it's probably more important to get the information on the live site than to deal with technical details, so I'll go ahead and merge this now. As I mentioned above, you're welcome to stay with this model for this event, or to use the dynamic model, as you like.

sergiubodiu:
Noted, I will check the templates, thanks for merging @phrawzty

PR #1054, opened 2016-09-07
[berlin-2016] add platinum sponsor thoughtworks

nmesstorff:


phrawzty:
p.s. thanks for tagging the commit message with `[berlin-2016]` - it really makes searching through history easier. 👍

PR #1068, opened 2016-09-07
Set the color of the bikeshed

mattstratton:


mattstratton:
This is quite a silly PR. I shall close it.

PR #1072, opened 2016-09-08
Add powershell script for dockering all the things

### LINKED ISSUE #1063: get the hugo docker script to work on Windows
There is now a script that will load all the hugo goodness in a docker container [here](https://github.com/devopsdays/devopsdays-web/blob/master/hugoserver.sh).

We need to turn this into a functional PowerShell script that will work on a Windows machine running Docker.

I've made a couple passes at it, but I'm having some issues translating some of the bash stuff into posh. Anyone want to take a swing?

mattstratton:
Fixes #1063

### DIFF
--- a/hugoserver.ps1
+++ b/hugoserver.ps1
@@ -1,7 +1,6 @@
 # You will need to make sure you have your C drive (or whatever drive you have the devopsdays code on) shared in Docker
 
 $MyPath = (Get-Item -Path ".\" -Verbose).FullName
-Write-Host $MyPath
 
 docker stop hugo-server
 docker rm hugo-server 
--- a/hugoserver.ps1
+++ b/hugoserver.ps1
@@ -1,6 +1,6 @@
 # You will need to make sure you have your C drive (or whatever drive you have the devopsdays code on) shared in Docker
 
-$MyPath = $MyPath = $PSScriptRoot
+$MyPath = $PSScriptRoot
 
 docker stop hugo-server
 docker rm hugo-server 
--- a/hugoserver.ps1
+++ b/hugoserver.ps1
@@ -1,0 +1,8 @@
+# You will need to make sure you have your C drive (or whatever drive you have the devopsdays code on) shared in Docker
+
+$MyPath = (Get-Item -Path ".\" -Verbose).FullName
+Write-Host $MyPath
+
+docker stop hugo-server
+docker rm hugo-server 
+docker run -p 1313:1313 -v ${MyPath}:/site --name hugo-server devopsdays/docker-hugo-server:v0.16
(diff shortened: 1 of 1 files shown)


glennsarti:
@mattstratton Unfortunately I am unable to test this PR on DfW due to https://forums.docker.com/t/failed-to-start-vm-mobylinuxvm-the-vm-couldnt-get-an-ip-address-after-60-tries/19551/6

In my opinion this does not fix #1063 .  Installing Docker is an extremely heavy handed workaround for a simple backslash/forward slash issue.

hugoserver.ps1
```diff
@@ -0,0 +1,8 @@
+# You will need to make sure you have your C drive (or whatever drive you have the devopsdays code on) shared in Docker
+
+$MyPath = (Get-Item -Path ".\" -Verbose).FullName
```
glennsarti: This will get the path to the current working directory, not the root of the hugo website.  Would instead use `$MyPath = $PSScriptRoot`

Note that `$PSScriptRoot` is in Powershell 3.0 only, but as Docker for Windows is only supported on Windows 10 this is not an issue.

hugoserver.ps1
```diff
@@ -0,0 +1,8 @@
...
+Write-Host $MyPath
```
glennsarti: Try not to use `Write-Host` as it assumes an interactive console.  Instead use `Write-Verbose` e.g.
`Write-Verbose "Hugo directory is $MyPath"`

hugoserver.ps1
```diff
@@ -0,0 +1,8 @@
...
+
+docker stop hugo-server
+docker rm hugo-server 
+docker run -p 1313:1313 -v ${MyPath}:/site --name hugo-server devopsdays/docker-hugo-server:v0.16
```
glennsarti: Need to be careful of paths with spaces in them for `-v ${MyPath}:/site`

mattstratton:
For now, this is the workaround. We cannot assume that everyone is going to defensively code against the slashes any time they are looking to use them. The workaround is using Docker.

The write-host piece is actually a troubleshooting/debugging thing I put in, and there's no need to keep it in there :) I'll remove that part.

hugoserver.ps1
mattstratton: The script is run within the root of the hugo website, so it should work...

glennsarti: That's an assumption.  You can run that script from any working directory via an absolute path.

bridgetkromhout:
Right now @mattstratton is otherwise occupied and will get back to this when he's able.

glennsarti:
@bridgetkromhout Do you have an ETA or is this more of an open-ended thing?

mattstratton:
I'm getting married on 10/1. I will be back at "work" by 10/6.

hugoserver.ps1
mattstratton: Can you check on the change I just made and see if that looks right now?

mattstratton:
I'm going to go ahead and merge this since it's just the one script, and it seems to work...

PR #1079, opened 2016-09-08
Raleigh Updates

mrry550:
Updated Sponsor page, program page and updated the Thoughtworks logo (we apparently have an outdated one according to our contact).


mattstratton:
If the ThoughtWorks-products file was used by other events, please create a new sponsor file so as to not rewrite history.

bridgetkromhout:
> If the ThoughtWorks-products file was used by other events, please create a new sponsor file so as to not rewrite history.

I'll merge this and then deal with that.

Issue #989, opened 2016-08-29
Many shortcode templates are failing due to path seperator failure


glennsarti:
There are many places within the codebase that use a section such as

```
{{ $path := split .Page.Source.File.Path "/" }}
```

Unfortunately in Hugo `File.Path` returns an OS dependent path i.e. `/` on unix style OS and `\` on Windows.  While Go itself is able to handle this easily, Hugo does not seem to expose any standard functions to either detect the path separator (`PathSeparator` in https://golang.org/pkg/os/#FileInfo) or canonicalise a path in a safe way (ala `expand_path` in Ruby)

This means the hardcoded separator code fails in Windows operating systems

My suggestion would be to replace the code above to be similar to something below:

```
{{ $thisPath := replace .Page.Source.File.Path "\\" "/" }}
{{ $path := split $thisPath "/" }}
```

While I hate the use of string replacement here as it is a hack, it should be an ok shortcut to use.   There should be no file or directory names with `\` as it is an illegal filename in Windows.

If you're ok with this solution I'm happy to raise a PR with the relevant changes.

mattstratton:
My concern is that this may fix the existing issues, but there's no guarantee that new code won't have these issues. 

Since this can run using Docker locally as a "mini-VM", having to put in place a requirement to defensively code around Windows for all shared code dev isn't something we can really do.

mattstratton:
I've got an idea for making this easier on Windows, but it will have to wait until after Chicago this week :)

glennsarti:
@mattstratton Ok.....In the mean time, I'll raise a ticket against Hugo to get an "agnostic" relative path or something to help with this situation.

glennsarti:
To be fair, this doesn't appear to be a fault with this project, but how Hugo is handling paths.  Unfortunately this project may need to carry a workaround until Hugo is updated :-(

mattstratton:
Would this help in the meantime? https://discuss.gohugo.io/t/hugo-on-bashonwindows-windows-subsystem-for-linux/3976

PR #1085, opened 2016-09-09
(GH989) Fix building Hugo website on Windows

glennsarti:
Previously the `File.Path` property would use OS dependant path seperator, back
slash on Windows and forward slash on *nix and Mac.  However as we are parsing
the event slug based on a path splitting this is problematic.

This commit changes the path splitting character to a Site level parameter which
can be changed prior to building via Hugo. The default value should be set to
`/` otherwise all of the building automation will fail, amongst other things.

A later commit will add documentation around this configuration item.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,26 @@
+version: 1.0.{build}
+install:
+#  Install Hugo from chocolatey and display the version number
+- cmd: >-
+    choco install hugo -y
+
+    hugo version
+
+# Change the Path seperator to a Windows version (backslash)    
+- ps: >-
+    $buildRoot = $ENV:APPVEYOR_BUILD_FOLDER
+
+    $content = (Get-Content ($buildRoot + '\config.toml')) -join "`n"
+
+    $content = $content -replace 'PathSeperator = "/"','PathSeperator = "\\"'
+
+    Write-Output "config.toml content changed to:`n----`n$($content)"
+
+    $content | Set-Content ($buildRoot + '\config.toml')
+
+build_script:
+#  Build the devopsdays-web site once
+- cmd: >-
+    hugo --verbose
+
+test: off
--- a/config.toml
+++ b/config.toml
@@ -19,3 +19,7 @@ buildDrafts = false
 	Twitter = "http://www.twitter.com/devopsdays"
 	Linkedin = "http://www.linkedin.com/groups?home=&gid=2445279"
 	Groups = "http://groups.google.com/group/devopsdays"
+
+	# Set to "/" on *nix and Mac, "\\" on Windows
+	# Refer to https://github.com/spf13/hugo/issues/2394 for more information
+	PathSeperator = "/"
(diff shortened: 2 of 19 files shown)


glennsarti:
This is an alternate solution for #989 

- By moving the hard-coded path separator to a configuration parameter it is easy to change for a Windows or *nix platform
- Doesn't require the installation of any additional software (Docker)
- Default value for Linux so no workflow changes for anyone else
- The files affected are the templates and shortcodes which the majority of users will not be modifying.  The significant number of commits to shortcodes and themes are by yourself or Bridge.  If any other user copy/pasta's code it will use the PathSeperator parameter already
- There is still a risk that someone may use a platform specific value of '/' or '\' for that matter, but IMHO that risk appears very low and easy to rectify if committed.  By adding a Windows based PR CI step (e.g. Appveyor) would reduce that risk to practically zero.

bridgetkromhout:
Thanks for working on this, @glennsarti! The best person to work through this with you is @mattstratton  so I'm ensuring nobody else merges it by accident until he's worked on it with you. Thanks.

mattstratton:
I do agree that if we put in an appveyor step, that will help us catch any windows gotchas in the build. I think that having the Docker option as well can be helpful (it's nice to not have to install other things, and it also ensures that everyone is using the version of Hugo that is standard, rather than having to make sure to keep Hugo up to date), but this solution does provide Docker not being in the critical path. 

I'll dig into this a bit more today and vet it out. This looks pretty rad though.

glennsarti:
I agree, this work certainly doesn't exclude a docker build for Windows.

glennsarti:
Pending @mattstratton return mid October.

mattstratton:
@glennsarti I think that before we take on this work, I want to make sure I have a good step in place to check windows compat (via Appveyor as you suggested). However, I am woefully ignorant in appveyor and when I tried it, I broke stuff (well, maybe that was a red herring). 

Would you be willing/able to help get the appveyor thing set up so we can include it in the testing pipeline?

For the overall issue, I'm planning to refactor the way that the directories, etc, work in the redesign later this year, so it will definitely crop up more and we need to solve for it. I need to think about doing it this way vs another tricky method (mostly I'm worried about making sure that a coder, i.e., me, doesn't forget to use this everywhere we split on a path).

glennsarti:
@mattstratton Absolutely no problem to set that up for you.  It was on my list of things to do anyway.  I won't be able to get to it until late October due to work commitments and vacation time.  Is that ok?

mattstratton:
@glennsarti totally fine with me, since I'm pretty wrapped up myself. we can talk in slack about the account to use, etc, so that once it's set up the devopsdays team can own the process.

glennsarti:
Haven't forgotten about this...Will be back on it as soon as I finish my vacation.

glennsarti:
@mattstratton Updated this PR with the required `appveyor.yml` file.  You can see the result of this on my fork and my personal AppVeyor account

https://ci.appveyor.com/project/glennsarti/devopsdays-web

glennsarti:
The `-replace` statement is a bit naive but it should be ok.

glennsarti:
ping @mattstratton

PR #1101, opened 2016-09-13
KC - Adds OTSI as sponsor

barkerd427:
I'm assuming the extra lines are a tabs versus spaces issue, but I don't really have time to argue that point right now. Everything else looks good.


bridgetkromhout:
No worries. Thanks!

PR #1104, opened 2016-09-13
Fixing links

bridgetkromhout:
Dealing with accents. /cc @peterjshan


peterjshan:
@bridgetkromhout Thanks!

PR #1117, opened 2016-09-16
No title yet for Berenice's talk - do not assign a made up one

clintoncwolfe:
We don't have a title yet for Berenice Cotero's talk.  I'm working with here to pin down a topic, but we should advertise it as "Infrastructure Automation".


clintoncwolfe:
Talked to Pete, he said BK said a TBD would read poorly.  Withdrawing.

PR #1120, opened 2016-09-16
Update Maidson

stahnma:
New sponsors: Adorable, Pivotal Victorops


bridgetkromhout:
Removed inactive organizer from email and Slack per this removal from the list. Thanks.

PR #1129, opened 2016-09-18
Add OS detection for correct 'sed' syntax

zl4bv:
This will allow the 'add speakers' script to work as expected on Linux in addition to OS X.


zl4bv:
Closing - will reopen from another branch in our fork.

PR #1132, opened 2016-09-19
Added OS detection for correct 'sed' syntax

zl4bv:
This will allow the 'add speakers' script to work as expected on Linux in addition to OS X.

Replacement PR for #1129.


bridgetkromhout:
Thanks! I'll give this a try tomorrow. :)

PR #1146, opened 2016-09-21
Madison-2016: Add Healthfinch as a gold sponsor

stahnma:
Add healthfinch as a Gold Sponsor


phrawzty:
Hi @stahnma,

I merged this because it's important to move quickly where sponsors are involved; however, if would be so kind as to re-size the sponsor logo to [200px square](https://github.com/devopsdays/devopsdays-web/blob/master/utilities/README.md#sponsor-logos) and then re-submit a PR, I'd really appreciate it.

Thanks!

PR #1148, opened 2016-09-22
[Boise 2016] Add HP as Silver sponsor

tresni:


phrawzty:
Hello!

There are no technical issues with this PR; however, I'd just like to make sure that this is for sure what the sponsor wants.  HP is a family of brands, and we already have a number of *instances* of HP within the sponsor framework - might one of those be appropriate?  Also, @bridgetkromhout mentions that HP had a re-org recently, and that may have an impact on which logo and/or brand that they'd like to express as part of your event's sponsorship deal.

All of this to say that we just want to be sure that this is exactly the logo and/or brand that the sponsor really wants, before merging. 😄

tresni:
Understood.  We actually have both HPE and HP sponsoring [which is confusing enough for me ;-) ]  That is the logo the sponsor sent over and (as of right now) the appropriate URL we've been made aware of.  I will request confirmation.

tresni:
I am being told by my co-organizer that this is correct.

PR #1150, opened 2016-09-22
[2016 new zealand] Added speaker and program data

DefSol:
Added speaker deets and program links. A few content changes as well


phrawzty:
Hello!

The text edits all look fine. Concerning the speaker images, ideally the files should be [exactly 500px wide](https://github.com/devopsdays/devopsdays-web/blob/master/utilities/README.md#speaker-images) (though we try not be sticklers about it).

There is one more pressing issue however: the image for Quentin (` static/events/2016-newzealand/speakers/quentin-adam.jpg `) appears to be broken.

DefSol:
Hey Dan, sorry I missed the image size - we got a lot of variance in these, so I can do but it will take some time. Any chance to push this through and I follow up in a couple of days after I reach out to some of the speakers?

Also the pic for Quentin is just the default (blank) image that get provisioned for in the add_speaker script. On my local (lol) and on the deploy preview this is rendering, i.e. the link does not appear to be broken.

image

Is this the place you were meaning?

r

Issue #636, opened 2016-06-23
Create data-driven program template


mattstratton:
This could be a huge undertaking, but given the amount of crazy html you have to write in a `program.md`, and how easy it is for those pages to drift from the standard bootstrap class usage, it would be rad if it could be data-generated (i.e., a datafile that has the program details in it, and it generates the template based upon it). 

This would easily be over-rideable by not making your `program.md` have the type of `program`, for the one-offs, but it would be nice to avoid having people needing to code html in the .md file.

mattstratton:
Being tracked here - https://github.com/devopsdays/devopsdays-theme/issues/37

PR #1157, opened 2016-09-23
program_entry Shortcode

tresni:
Includes example of use for Boise 2016.


bridgetkromhout:
Thanks, but please put the shortcode change in a different PR, per the guidelines: https://github.com/devopsdays/devopsdays-web/blob/master/CONTRIBUTING.md#guidelines

mattstratton:
Normally I would agree with @bridgetkromhout about splitting these, but since I would like to see what the shortcode actually DOES, we do need the content in the PR as well. 

I've usually done this in the past using my own city's event as the example event.

mattstratton:
We need to keep this change in mind when working on #636 in the future.

tresni:
@mattstratton Appropriate directory is actually `themes/devopsdays-responsive/layouts/shortcodes/` and I've moved the shortcode there.

PR #1183, opened 2016-09-28
Ohio 2016/bios abstrats

kjenkins19:
Ohio 2016 Bios, Abstract, and currently blank images. Panera is about to close and it's going to take a while to download and get the real ones.

Also, if it is possible to link to existing ones (we have a fair amount of speakers with existing images elsewhere), please let me know how (or point me in the right direction for documentation). For example, I want to use an image already in 2016-newyork. Thanks!


bridgetkromhout:
> Also, if it is possible to link to existing ones (we have a fair amount of speakers with existing images elsewhere), please let me know how (or point me in the right direction for documentation). For example, I want to use an image already in 2016-newyork. Thanks!

We deliberately keep a separate image (and speaker data file) in each event because speakers sometimes update their headshot and their bios, and maintaining a separate versioning system for that would be Not Fun. (Already doing so for a different site - arresteddevops.)

Copying in the image files from other recent cities where it's probably still their valid picture is pretty trivial. I'll take a pass at it for you.

PR #1191, opened 2016-09-28
master

gilzellner:
logz.io changed their logo


mattstratton:
We need to check to make sure no other events have used this file; if so, you should create a new sponsor file. We don't rewrite history.

bridgetkromhout:
I've corrected this in https://github.com/devopsdays/devopsdays-web/pull/1197.

Of note: the old "logz" file had a custom URL that was only for 2015-boston, and the old "2016-logz" file will likely be fine into 2017. That's why we now only want to use "logz" and it's thus corrected. Thanks.

PR #1194, opened 2016-09-28
[Raleigh-2016] Update Sponsor page

mrry550:
Marking gold & platinum as sold out.


bridgetkromhout:
You can also remove the "become a _whatever_ sponsor" link from the sidebar. Edit https://github.com/devopsdays/devopsdays-web/blob/master/data/events/2016-raleigh.yml#L86-L89 and to the ones that are sold out, add a max as shown on https://github.com/devopsdays/devopsdays-web/blob/master/utilities/examples/yyyy-city.yml#L56-L59. If the max is where they are right now, the link won't show.

bridgetkromhout:
I'll merge this for now and you can submit another PR if you'd like to remove those links.

PR #1241, opened 2016-10-05
SIN-16 sponsors: Add a group image of the logos

ctrabold:
We send the logos in the mailings which only allow inline images.


mattstratton:
In the future, it would have been better to put this static file in 
`/static/events/2016-singapore`

The resultant URL would have been identical, but it would prevent hugo from trying to parse the static file.

ctrabold:
@mattstratton Oops - shall I move it?

PR #1253, opened 2016-10-06
create_sponsors.sh: Support logo paths with whitespace

ctrabold:
This is useful when storing files on Google Drive for example:

    ~/Google\ Drive/DOD/sponsors/example.png

The whitespace issue hit me when I was dragging a file from Google Drive onto the terminal to set the logo path – the script just crashed  :collision:


ctrabold:
I think it's safe because for the destination the `$sponsor_slug` is used (which must not have white space: `$sponsor_slug.png`

The `$logo` variable is only used as file _source_.

PR #1279, opened 2016-10-11
[2017-paris] init stub

phrawzty:
This is really just a stub for Paris 2017 (as per @bridgetkromhout's request on Slack 😄). Only the welcome and contact pages are linked. The standard `organizers` address has been created (it only has me and @jooooooon on it for now).


jooooooon:
Huh. That didn't come out as I intended. Plus one. R+. I approve. Looks empty but good. You get the gist :-)

PR #1298, opened 2016-10-14
[philadelphia-2016] Swap Linhares for Windisch

clintoncwolfe:
Eric Windisch had to cancel, so we're activating one of our alternate speakers.


phrawzty:
Sorry to hear that a speaker had to drop out - but it'll be a great opportunity for your alternate! 😄

PR #1321, opened 2016-10-18
[2016-Ghent] Added Ordina as sponsor

tolomaus:


data/events/2016-ghent.yml
```diff
@@ -53,10 +53,12 @@ sponsors:
     level: gold
   - id: coscale
     level: gold
+  - id: ordina
+    level: gold
   - id:  combell
     level: silver
   - id:  openminds
-    level: silver
+    level: community
```
phrawzty: The commit message only references Ordina, but Openminds has had their level changed here.  Is this on purpose?  Just verifying.

tolomaus: I was a bit short in the commit message but indeed Openminds should have its level changed so all is good.

Thanks for verifying though!

PR #1356, opened 2016-10-25
Update Philly sponsor logos

davekonopka:
Updates Aweber and Cloudbees logos. Adds 2016 version of Cloudbees sponsor record, image to avoid changing other events usage unintentionally.


bridgetkromhout:
I'll merge this since your event is so close, and then I'll do the `sed`-ing necessary to properly preserve past state.

davekonopka:
Thanks @bridgetkromhout! Is there something I did incorrectly? Happy to clean up if you can point me in the right direction.

PR #1360, opened 2016-10-25
[2016-BNA] Additional ignite program adjustments

jfryman:
This PR updates our ignite schedule again. Have to remove someone due to personal issues, so adding another to the schedule.


phrawzty:
Sorry to hear about the issue, but on the plus side, the new ignite sounds fun! 😄

PR #1362, opened 2016-10-26
[zurich-2017] new sponsor and organizer

locolupo:
We've added a new sponsor and a new organizer to the event.


phrawzty:
Hello,

Please supply the email address of the new organiser as well (you can do so in a less public forum, if you so choose), so that we can add them to the mailing list and Slack.  Thanks!

phrawzty:
Communicated via Slack.

PR #1367, opened 2016-10-27
[2016-Nashville] Fix stat on Ignite talk

stephenyeargin:
I had this way, way wrong. A better stat:

>Colour (color) blindness (colour vision deficiency, or CVD) affects approximately 1 in 12 men (8%) and 1 in 200 women in the world.

Source: http://www.colourblindawareness.org/colour-blindness/

/cc @jfryman


phrawzty:
As it happens, *I'm* included in that group - thanks for raising awareness! 💯

PR #1368, opened 2016-10-27
[tlv 2016] sponsors

gilzellner:


phrawzty:
As a side note, thanks for getting a proper sponsor kit in for AWS. 😄

PR #1370, opened 2016-10-27
fix logo cloudinary + fix url aws

gilzellner:


bridgetkromhout:
Thanks! (NB: the logo is only used for this event this year; the URL definitely needed fixing.)

PR #1424, opened 2016-11-09
sponsor update url

gilzellner:


bridgetkromhout:
Fine to update without backwards-compat as this is the only event with this sponsor.

PR #1447, opened 2016-11-16
[2016-tlv] sponsors + update gitlab to new logo

gilzellner:


MikeRosTX:
Please review the github file name extension. I think it's missing the last letter of the extension because it's not displaying anything,

gilzellner:
I'm not sure about which file name issue you are referring to.

as for gitlab logo: gitlab is also used by: cuba, brazilia, Nashville
However, this is the newer logo and makes better use of space, so It most likely will be better for everyone since others will ask to replace it as well later.
I don't mind moving it to gitlabil or something of the sort but it seems kind of silly to me.

phrawzty:
Regarding the Gitlab logo, the previous one violates our dimension requirement, so this constitutes a fix in any case. Plus, as @gilzellner said, this one looks *way* better on our site. 👍

mattstratton:
1 - the filename issue that @MikeRosTX mentioned isn't an issue - for some reason he didn't see it in the preview link but I did. 

2 - even if the vendor changes their logo, you shouldn't change it if it affects past events. That was the logo at the time of that event and as @bridgetkromhout says, we don't rewrite history. The right thing to do would have been to create `gitlab-2016` or something. 

This is sticky because it was also a fix for the size of the logo as mentioned. 

I'm not saying to revert this change, but more for clarification for everyone involved for the future.

PR #1448, opened 2016-11-16
fix waffle image link

mattstratton:


mattstratton:
@glennsarti not sure if you can help with this - look at the checks (and then the details) it's building master as well as the branch when a PR comes in; do you know how to fix that in appveyor?

Issue #933, opened 2016-08-22
correct date hardcoding for future events


bridgetkromhout:
As noted in https://github.com/devopsdays/devopsdays-web/pull/910, I introduced date hardcoding that should be corrected with a variable.

bridgetkromhout:
This is resolved in https://github.com/devopsdays/devopsdays-web/pull/1455.

PR #1455, opened 2016-11-20
Deprecate the status: parameters (current & past)

bridgetkromhout:
The reason this was worth doing is that manually setting an event to "past" in a timely fashion is annoying to have to remember.

I think I've corrected all the places that were using this parameter. I've also corrected the data file for the template (and for upcoming events - while not strictly speaking necessary, might as well work cleaner). 

There are a few subtle tradeoffs - I went into some detail in the commit messages.

I'd appreciate @mattstratton taking a look since this is a change with some scope to it.

### DIFF
--- a/themes/devopsdays-responsive/layouts/partials/upcoming_headline.html
+++ b/themes/devopsdays-responsive/layouts/partials/upcoming_headline.html
@@ -1,12 +1,15 @@
 <div class="upcoming-headline">
   <h3 class = "upcoming-headline">
   {{ range sort $.Site.Data.events "startdate" }}
-    {{ if  and ( eq .status "current") ( .startdate ) }}
+    {{ if .startdate }}
+    {{ if ge (dateFormat "2006-01-02" .startdate) (dateFormat "2006-01-02" ($.Now.Format "2006-01-02")) }}
       {{ $.Scratch.Add "events" "<a href = '/events/" }}
       {{ $.Scratch.Add "events" .name }}
       {{ $.Scratch.Add "events" "/welcome' class = 'upcoming-headline'>" }}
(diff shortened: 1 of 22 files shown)


themes/devopsdays-responsive/layouts/partials/upcoming_headline.html
```diff
@@ -1,12 +1,15 @@
 <div class="upcoming-headline">
   <h3 class = "upcoming-headline">
   {{ range sort $.Site.Data.events "startdate" }}
-    {{ if  and ( eq .status "current") ( .startdate ) }}
+    {{ if .startdate }}
+    {{ if ge (dateFormat "2006-01-02" .startdate) (dateFormat "2006-01-02" ($.Now.Format "2006-01-02")) }}
```
bridgetkromhout: And yes, I know I'm using two `if` lines in all these cases instead of combining them with an `and` - it's because when I try to combine them I get this:
_ERROR: 2016/11/19 21:15:28 template.go:131: template: theme/partials/upcoming_headline.html:4:33: executing "theme/partials/upcoming_headline.html" at <dateFormat "2006-01-...>: error calling dateFormat: Unable to Cast <nil> to Time
 in theme/partials/upcoming_headline.html_
One needs to get evaluated first.

bridgetkromhout:
This will resolve https://github.com/devopsdays/devopsdays-web/issues/933 and https://github.com/devopsdays/devopsdays-web/issues/882 when merged.

mattstratton:
Coincidentally enough, I was just looking at a lot of this code due to the redesign, so this just saved me a lot of troubleshooting.

PR #1491, opened 2016-12-08
Add 3 sponsors

VitMosin:


MikeRosTX:
@VitMosin I'll approve it for now just to get the sponsor's logos on the site, but I would recommend the following:

1. To submit a new PR to move Media Sponsors in between Silver and Community Sponsors.
2. Also, ADO would be another Media Sponsor because they broadcast the event over podcasts.

PR #1509, opened 2016-12-12
Add 2017 Beijing

xiaotianguo:
Hi,I'm the 2017 Beijing DevOpsDays organizer,This is 2017 Beijing pull request


MikeRosTX:
@xiaotianguo before we can merge we need to know that you have a confirmed contract on the location. If you do, please add location to navbar. Also on the contacts, add all your local organizers.

PR #1524, opened 2016-12-17
Atlanta 2017 initial commit

cacorriere:
we have our venue reserved so our dates are included.


MikeRosTX:
Hey @cacorriere,  glad to see Atlanta! However, do you have a signed contract with a venue? If not, please remove the dates.

PR #1525, opened 2016-12-18
Updating sample proposal language.

bridgetkromhout:
I wrote new language for 2016 and a number of events (aside from Minneapolis) adapted it. I'm now ready to consider this the new standard. Feedback welcome. I'll merge this in a day or two after folks who want to weigh in have done so.

- I'm clarifying that people don't propose Open Space during the CFP
- I'm clarifying the inclusivity and desire for both local voices and non-pitchy proposals (and recommending sponsorship instead, for product demos).
- I'm removing language that commits an event to something they might not have, like single-track.
- While adding guidelines, I'm removing the label of "rules".

### DIFF
--- a/utilities/examples/sample-event/propose.md
+++ b/utilities/examples/sample-event/propose.md
@@ -7,32 +7,30 @@ draft = true
   {{< cfp_dates >}}
 
 <hr>
-There are three ways to propose a session:
+
+There are three ways to propose a topic at devopsdays:
 <ol>
-  <li><strong><em>A proposal for a talk/panel</em></strong> during the conference part : these are 30 minute slots that will have the full attention of all attendees, as everybody will be in that one room.</li>
-  <li><strong><em>An Ignite talk</em></strong> that will be presented during the<a href="/pages/ignite-talks-format"> Ignite sessions</a>. These are 5 minutes slots with slides changing every 15 seconds (20 slides total) which are also presented to all attendees in one room</li>
-  <li><strong><em>Open Space session</em></strong> : even without a prepared presentation we welcome the discussion and interaction by having people propose a session on the fly during Open Space. Check the <a href="/pages/open-space-format">Open Space explanation</a> for more information.
+  <li><strong><em>A 30-minute talk</em></strong> presented during the conference, usually in the mornings.</li>
+  <li><strong><em>An Ignite talk</em></strong> presented during the <a href="/pages/ignite-talks-format">Ignite sessions</a> (scheduling varies). These are 5 minutes slots with slides changing every 15 seconds (20 slides total).</li>
+  <li><strong><em>Open Space</em></strong>: If you'd like to lead a group discussion during the attendee-suggested <a href="/pages/open-space-format">Open Space</a> breakout sessions, it is not necessary to propose it ahead of time. Those topics are suggested in person at the conference. If you'd like to demo your product or service, you should <a href="../sponsor">sponsor the event</a> and demo it at your table.
 </ol>
 
-### Even if you don't propose, please consider {{< event_link page="proposals" text="commenting on proposals submitted by others" >}}
+<hr>
 
-Our main criteria to make it to the top selection are:
+Choosing talks is part art, part science; here are some factors we consider when trying to assemble the best possible program for our local audience:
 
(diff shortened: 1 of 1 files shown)


utilities/examples/sample-event/propose.md
```diff
@@ -7,32 +7,30 @@ draft = true
...
 </ol>
 
-### Even if you don't propose, please consider {{< event_link page="proposals" text="commenting on proposals submitted by others" >}}
+<hr>
+
+Choosing talks is part art, part science; here are some factors we consider when trying to assemble the best possible program for our local audience:
 
-Our main criteria to make it to the top selection are:
+- _broad appeal_: How will your talk play out in a room of people with a variety of backgrounds? Technical deep dives need more levels to provide value for the whole room, some of whom might not use your specific tool.
+- _new local presenters_: You are the only one who can tell your story. We are very interested in the challenges and successes being experienced in our local area. We are happy to provide guidance/coaching for new speakers upon request.
```
phrawzty: Should this be "in **y**our local area"?

utilities/examples/sample-event/propose.md
```diff
@@ -7,32 +7,30 @@ draft = true
...
+- _under-represented voices_: We want to hear all voices, including those that may speak less frequently at similar events. Whether you're in a field not typically thought of as a technology field, you're in a large, traditional organization, or you're the only person at your organization with your background, we are interested in your unique experience.
+- _original content_: We prefer talks not already presented at another event, especially another one in the local area or one available in recorded form.
```
phrawzty: Does this need the additional clarification?  For example, why can't we just say: 
"We prefer talks that haven't already been presented elsewhere."
It encompasses the local and/or recorded form cases, no?

If anything, I wonder if additional clarification should go in the other direction, i.e. that previously-presented talks will indeed still be considered.

utilities/examples/sample-event/propose.md
bridgetkromhout: I think that the organizers in City X want to hear about City X's local area, not Random Town Proposer Is From's local area.

utilities/examples/sample-event/propose.md
```diff
@@ -7,32 +7,30 @@ draft = true
...
+- _no third-party submissions_: If a PR firm or your marketing department is proposing the talk, you've already shown that as a speaker you're distant from the process. This is a small community-driven event, and speakers need to be directly engaged with the organizers and attendees.
```
phrawzty: I agree 100% with the spirit of this bullet point, but I worry that it might exclude people that work for organisations wherein talk proposals *must* pass through $some_department (Facebook comes to mind).

Maybe just the second sentence is sufficient as amplification for the bullet point?

utilities/examples/sample-event/propose.md
phrawzty: Ah, OK, I mis-interpreted the implied locale of the narrator here.

utilities/examples/sample-event/propose.md
phrawzty: Yes, I like that much better - it leaves a comfortable opening for interpretation. :smile:

utilities/examples/sample-event/propose.md
bridgetkromhout: Minneapolis gets proposals being shopped around by third-part PR firms every year (who are looking for a commission or cut of the non-existent speaker fee). We also get the marketing department of some startup who wants to put their CTO on stage to give a thinly disguised vendor pitch. I definitely want to discourage such proposals.

utilities/examples/sample-event/propose.md
mattstratton: I also think that this is something that will vary from event to event; remember, this is the template that will be generated; it's not mandated that each event use all of these words :)

utilities/examples/sample-event/propose.md
phrawzty: Ok, how about reversing the order of the two statements? It feels less antagonistic, imho.

(This could just be my :canada: showing, hah.)

phrawzty:
In general I'm all :+1: on these edits. Critically, I feel that the *intent* of these guidelines is much more clear, which is great. Thanks for proposing, @bridgetkromhout!

PR #1530, opened 2016-12-21
dates for Atlanta - contract signed

cacorriere:
we have a contract with the Academy of Medicine (same venue as last year) for April 18th-19th signed


MikeRosTX:
Hi @cacorriere! Add a Location on the navbar with the address on the Location before we merge it.

PR #1533, opened 2016-12-21
Fixing London data and vim ignores in .gitignore

kramos:
Fixing London data and vim ignores in .gitignore


deanwilson:
Looks good. Merge away.

PR #1545, opened 2016-12-23
Seattle 2017 sponsors

kmugrage:
Adding Deis as a Gold. 

They asked me to replace the logo because they felt this one is clearer, but I don't see a material difference. Because of that, I did not create a 'before' version. Would like feedback from @mattstratton or @bridgetkromhout before merging.


bridgetkromhout:
Yeah, I find the "after" version indistinguishable from the "before" version. Fine by me.

PR #1583, opened 2017-01-10
Added details for DevOpsDays Denver

tavisto:


MikeRosTX:
@tavisto need to know the following before merging:

1. Does Denver have a confirmed contract signed with Exdo Event Center, Denver?  If so, then enable Location on the navbar with the address location.
2. Remove Site from the navbar, because it is pointing to the 2016 site and it's confusing since the venue from 2016 is different from 2017.  Once 2017 site has been updated, you can reactive it on the navbar.

tavisto:
Thanks. We just signed the contract with Exdo so I will update the
navigation etc.
I will also remove the site link until I update that as well.

PR #1618, opened 2017-01-18
Added ticket link to 2017 denver event

tavisto:


bridgetkromhout:
Hey, @tavisto - this isn't actually how to use event_link since you need to send visitors offsite, and don't have a page with the name `/tickets`. You'll want to use a normal href link on your welcome page.

tavisto:
Ahh right thanks @bridgetkromhout I will fix and submit an update shortly.

PR #1620, opened 2017-01-18
Removed old webby references

mattstratton:
Many of the sites still had commented out code that referenced webby partials. This is causing issues on the new design.

This code:

```
<a href="mailto:<%= render(:partial => "/#{@eventhome}/_email_organizers") %>?subject=Sponsor devopsdays <%= @eventid %>">send us an email</a>
```
has been replaced with `send us an email`. This is only on older events, so it should not be an issue.


vjanelle:
Not sure why the content for the vancouver page changed?

mattstratton:
The current Vancouver page?

mattstratton:
Oh.  Because you had the old webby code that doesn't even work. Look at the commit. Y'all must have copied that from a pre-Hugo page. It wouldn't have even rendered properly.

mattstratton:
The change is also in commented-out code anyway :)

PR #1632, opened 2017-01-21
[Scripts] Changing the default to 2017

cobusbernard:
Using `utilities/make_new_event`, when not entering a value for the year, it displays that it will use `2017`, but creates the event as `2016`.  

```
$ ./make_new_event.sh
Enter your event year (default: 2017):
Enter your city name: Cape Town
Enter your devopsdays event twitter handle (defaults to devopsdays): devopsdayscpt

$ git status
  content/events/2016-cape-town/
  data/events/2016-cape-town.yml
```

Updated the bash scrip to fix this.


bridgetkromhout:
Thanks for catching my error, Cobus! =)

PR #1657, opened 2017-01-27
[Baltimore-2017] More sponsors for Baltimore

nathenharvey:
* Circonus (moved to correct location)
* ServiceNow
* VictorOps


mattstratton:
What does adding the `twitter` field to a sponsor do?

nathenharvey:
@mattstratton technically...nothing.

see also https://github.com/devopsdays/devopsdays-web/pull/706#issuecomment-231562025

>I was also thinking that the Twitter handle would be helpful for other organizers. As we start getting new sponsors, we want to thank them on Twitter. If they've sponsored in the past, finding the Twitter handle is as easy as reusing their logo. There may also be some future programmatic use for that data.

> I happy to remove it but was thinking it might be of use to other organizers.

bridgetkromhout:
Hey @nathenharvey the future-looking Twitter field is great but can you remove the newline? Thanks.

nathenharvey:
@bridgetkromhout not sure how that snuck in there but should be gone now, thanks!

bridgetkromhout:
Thanks @nathenharvey - still a bit weird? Your editor is being tricksy. :) I mean, we want an invisible newline, just not a control-M character.

screen shot 2017-01-27 at 1 33 56 pm

nathenharvey:
yeah, IDK why my editor decide to switch from `LF` to `CRLF` :(

fixed the right way now.

PR #1659, opened 2017-01-27
[CPT-2017] Adjust event pages styling to suit mobile screens

manicminer:
Adjust styling for Cape Town 2017 pages to fit mobile screens


mattstratton:
This should be fine, although you have Bootstrap at your disposal, so you can make things more responsive by using the `<div>` classes from Bootstrap. Since we're going to be refactoring the theme shortly anyway, I'm not going to sweat it too much :)

Issue #1664, opened 2017-01-30
Appveyor job throwing error on hugo command


mattstratton:
```
Build started
git clone -q --branch=dodAMS-2017-content-chgs https://github.com/devopsdays/devopsdays-web.git C:\projects\devopsdays-web
git checkout -qf 2081c53942d173e6198a0c8798697b190c31dd08
Running Install scripts
choco install hugo -y
Chocolatey v0.10.3
Installing the following packages:
hugo
By installing you accept licenses for the packages.
hugo v0.18.1 [Approved]
hugo package files install completed. Performing other installation steps.
Downloading hugo 64 bit
  from 'https://github.com/spf13/hugo/releases/download/v0.18.1/hugo_0.18.1_Windows-64bit.zip'
Progress: 100% - Completed download of C:\Users\appveyor\AppData\Local\Temp\1\chocolatey\hugo\0.18.1\hugo_0.18.1_Windows-64bit.zip (5.11 MB).
Download of hugo_0.18.1_Windows-64bit.zip (5.11 MB) completed.
Hashes match.
Extracting C:\Users\appveyor\AppData\Local\Temp\1\chocolatey\hugo\0.18.1\hugo_0.18.1_Windows-64bit.zip to C:\ProgramData\chocolatey\lib\hugo\tools...
C:\ProgramData\chocolatey\lib\hugo\tools
 ShimGen has successfully created a shim for hugo_0.18.1_windows_amd64.exe
 The install of hugo was successful.
  Software installed to 'C:\ProgramData\chocolatey\lib\hugo\tools'
Chocolatey installed 1/1 packages. 0 packages failed.
 See the log for details (C:\ProgramData\chocolatey\logs\chocolatey.log).
hugo version
'hugo' is not recognized as an internal or external command,
operable program or batch file.
Command exited with code 1
```

mattstratton:
Until the chocolatey package is updated to rename the hugo executable, we will need to refer to it explicitly, which isn't good for versions.

mattstratton:
Fixed in 3a819d0bcf5172ab11c75b9008664bd0d6e0edce

I'm leaving this open until the [chocolatey package](https://chocolatey.org/packages/hugo) gets updated.

chantisnake:
I am the maintainer of Hugo. I saw your message. I personally think the original renaming command is not secure (there will be error if there is more than one exe in the root.

I will find a way soon.

I also recommend you guys to find a way to adopt the long name.

[...]

PR #1679, opened 2017-02-01
Add twitter handles to upcoming events & template.

bridgetkromhout:
Adding this element to the event yaml (and template for the event yaml) in preparation for using it.


bridgetkromhout:
Merging despite appveyor issue - see https://github.com/devopsdays/devopsdays-web/issues/1664

PR #1681, opened 2017-02-02
[VAN-2017] adding Kyle to organizers list

rmkbow:


phrawzty:
Hello,

If Kyle needs to be added to the mailing list and/or Slack, please forward their email address to us, and we'll set that up.

PR #1684, opened 2017-02-02
[ZRH-2017] Added workaround URL in case of payment page problems

Roderick-Jonsson:
We added a workaround url into the registration page as it some time seems that the devopsdays.org rocketscript js intervenes with our javascript snippet.

### DIFF
--- a/content/events/2017-zurich/registration.md
+++ b/content/events/2017-zurich/registration.md
@@ -5,7 +5,8 @@ type = "event"
 
 
 +++
-Registration for DevOpsDays Zürich is now open. If you have any questions please contact us at {{< email_organizers >}}.
+Registration for DevOpsDays Zürich is now open.<br> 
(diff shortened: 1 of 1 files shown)


bridgetkromhout:
This is worth @mattstratton looking at (due to the JS issues you're encountering).

Roderick-Jonsson:
Can someone merge this request as it seems that now the bug is setting it

bridgetkromhout:
Looks like Netlify's deploy previews are still broken; I'll merge and then check to see if it worked.

Roderick-Jonsson:
Thank @bridgetkromhout 🥇

bridgetkromhout:
Looks like the workaround link is working. We probably still want to figure out what the JS conflict is.

mattstratton:
I could be totally wrong, but you can't really embed JavaScript inside the markdown files as Hugo removes them for safety. I'll triple confirm when I'm off mobile.

mattstratton:
OK, I was wrong. You can embed html <script> tags in the post for sure.

mattstratton:
I disabled rocketscript as it's "beta" and if it's interfering with our stuff, then it's not valuable.

PR #1710, opened 2017-02-05
[IST-2017] Welcome and location page updated

funkydorian:
Welcome and location page updated

### DIFF
--- a/content/events/2017-istanbul/location.md
+++ b/content/events/2017-istanbul/location.md
@@ -5,6 +5,29 @@ type = "event"
 
 +++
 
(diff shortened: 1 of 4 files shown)


bridgetkromhout:
Hi -

If your call for participation is going to be open now, then please also link it in the navigation bar and announce its dates. (You uncommented it, but maybe that was an error?) 

Also, you might want to uncomment the "location" link in your navigation bar.

funkydorian:
We haven't decide the channel of registration (meetup, eventbrite,..) that's why it is still uncommented

PR #1730, opened 2017-02-08
[OSL-2017] Added dates and location

steinim:


steinim:
We cannot merge this just yet. The venue just contacted me that there may be some vacancy problems with the dates.

PR #1735, opened 2017-02-09
[OSL-2017] The dates have been confirmed and CFP is open

steinim:


MikeRosTX:
@steinim place the date at the top on the **Welcome** page underneath the navbar where you have **_For the second time, DevOpsDays is coming to Oslo!_** then we'll merge it.

MikeRosTX:
@steinim mandate a layout of a certain style, I'm just making a simple suggestion so that your audience can quickly see the date big at the top and mark their calendars.  We'll go ahead and merge it and if you would like to change it later, just submit a new PR.

As an example of the date suggestion I was making, see [Baltimore's page](https://www.devopsdays.org/events/2017-baltimore/welcome/)

PR #1754, opened 2017-02-14
[LA-2017] Add Thoughtworks, update CFP dates

irabinovitch:
* Add ThoughtWorks as a sponsor
* update CFP dates

### DIFF
--- a/data/events/2017-los-angeles.yml
+++ b/data/events/2017-los-angeles.yml
@@ -48,6 +48,8 @@ sponsors:
     level: sponsor
   - id: 2016-carsdotcom
     level: sponsor
+  - id: thoughtworks-products
+    level: sponsor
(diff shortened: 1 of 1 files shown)


MikeRosTX:
@irabinovitch what is confusing is that the **_welcome_** page says that CFP is open, but at the top of the **_propose_** page it says SFP is TBD.  Wouldn't it be better to fix the propose page to have like dates of when CFP is Open and Closes?

irabinovitch:
@MikeRosTX That page seems to be populated by the data files updated in this commit?

MikeRosTX:
@irabinovitch I merged the PR to get the other changes updated.  But look at the top of the **_propose_** page, it says **_Call for proposals is TBD_**.

PR #1765, opened 2017-02-15
[LA-2017] Add OpsGenie as sponsor

irabinovitch:
- add OpsGenie as sponsor
- remove duplicate thoughtworks logo


phrawzty:
Afaik the Travis failure is not a blocker for this merge (cc @mattstratton).

PR #1790, opened 2017-02-20
Typo fix

bridgetkromhout:
Reading https://github.com/spf13/hugo/issues/2394 I believe that we need this typo fix.


bridgetkromhout:
There's something I don't grasp here, but clearly this shouldn't be merged; closing.

mattstratton:
It needs to be in all lower case with the incorrect spelling as that is called in a LOT of partials.

glennsarti:
Apologies for the typo :-(

PR #1797, opened 2017-02-21
[ZRH-2017] Add confirmed speakers as talk page	

Roderick-Jonsson:
This is a high priority commit as tomorrow a blog post will be made to an audience of around 30'000 people. 

The talks menu entry is disabled and will be enabled with the next high priority commit.


bridgetkromhout:
Happy to merge it if you like, although you might want to use https://github.com/devopsdays/devopsdays-web/blob/master/utilities/add_speakers.sh instead of doing it this way; then you get a page like https://www.devopsdays.org/events/2016-minneapolis/speakers/ for free and can still also build a page like https://www.devopsdays.org/events/2016-minneapolis/program/ too.

mattstratton:
I agree with @bridgetkromhout, especially since there is going to be a more robust program template in the new theme being released soon.

However, if this is important to you to make the page look this way, it's fine. However, I don't know what's going to happen with the new theme with the layout the way you set it up.

Roderick-Jonsson:
@bridgetkromhout @mattstratton As we have all information about the talks and incomplete speaker information it would be great to be able to prominently show off the talks we have gathered. With the proposed speakers template it would look a bit sparse and maybe people wouldn't even notice the talks as we don't have a schedule yet. So yes I would be happy if we could make an exception in this case :)

bridgetkromhout:
@Roderick-Jonsson Understood! Merged, and it's very much "we recommend" not "we decide" - you always decide. :)

Roderick-Jonsson:
@bridgetkromhout Thank you Bridget!

PR #1799, opened 2017-02-21
[ZRH-2017] Enable talks menu entry and added basic speaker list

Roderick-Jonsson:
Last High priority commit to enable menu entry. Please merge as fast as possible. 
I also updated the speakers list with basic names so we can still promote Charity Majors and Hannah Foxwell, as we don't have a talk from them to list in the talks section. @bridgetkromhout @mattstratton


Roderick-Jonsson:
@bridgetkromhout Thank you for the super fast merge!

PR #1817, opened 2017-02-23
Prepare configuration files and other items for new theme

### LINKED ISSUE #1780: Add new config.toml parameters for new theme
So far, the only one I know of is the 'weburl'.

It should be set to:

```
weburl = "https://www.devopsdays.org"
```
This will NOT affect local builds; this is used only for creating outside metadata.

Also, this needs to be changed to be all lowercase, as such:

`  pathseperator = "/"`

### LINKED ISSUE #1816: Need to move partials and shortcodes out of layouts
There are partials and shortcodes in the `layouts` directory of this project that should be moved into `themes/devopsdays-responsive` or we will have a conflict when we try to use the new theme.

mattstratton:
This change moves the shortcodes and partials previously located in the root `layouts` folder into the `devopsdays-responsive` theme, to prevent a collision when we release `devopsdays-theme`.

This also prepares the `config.toml` for use with the new theme.

This has been been regression tested with the existing site and should introduce issues.

Fixes #1816
Fixes #1780


bridgetkromhout:
👍

PR #1866, opened 2017-03-01
[ZRH-2017] add atlassian as evening event sponsor and display on top of list

Roderick-Jonsson:
Atlassian will sponsor the whole evening event and as they are the biggest contributor we will add them to the top of our sponsoring list


Roderick-Jonsson:
@MikeRosTX Thank you for the fast merging 👍

PR #1867, opened 2017-03-01
Remove config.toml from .gitignore

mattstratton:
When the `config.toml`is in `.gitignore`, it means that folks who do a `git pull` won't get any updates. This is going to be a big problem when we release the new theme.

However, my concern is with people making changes to this file. We need to watch for changes on this file before merging PR's.

I need a couple 👍  from maintainers before this gets merged.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,7 @@
 /public
 # General CodeKit files to ignore
 config.codekit
-config.toml
+# config.toml # This is a dangerous file, but we need it not in here.
 /min
 .DS_Store
 .AppleDouble
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -19,6 +19,11 @@ To watch for changes and rebuild on the fly, open a new terminal, change directo
 hugo server -w --baseUrl="http://localhost:1313"
 ```
 
+If you are running Windows, change the command to 
+```
+hugo server -w --baseUrl="http://localhost:1313" --config config-windows.toml
+```
+
 Now open `baseURL` in a browser and navigate to the content that you're editing - voilà!
 
 ## Pull requests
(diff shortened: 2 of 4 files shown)


bridgetkromhout:
👍  We just have to watch if we can't add a post-commit hook or something to warn us.

mattstratton:
@bridgetkromhout I want a hook like that for other files too, so I think we're okay right now. 

I kinda want an easy way to get a message to everyone who has commit bits on this repo; github notifications aren't reliable because people turn them off. Anyway, that's a problem for another day. I'll merge this once I get one more 👍

MikeRosTX:
@mattstratton my concern for adding **config.toml** in to **.gitignore** was because I was doing all my changes in Windows and every time I changed the **PathSeparator** to **\\** it wanted to merge into the main branch. I'm a little unclear how we should making these changes to work in all environments without affecting the main branch.

mattstratton:
Based upon the conversation I just had with @MikeRosTX, what we need to do is create a `config-windows.toml` which looks just like the `config.toml` but has the `PathSeparator` set for the Windows version.

Then when someone wants to run on windows, they run `hugo -w -b "https://localhost:1313 --config windows-config.toml`.

The only thing this means is we have to keep both config files in sync, but they change rarely.

I'll keep this PR open until I can create and test that Windows config file.

mattstratton:
OK I think it worked; insomuch as I tried pointing to that config file using that flag on my Mac and it blew up. 

i'll make one more documentation change before merging.

mattstratton:
I think I caught it all.

mattstratton:
I just tested this on Windows and it works, so I am gonna merge this.

EVERYONE WATCH OUT FOR PR's WITH THESE FILES.

PR #1882, opened 2017-03-02
[SEA-2017] added HPE as Gold

kmugrage:
Added this one in a different way because of the situation. HPE wanted a different URL for and Twitter handle for this specific event. Instead of moving hpe.yml to a -before format I kept it the same for other events, created a new specific yml and duplicated the logo.

@mattstratton - Please let me know if this should be handled in a different way.


mattstratton:
I think that creating a new one the way you did is totally rad.

Bear in mind that the twitter field is not used by ANYTHING on the site, other than a human being looking at the code and seeing the twitter. We have been considering how we might leverage this as introduced by @nathenharvey in this issue - https://github.com/devopsdays/devopsdays-theme/issues/239.

That being said, it doesn't HURT anything to put a twitter handle in there; just making sure you know that it won't do anything other than give you, the organizer, a place to look to remember what twitter handle that sponsor has :)

PR #1888, opened 2017-03-02
Test the mention bot

mattstratton:
This hsould only mention matt


mattstratton:
This should have assigned me as a reviewer, but that's probably because it's my own. But this was a good start.

PR #1932, opened 2017-03-06
Use Markdown instead of HTML

matiasfrndz:
Use Markdown instead of HTML in program


matiasfrndz:
Sorry for the inconvenience @mattstratton, I had prematurely deleted my branch.

PR #1937, opened 2017-03-07
Updating theme to 0.9.2

mattstratton:
Note that this only adds the theme into the repo; it does NOT enable it. Please do not start using this theme. This is a non-destructive change.


devopsdays-service:
@mattstratton, thanks! Based upon the content in this PR, we need a review from @mattstratton and @bridgetkromhout

PR #1950, opened 2017-03-08
[Baltimore-2017] - Fixes to home pages

nathenharvey:
* Adds Baltimore back to the home page of the site (we're still here ;))
* Updates the home page of the Baltimore event.


mattstratton:
Whoops, buggy. Although this is going to look weird in the past events thing later, so let's fix it after the event is over?

The good news is that I just tested in the new theme, and you would still be showing up, so we somehow fixed this bug premptively :)

nathenharvey:
Thanks @mattstratton.  I opened #1951 so that I might remember to do so.

PR #1957, opened 2017-03-08
[DET-2017] Add Zach Zeid to organizers

bchess:
as it says. We'll also want him added to the Google Group


MikeRosTX:
@bchess - don't forget to send an email to info@devopsdays.org including Zach Zeid's email so he can be added to Slack and Detroit's team email.

PR #1969, opened 2017-03-10
[ZRH-2017] Remove slot from program and update text morning slot - do-not-merge

Roderick-Jonsson:
- change Registration and Breakfast to Regisitration and Welcome coffee
- remove the sponsor slot from the program

### DIFF
--- a/content/events/2017-zurich/program.md
+++ b/content/events/2017-zurich/program.md
@@ -32,7 +32,7 @@ type = "event"
         <time>08:00-09:00</time>
       </div>
       <div class = "col-md-6 box">
-        Registration Breakfast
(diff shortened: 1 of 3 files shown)


MikeRosTX:
@Roderick-Jonsson why don't you want it merged?  Just curious since it in the title of the PR.  Or is it a mistake?

Roderick-Jonsson:
@MikeRosTX We didn't want to merge this as henri and jorge are blank speakers we temporary filled out. We would then send them an e-mail with the link to the netlifly site and if they still don't respond we would push the updated descriptions to the live site. As you have merged the pull request, we will just work with how it is now.

I work fixed times on devopsdays so the answer came a late this time.

PR #1989, opened 2017-03-13
[ATL-2017] schedule update

cacorriere:
Schedule is live


MikeRosTX:
Just a second while I review it before merging.

PR #1997, opened 2017-03-14
Add gulp tasks

mattstratton:
This adds the gulp tasks and netlify configuration needed for the new theme. This should not affect use of the current theme.


devopsdays-service:
@mattstratton, thanks! Based upon the content in this PR, we need a review from @mattstratton, @MikeRosTX, @phrawzty and @bridgetkromhout

PR #2012, opened 2017-03-14
Get master merged down into test

mattstratton:
Test is 10 commits behind master. This should get it in sync?


devopsdays-service:
@mattstratton, thanks! Based upon the content in this PR, we need a review from @mattstratton, @MikeRosTX, @phrawzty and @bridgetkromhout

PR #2031, opened 2017-03-16
Adding logo squares for upcoming events

bridgetkromhout:
I'm adding plausible square logos for some upcoming events so we don't have quite so many repeating brains on the front page. (The brain is the default now if an event hasn't chosen a specific square logo to use.)

Before: 
screen shot 2017-03-15 at 9 42 19 pm

After:

screen shot 2017-03-15 at 9 42 47 pm

If you'd like to use a different square logo on the new front page that we started using with today's theme release, please submit a PR replacing the ones I'm adding in this PR. /cc @phrawzty @nathenharvey @peterjshan @cacorriere


peterjshan:
@bridgetkromhout I like it, thanks!

PR #2039, opened 2017-03-16
[Paris-2017] adjust style to avoid auto-caps

phrawzty:
The default style for `##` auto-capitalises each word, which looks horrendous in French. 😉  This makes it hurt less.

`r?` @jooooooon

### DIFF
--- a/content/events/2017-paris/welcome.md
+++ b/content/events/2017-paris/welcome.md
@@ -25,7 +25,7 @@ discussions from industry practitioners and taste-makers.
 
 **The principal language of this conference will be French.**
 
-## Pour plus d'informations, veuillez visiter notre site principal: **https://www.devopsrex.fr/**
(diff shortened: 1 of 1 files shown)


mattstratton:
Just for my reference, can you give me an example page?

mattstratton:
Oh I see it now.

I'm not sold that is a good styling choice myself - I think we will fix it a different way. Feel free to use this one, but I'm thinking that we can adjust this.

phrawzty:
> Just for my reference, can you give me an example page?

Literally any page that uses headers of any type. :)

phrawzty:
p.s. 2017-seattle saw this problem and fixed it this way already. 😉

mattstratton:
yeah, that's bootstrap. Easy enough to override. I'll file a bug and fix that; I don't think I did that intentionally.

mattstratton:
Oh, I did do it on purpose for some reason. I'll put in a bug to fix that and override it where I need them to be capitalized (there are places that it makes sense at a higher level I think?)

jooooooon:
👍  Thanks @phrawzty

PR #2047, opened 2017-03-17
Fix broken sed commands on ubuntu

Eronarn:
Couldn't get the add_new_event script to work on 16.04; turned out the SEDCMD wasn't used everywhere in the script.


devopsdays-service:
@Eronarn, thanks! Based upon the content in this PR, we need a review from @mattstratton, @MikeRosTX, @phrawzty and @bridgetkromhout

PR #2048, opened 2017-03-17
[BOS-2017] Create Boston event!

Eronarn:
We currently are 5 organizers + 2 advisors (returning organizers from last year).


Eronarn:
(realized I had cherrypicked into an old master, re-pushed.)

Eronarn:
Hmm, weird. This builds fine for me locally... any pointers?

data/events/2017-boston.yml
```diff
@@ -0,0 +1,126 @@
...
+  - name: "Matthew Williams"
+    twitter: "technovangelist"
+    employer: "Datadog"
+    github: "technovangelist"
+    facebook: "https://www.facebook.com/technovangelist"
+    linkedin: "https://www.linkedin.com/in/technovangelist"
+    website: "http://technovangelist.com"
+    image: "MattOnBlue2.png"
+    bio: >
+      Matt Williams is an Evangelist at Datadog. He is passionate
```
mattstratton: This is the problem I think. Need to be in quotes. Look at 2017 chicago for example.

Eronarn: Huh, that's valid YAML syntax for a multi-line string - in fact you actually can't wrap that style in quotes. So I'm really not sure why it would be barfing, but I can change it. Thanks!

mattstratton:
I don't know why I created a new PR, but check out #2049 

https://deploy-preview-2049--devopsdays-web.netlify.com/events/2017-boston/

If this looks good to you, we can merge this one and drop your other one.

mattstratton:
Handled by #2049

PR #2054, opened 2017-03-17
[SEA-2017] Update sponsor.md to trigger deploy

kmugrage:
Whitespace change to trigger deploy


bridgetkromhout:
Fwiw I don't think whitespace changes trigger a deploy :/ but we'll see!

PR #2071, opened 2017-03-20
[KC-2017] Updated organizer info

barkerd427:


MikeRosTX:
@barkerd427 I'm merging, but some images got worse and others better. They need to be 600x600, or at minimum 300x300.

PR #2078, opened 2017-03-21
[OSL-2017] changed organizer images to 300x300

steinim:


mattstratton:
They don't need to be 300x300 - they'll be resized to that and set for retina at build into master.

Did you make them smaller or bigger? You can do 600x600 and they'll look a little odd in deploy preview/local but they'll build just fine.

steinim:
I made them bigger.

PR #2088, opened 2017-03-21
Change build to use hugo 0.19

### LINKED ISSUE #1845: Update to hugo 0.19
We are NOT going to do this right now, but when we do so, one very useful thing is going to be adding some configuration to not generate pages we don't need (specifically taxonomy pages, like tags and categories).

Just referencing this here for now so we don't forget it. 

See http://bepsays.com/en/2017/02/27/hugo-019/

Specifically:

--
With Hugo 0.19, you can control this behaviour by turning off page types you do not want ( [#2534](http://github.com/spf13/hugo/issues/2534) ). In its most extreme case, if you put the below setting in your `config.toml`, you will get nothing!:

```disableKinds = ["page", "home", "section", "taxonomy", "taxonomyTerm", "RSS", "sitemap", "robotsTXT", "404"]```

--

mattstratton:
Change netlify build to use hugo 0.19

Fixes #1845

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -10,7 +10,7 @@ If you'd like to edit a specific devopsdays event site (and/or contribute code),
 
 ### Quick Overview
 
-1. Install [Hugo v0.18+](http://gohugo.io).
+1. Install [Hugo v0.19+](http://gohugo.io).
 1. Fork this repo.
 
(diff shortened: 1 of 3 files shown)


mattstratton:
FWIW, appveyor and travis have been using 0.19 for quite some time and we have not seen errors.

PR #2089, opened 2017-03-22
Update Docker scripts for hugo 0.19

mattstratton:
Updates the two docker scripts for launching containers using hugo 0.19


mattstratton:
This is probably fine, but I'm doing one local test before merging.

PR #2103, opened 2017-03-23
[Edinburgh-2017] Added event dates and location information

mjslane:


MikeRosTX:
@mjslane once you get the date fields updated on the welcome.md file so that TBD is removed, we'll proceed to merge it.  See image below:

edinburg

mjslane:
@MikeRosTX Thanks for pointing this out. I've updated the welcome.md page now.

PR #2113, opened 2017-03-24
[OSL-2017] Added another organizer

steinim:


MikeRosTX:
@steinim where is **alexandru-albu.jpg**? Did it get uploaded?  Because it's not showing up.

steinim:
Forgot to add it. Did it now.

MikeRosTX:
@steinim says the image for Alex is 600x600 but in the Deploy preview it's not adjusting. If after the merge it continues to be the same displayed size, then @steinim will resize to 300x300 like the others.

PR #2123, opened 2017-03-25
Adding speaker slug

bridgetkromhout:
Before: 

screen shot 2017-03-25 at 6 04 35 am

After:

screen shot 2017-03-25 at 6 04 56 am


bridgetkromhout:
Heads up @tavisto (unsure if this was a script failure or other). /cc @mattstratton in case this is a common issue.

tavisto:
Ahh thank you. I think that might have been me and not the script.

PR #2124, opened 2017-03-25
Trivial change for build

bridgetkromhout:
The sort order in the left sidebar is wrong again (that mapping issue). Trying to trigger a build.

screen shot 2017-03-25 at 9 27 16 am


mattstratton:
You can actually just kick a build directly in Netlify if it has been merged to master.

PR #2128, opened 2017-03-26
consolidating tracking-URL sponsors

bridgetkromhout:
I noticed we were getting a proliferation of duplicate sponsor entries so people could add event-specific tracking URLs. Happily, @mattstratton added a feature allowing tracking URLs (https://github.com/devopsdays/devopsdays-theme/issues/398), so now we don't need to do this. 

On the bright side, this gave us some larger sponsor images, so I'm replacing these two smaller standard ones.

Heads up to @cacorriere @mattk42 @davidasnider @lsc so that you know why this changed on your sites. The tracking URL will still work fine - but you now don't need a different sponsor id.


lsc:
Thanks for the heads up @bridgetkromhout!

PR #2132, opened 2017-03-27
[AMS-2017] - Add Red Hat as official sponsor of the devopsdays AMS beer

yvovandoorn:
Confirmed Red Hat as official sponsor today with devopsdays AMS. Adding to event page.


MikeRosTX:
Give me a moment to review while it's going through the checks.

PR #2136, opened 2017-03-27
[2017-Hartford] Adds CFP dates, announce & link

crayzeigh:
Adds the dates for CFP open, close, announce and adds link to papercall.io for our CFP submissions.


MikeRosTX:
Reviewing now, merging shortly.

MikeRosTX:
Hey @crayzeigh it doesn't look like it's showing up anywhere because the following was removed which were [https://github.com/devopsdays/devopsdays-web/blob/master/data/events/2017-hartford.yml#L17-L18](https://github.com/devopsdays/devopsdays-web/blob/master/data/events/2017-hartford.yml#L17-L18) and should be change to **true**:

cfp_open: "false" 
cfp_link: "" #if you have a custom link for submitting proposals, add it here. This will control the Propose menu item as well as the "Propose" button

PR #2150, opened 2017-03-28
[DC 2017] add InfluxData as gold sponsor

markcornick:


MikeRosTX:
@markcornick I think you forgot to upload their new attached logo from his email.  ;-)

markcornick:
I got Admonished™ over changes to existing sponsor logos last year, but I’ll go ahead and add the new one.

MikeRosTX:
Understood @markcornick, but the CMO is requesting all cities they're sponsoring change their logo to the new one. Since they're paying ;-) we should respect their request.

bridgetkromhout:
> I got Admonished™ over changes to existing sponsor logos last year, but I’ll go ahead and add the new one.

lol, probably by me - fwiw the thing we're trying to prevent is ret-conning the past. Since this change is all for future events, it's fine that we're changing it on all upcoming events without going through the namespacing/archiving exercise we'd otherwise need to do.

PR #2154, opened 2017-03-28
[CPT-2017] Adding the proposals page

cobusbernard:
Enables the propose page, using an embedded Google Form for the submissions.

### DIFF
--- a/data/events/2017-cape-town.yml
+++ b/data/events/2017-cape-town.yml
@@ -8,9 +8,9 @@ description: # A short blurb of text to describe your event, defaults to common 
 startdate: 2017-11-06
 enddate: 2017-11-07
 # Leave CFP dates blank if you don't know yet, or set all three at once.
-cfp_date_start: 2017-02-01 # start accepting talk proposals.
-cfp_date_end: 2017-04-30 # close your call for proposals.
-cfp_date_announce: 2017-05-07 # inform proposers of status
+cfp_date_start: 2017-06-01 # start accepting talk proposals.
+cfp_date_end: 2017-07-31 # close your call for proposals.
+cfp_date_announce: 2017-08-08 # inform proposers of status
 
 # Location
 #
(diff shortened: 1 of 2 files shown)


data/events/2017-cape-town.yml
```diff
@@ -8,9 +8,9 @@ description: # A short blurb of text to describe your event, defaults to common
 startdate: 2017-11-06
 enddate: 2017-11-07
 # Leave CFP dates blank if you don't know yet, or set all three at once.
-cfp_date_start: 2017-02-01 # start accepting talk proposals.
-cfp_date_end: 2017-04-30 # close your call for proposals.
-cfp_date_announce: 2017-05-07 # inform proposers of status
+cfp_date_start: 2017-06-01 # start accepting talk proposals.
```
bridgetkromhout: Are you sure you want to link to a proposal form and then claim you're not going to accept talks until June?

cobusbernard: We've already had a few submissions, our initial planning was to only open it up later, but doubt that would prevent people from sending us emails. By setting the opening date, we hope to manage expectations w.r.t. when we do our selection. 

Any advice on how to handle eager, early submissions?

bridgetkromhout: Set an auto-reply on the google form that sends them a copy of their submission and re-iterates that you'll be reviewing submissions starting on $date. With that mechanism I was able to open Minneapolis submissions in December and I just started taking a look at the results this week. :)

bridgetkromhout: Also, the current logic in https://github.com/devopsdays/devopsdays-theme/blob/master/layouts/partials/speaking.html#L9 assumes that if you have a CFP end-date listed, then it's open. So you'll still appear on https://www.devopsdays.org/speaking/ even if you've tried to push the open date forward, unless we decide to support this (so please open an issue in the theme repo if you think it's actually a use case we want).

cobusbernard: Thanks! 🙏  
We've updated the Google Form to have it at the top and in the response to make it crystal clear when we will review submissions. Can we have this merged please?

bridgetkromhout:
Hi, @cobusbernard - @mattstratton fixed it (see https://github.com/devopsdays/devopsdays-theme/issues/448) so this won't show up on https://www.devopsdays.org/speaking/ until your open date.

adrianmoisey:
Awesome! Thanks.

PR #2156, opened 2017-03-28
No longer configurable

bridgetkromhout:
We no longer configure event_logo in this fashion.


devopsdays-service:
@bridgetkromhout, thanks! Based upon the content in this PR, we need a review from @mattstratton, @MikeRosTX, @phrawzty and @bridgetkromhout

PR #2160, opened 2017-03-28
[PDX-2017] The Influx folks asked us to update their logo. Right on

BillWeiss:
Looks the same to me, but, they asked...


MikeRosTX:
@BillWeiss when I update DC it update all the sites Influxdb chose to sponsor. So I'm closing and not merging this request.

mattstratton:
I wish it was a little bigger...but I'm not going to worry about it too much. At least 200px wide is recommended. If you have a higher-res version they sent you, get it to me somehow and I'll make some sponsor file magic, but I'm okay with this merge now.

mattstratton:
IMHO, this is a small enough difference in the logo file that it's not rewriting history, so I wouldn't mind, but if we are gonna do it, I want a higher resolution file :)

mattstratton:
What I really don't want to do is add another image file (`influxdb-pdx.png`) for what appears to not be a change at all; the blast radius to other events seems very small, vs the impact to the repo of having yet another file to process.

mattstratton:
given the minimal blast radius, I'm going to re-open this PR, but @BillWeiss I would like you to update it with a higher res version of the logo (if they provided you with one) and then we can merge it.

bridgetkromhout:
Ah, I see what's going on here. So, this was actually already updated in https://github.com/devopsdays/devopsdays-web/pull/2151. The sponsor didn't understand that they don't have to contact every event. That's why we can't see the change - because it already happened.

Also, as I explained in https://github.com/devopsdays/devopsdays-web/pull/2150#issuecomment-289810970 we don't change the past. None of these events are past, so the change is fine as it stands. We don't need to rename/keep the old logo. See the discussion in https://github.com/devopsdays/devopsdays-web/pull/503 for how we preserve history when updating a logo that was used on an event that's already started/is past.

mattstratton:
So basically, we can close this PR after all, unless @BillWeiss has a higher res versionto supply :)

PR #2161, opened 2017-03-28
[DC 2017] add eGlobalTech as platinum sponsor

markcornick:


mattstratton:
Don't sweat the failing travis check...it's doing that thing where it cannot install hugo again. Stupid github.

PR #2177, opened 2017-03-30
[PDX-2017] Add another CFP link, just in case

BillWeiss:
I couldn't figure out how to use the cfp_link variable set in data. If that's trivial, would love to see it :)


MikeRosTX:
Marked **do-not-merge** waiting on reply from #website.

mattstratton:
:shipit:

PR #2212, opened 2017-04-02
Removal of unused proposal pages and notes.

bridgetkromhout:
This looks like it's left over from when we thought we'd create such a shortcode. None of these events where I'm removing the file have proposals directories. Seattle and Toronto do, so I'm just removing the misleading "TBD". I'm reasonably certain this will have no ill effects but I'd appreciate a sanity check by @mattstratton.


mattstratton:
LOL @ the TBD. Well, merging!

PR #2282, opened 2017-04-14
London logo updates

bridgetkromhout:
Hi, @kramos - I'm moving your logo to the right location, and also added the version that will show up on the front page of the site.


kramos:
Thanks @bridgetkromhout 
When I checked checked on my phone, it prompted me to raise this to align the images: https://github.com/devopsdays/devopsdays-theme/issues/483

PR #2284, opened 2017-04-15
Script to generate new organizer fields (+README update)

bridgetkromhout:
This is some basic bash that generates the entries for new organizers. I minimally updated the README too. This is MVP and doesn't prompt for every conceivable field; we could update the README to link to the relevant doc. @mattstratton, please take a look and see if this is okay to add as MVP.

### DIFF
--- a/utilities/README.md
+++ b/utilities/README.md
@@ -76,9 +76,14 @@ Guidelines regarding sponsor logo files and formatting:
 
 All logos will be resized at release time, to 200px wide. Versions for high-density display (ex. Retina) will also be created. It is recommended that you do not use a sponsor logo that is smaller than 200px wide to keep from quality degradation at resize time.
 
-## Team members
+## Local Organizers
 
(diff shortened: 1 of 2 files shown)


MikeRosTX:
Looks great @bridgetkromhout.  If you may satisfy my curiosity :wink:,... Why not add the fields for **github**, **website**, and **facebook**?

bridgetkromhout:
> Looks great @bridgetkromhout. If you may satisfy my curiosity 😉,... Why not add the fields for github, website, and facebook?

I don't want people who are setting up their organizer list to stop midway because they don't have so much info at hand. Their local team can add extra fields later if needed.

MikeRosTX:
>I don't want people who are setting up their organizer list to stop midway because they don't have so
>much info at hand. Their local team can add extra fields later if needed.

Point understood. But you add the empty fields, so whomever is coming behind them can fill them out?

mattstratton:
@mikerostx imho adding blank fields would imply they are required. 

I think at this point since the tooling is "add only" and not "edit later" this is fine.

PR #2288, opened 2017-04-17
[DET-2017] Sponsorship info and package table

bchess:
Fleshing out the Detroit 2017 sponsorship page with packages, benefits, and pricing


bchess:
Hang on I want to add a link to a static PDF as well

PR #2323, opened 2017-04-24
minor Stockholm speaker-page edits

bridgetkromhout:
Checking out Stockholm (great lineup!) and I saw a few things it was easier to just fix than write up instructions to fix.

(It started with me noticing the speaker page wasn't linked, and I noticed a few more things after that.)

This needs edits and/or a 👍  from @lsc or @jessabe or another Stockholm organizer before merging.


lsc:
Looks fine, thanks @bridgetkromhout !

PR #2324, opened 2017-04-24
[Salt Lake City 2017] Adding more current links

bridgetkromhout:
Hi, @davidasnider - I noticed that your 2017 event on devopsdays.org was missing links that would help people find where your info actually lives. Please 👍  this if you want these links. Thanks!


davidasnider:
👍

PR #2394, opened 2017-05-09
Try hugo 0.19

mattstratton:
Due to issues with the build, this reverts to hugo 0.19 to see if that helps


mattstratton:
Reverting back to hugo 0.19 for now.

PR #2408, opened 2017-05-11
[Ohio 2017] Initial updates for setting the date and location

kjenkins19:
A few style updates since the new theme is in place.

Updating the dates, location, and list of organizers. 
Adding in location page.


MikeRosTX:
We will proceed with the merge because another global core organizer agreed to "grandfathered-in" the naming convention to be Ohio.  But moving forward in to 2018, it will most likely need to be named by the city it's help in as oppose the state.

MikeRosTX:
@kjenkins19 based on the **contact** page I'm not sure you've read [https://github.com/devopsdays/devopsdays-web/blob/master/themes/devopsdays-theme/README.md](https://github.com/devopsdays/devopsdays-web/blob/master/themes/devopsdays-theme/README.md) to take advantage of all the new rich feature we now have to offer with the new theme that's in place.  Looks like you're following the style of the old theme.

PR #2417, opened 2017-05-11
Use binrc for hugo

mattstratton:
This change removes the dependency on the embedded hugo binary and uses the mappings in `netlify.toml`.


bridgetkromhout:
Let's get this pushed through while watching and see if it breaks anything?

PR #2493, opened 2017-06-02
[DET-2017] Additional Registration and Propose options

broamski:
Add registration and proposal links to the Welcome page


broamski:
Thanks!

PR #2500, opened 2017-06-05
Use binrc to install hugo

mattstratton:
This changes the Hugo installation to pull down binrc, and then use binrc to install hugo.


bridgetkromhout:
Closing per @mattstratton for now.

PR #2506, opened 2017-06-06
[SIN-17] Initial version of event page

ctrabold:
This commit creates the initial event page with a basic set of pages. 

The goal is to have a landing page for the event that we can use when reaching out to sponsors or speakers.

No changes on the global scope are made and the event is not yet added to the list of upcoming events as we are still waiting for a final confirmation of the venue.

### DIFF
--- a/data/events/2017-singapore.yml
+++ b/data/events/2017-singapore.yml
@@ -1,0 +1,94 @@
+name: "2017-singapore" # The name of the event. Four digit year with the city name in lower-case, with no spaces.
+year: "2017" # The year of the event. Make sure it is in quotes.
+city: "Singapore" # The displayed city name of the event. Capitalize it.
+event_twitter: "devopsdaysSG" # Change this to the twitter handle for your event such as devopsdayschi or devopsdaysmsp
+description: "DevOpsDays is coming to Singapore!" # Edit this to suit your preferences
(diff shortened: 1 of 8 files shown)


data/events/2017-singapore.yml
```diff
@@ -0,0 +1,94 @@
+name: "2017-singapore" # The name of the event. Four digit year with the city name in lower-case, with no spaces.
+year: "2017" # The year of the event. Make sure it is in quotes.
+city: "Singapore" # The displayed city name of the event. Capitalize it.
+event_twitter: "devopsdaysSG" # Change this to the twitter handle for your event such as devopsdayschi or devopsdaysmsp
+description: "DevOpsDays is coming to Singapore!" # Edit this to suit your preferences
+ga_tracking_id: "" # If you have your own Google Analytics tracking ID, enter it here. Example: "UA-74738648-1"
+
+# All dates are in unquoted 2017-MM-DD, like this:   variable: 2016-01-05
+startdate: 2017-10-25 # The start date of your event. Leave blank if you don't have a venue reserved yet.
```
MikeRosTX: @ctrabold please remove the **startdate** & **enddate**, because until you have a confirmed signed contract with the venue we cannot merge this PR. But we can merge without date just as a placeholder.

ctrabold: @MikeRosTX noted! removed the dates

PR #2518, opened 2017-06-06
[RAL-2017] Initial Program Layout

chris-short:
Time to make the schedule!


phrawzty:
Hello,

Looking at the Netlify preview, it looks like the programme is a mix of bolded and unbolded elements. Is that intended?

PR #2530, opened 2017-06-07
[OSL-2017] Added keynote speakers and a preliminary program.

steinim:


mattstratton:
I love this way of teasing it out and having the links for proposals in the program!!!

PR #2534, opened 2017-06-08
Elijah is out, sadly

BillWeiss:


bridgetkromhout:
Removed from email list. Not on Slack.

Issue #2559, opened 2017-06-12
Date in speaker files needs to be wrapped in quotes


mattstratton:
The script needs to be updated - the latest hugo is picky about the dates being wrapped in quotes for frontmatter.

bridgetkromhout:
Looks like the script has a vestigial datestamp that the template no longer uses. I've submitted a correction. Please let me know if this doesn't clear things up from your POV, and if so, what else you'd like to see.

mattstratton:
Apparently in the latest version they changed the TOML parser, so it's more strict on a few things...it only seems to have affected Zurich's speakers.

bridgetkromhout:
Okay, I _think_ we are good here - I'll leave this up to you as to whether it's resolved.

mattstratton:
Seems okay to me, so I'll close it.

PR #2561, opened 2017-06-12
Remove unneeded requests

bridgetkromhout:
Speaker creation shouldn't ask for data it doesn't use. The "datestamp" frontmatter field no longer exists in those templates, so this sed command is a no-op.

This is meant to address https://github.com/devopsdays/devopsdays-web/issues/2559 - we can bikeshed over there as to whether it solves what that issue was meant to solve.


mattstratton:
I am 99% sure that the page will work without a date; so it should be okay.

We'll try it and see :)

PR #2582, opened 2017-06-15
[AMS-2017] openshift workshop second speaker

ams0:
Added @sterburg as a second speaker/mentor to the Openshift workshop.


mattstratton:
This fixed it. I still don't love how the workshops are displaying (they got built as a separate file than just another program element, as @yvovandoorn and I bikeshed about) but this change is building properly :)

PR #2590, opened 2017-06-17
[AMS-2017] Add tracking link to OpsGenie sponsorship

thiagoavadore:
OpsGenie Marketing team wants to track clicks on individual event links - starting on [AMS-2017].


bridgetkromhout:
We've done this a number of times before. Other than using a little extra space for images it has minimal ill effects.

yvovandoorn:
👍

PR #2653, opened 2017-06-29
Correcting Tel Aviv organizer images

bridgetkromhout:
Hi, @gilzellner - I corrected some issues to make your images show up (and be in the correct size/aspect ratio).


gilzellner:
Hi @bridgetkromhout ! thanks for the assistance.
I'm slightly confused, was the only reason it wasn't showing up the aspect ratio ?

mattstratton:
I started writing a long reply, and github ate it.

The problem is that the other organizers have a refernece to an image in your data file, but no corresponding image file in `static/events/2017-tel-aviv/organizers/`, so they were showing as broken images. Bridget removed the references to those files.

I'm not sure why @bridgetkromhout changed the two existing image files though.

bridgetkromhout:
> I'm not sure why @bridgetkromhout changed the two existing image files though.

For posterity: I resized one to make it our minimum size, and cropped the other to make it square. I also needed to change `static/events/2017-telaviv/organizers/` to have the required hyphen: `static/events/2017-tel-aviv/organizers/`

PR #2670, opened 2017-07-02
[TLV-2017] adding more organizers and photos + sponsor photos

gilzellner:


bridgetkromhout:
Please remove the cloudify update from this PR and put them in their own PR - handling updates to sponsors requires not retconning history and is a little tricky. It's easier to handle it separately.

gilzellner:
put the other picture back, not sure why its seen as different.

data/sponsors/cloudify.yml
```diff
@@ -1,2 +1,2 @@
 name: "Cloudify"
-url: "http://getcloudify.org/"
+url: "http://cloudify.co"
```
bridgetkromhout: Normally I'd ask for a new version with a URL change - since the old URL redirects to this new one, though, it doesn't much matter.

bridgetkromhout:
Looks good! If you decide to change the picture, the old one needs to be renamed and preserved. Thanks.

PR #2718, opened 2017-07-13
[KC-2017] Add program layout and organizer bio

aaronblythe:
Add the program layout for the event
Add organizer bio for organizer Bert Weidt


bridgetkromhout:
Hey, please take a look at https://deploy-preview-2718--devopsdays-web.netlify.com/events/2017-kansascity/program/ and make sure this is what you want? (tl;dr: you can't have placeholder names for the non-ignite talks - they need to be real - so `-` is what shows.)

aaronblythe:
Thank you for reviewing that closely. At this point, this IS what we want.

It is not currently linked from the welcome page yet, however this gives our organizers a visual in the final format to discuss, rather than doing the work to lay this out in Excel/Powerpoint, then eventually have to come back and format for this view.

If this doesn't cause issues for the rest of the site we would like this merged while we contact/confirm our speakers so we can go back and forth on the layout of the two days.

Issue #2717, opened 2017-07-13
hugo version issue


aaronblythe:
Feel free to close, this is merely because I did not see these errors documented, and documentation on hugo version may need updating.  TL;DR upgrade hugo.

Tonight I started working and received these errors even though a couple months ago everything was working:

```
>> hugo server -w --baseUrl="http://localhost:1313"
Started building sites ...
ERROR: 2017/07/12 23:30:06 template.go:495: Failed to add template theme/events/single.html in path /Users/ablythe/code/opensource/devopsdays-web/themes/devopsdays-theme/layouts/events/single.html: template: /Users/ablythe/code/opensource/devopsdays-web/themes/devopsdays-theme/layouts/_default/baseof.html:15: function "now" not defined
ERROR: 2017/07/12 23:30:06 template.go:495: Failed to add template theme/index.html in path /Users/ablythe/code/opensource/devopsdays-web/themes/devopsdays-theme/layouts/index.html: template: /Users/ablythe/code/opensource/devopsdays-web/themes/devopsdays-theme/layouts/_default/baseof.html:9: function "now" not defined
ERROR: 2017/07/12 23:30:06 template.go:495: Failed to add template theme/partials/events/cta.html in path /Users/ablythe/code/opensource/devopsdays-web/themes/devopsdays-theme/layouts/partials/events/cta.html: template: theme/partials/events/cta.html:24: function "now" not defined
...
ERROR: 2017/07/12 23:30:06 template.go:529: template: theme/partials/events/cta.html:24: function "now" not defined
ERROR: 2017/07/12 23:30:06 template.go:529: template: theme/partials/footer.html:50: function "now" not defined
ERROR: 2017/07/12 23:30:06 template.go:529: template: theme/partials/future.html:3: function "now" not defined
ERROR: 2017/07/12 23:30:06 template.go:529: template: theme/partials/head/seo/open_graph.html:28: function "now" not defined
...
=============================================================
Your rendered home page is blank: /index.html is zero-length
 * Did you specify a theme on the command-line or in your
   "config.toml" file?  (Current theme: "devopsdays-theme")
 * For more debugging information, run "hugo -v"
=============================================================
[...]

PR #2721, opened 2017-07-13
Update references to versions of Hugo

mattstratton:
There are a few misleading/inaccurate references to Hugo versions, and contributing files. This PR should fix most of them.


mattstratton:
Fixes #2717

PR #2732, opened 2017-07-16
[INDY-2018] Add new event for Indianapolis in 2018

ghoneycutt:


bridgetkromhout:
Awaiting confirmation from existing organizer & organizer list update on Slack & email.

bridgetkromhout:
Confirmation received.

PR #2756, opened 2017-07-19
[PHX-2017] Added registration and sponsor links, updated links, cleaned up text.

ericharris:


phrawzty:
An observation regarding `static/events/2017-phoenix/sponsorship.png`: it looks like it's just a screenshot from a spreadsheet, which is fine, but you can see the wavy typo underlines and other elements. I've no problem merging this, but you might want to clean it up in the future. 😺

ericharris:
Thanks! The screenshot is short term fix...just got way too late to deal with html tables and wanted to get the info on the site. :)

PR #2775, opened 2017-07-21
[PDX-2017] We've agreed to refund Rigor on their sponsorship

BillWeiss:


BillWeiss:
No hard feelings, there was just a misunderstanding of what they were getting for their money, and we'd hate to have a sponsor feel like it was a waste of an event.

PR #2789, opened 2017-07-23
[SIN-17] Add community sponsors

ctrabold:
- We have our first GOLD sponsor!
- We shout out to the Singaporean community and publish their logos on the website for community building.
- We set the CfP attribute to `true` to enable CfP


streichsbaer:
Looks good to me!

PR #2841, opened 2017-07-28
[HFD-2017] Fix sponsor levels, remove Serena & Scott as orgs, add José as org

crayzeigh:
Adjusted the sponsor levels to reflect the actual sponsor levels for Hartford while adding VictorOps as a Mug Sponsor. 

Fixed our organizers to match our actual organizers after the recent changes. Serena and Scott are no longer organizing for Hartford, José Padilla now is.


phrawzty:
Hello!

> Serena and Scott are no longer organizing for Hartford, José Padilla now is.

1. Should Serena and Scott also be removed from Slack and/or the mailing list?
1. Should José be added to either or both of those resources?

crayzeigh:
1. I don’t particularly mind if Serena and Scott stay in either or both resources, it was a matter of not having the time rather than something particularly negative. So I suppose that’s up to whatever the policy is for DevOpsDays (if there is one) They are non-active past organizers as opposed to actively organizing DevOpsDays Hartford.

2. José is in the slack already and I’m assuming that means he’s also on the email, but he should be on both if he’s not.



- Aaron

On Jul 28, 2017, at 11:24, Daniel Maher <notifications@github.com<mailto:notifications@github.com>> wrote:


Hello!

Serena and Scott are no longer organizing for Hartford, José Padilla now is.

  1.  Should Serena and Scott also be removed from Slack and/or the mailing list?
  2.  Should José be added to either or both of those resources?

PR #2861, opened 2017-08-01
[Nashville-2017] Add Data Blue sponsor

ebpitts:


data/sponsors/datablue.yml
```diff
@@ -0,0 +1,2 @@
+namex: "Data Blue"
```
bridgetkromhout: This should say `name: ` not `namex: `

Also, there should be an image for a newly added sponsor.

ebpitts:
Oops, mea culpa for not `git add` ing the image.

Surprisingly, when I ran hugo locally it still added the sponsor without complaint despite that field being called `namex` rather than `name`.

Pushing a fixed version now.

PR #2940, opened 2017-08-14
[DET-2017] Introducing program

bchess:
All of the talks and speaker info. We still need to add more speaker photos and will likely shuffle around the speaker order.


kmugrage:
You may want to standardize on [500px for the speaker images](https://github.com/devopsdays/devopsdays-web/blob/master/utilities/README.md)... they will look much better across devices.

PR #2943, opened 2017-08-15
[DET-2017] Adding silver sponsor

joearnet:
Adding microsoft as a silver sponsor.


bridgetkromhout:
The netlify issue looks like an intermittent failure.

PR #2944, opened 2017-08-15
[Nashville-2017] Add shadow-soft to sponsors

ebpitts:


ebpitts:
Oops, I misread their sponsorship level.

PR #2952, opened 2017-08-15
[SEA-2018] Seattle 2018 dates set

kmugrage:
Confirmed venue. Setting the dates for Seattle 2018


kmugrage:
Would prefer someone else merge this since it's a) mine and b) technically a new event

PR #2964, opened 2017-08-17
RAL-2017: Outsourcing Program; Not worth the effort to do it natively

chris-short:
Don't take offense but the speaker/program workflow needs some improvement. I have ideas but no time to implement.


mattstratton:
Is the comment necessary?

PR #2967, opened 2017-08-17
Refactored gocd to thoughtworks-gocd to reduce confusion

kmugrage:
Trying to reduce confusion from ThoughtWorks sponsorships. Most of them are for GoCD, but occasionally the parent company sponsors on their own. 

For GoCD I created gocd.yml some time ago, but people understandably defaulted to one of the thoughtworks files. The (not perfect) idea here is to at least include the company name in the yml filename so that it's more visible.


devopsdays-service:
@kmugrage, thanks! Based upon the content in this PR, we need a review from @mattstratton and @bridgetkromhout

kmugrage:
Added the do-not-merge label until @mattstratton and @bridgetkromhout can review

PR #2971, opened 2017-08-17
Updating spreedly

bridgetkromhout:
After chatting with @mattstratton I attempted this to rewrite the path to a sponsor I updated, in the old static files.

This is what I did:
```
grep -rl '2016-spreedly' . | xargs sed -i '' 's/2016-spreedly/spreedly/g'
```

I can't get a good picture of whether this looks right locally (all the paths to all the sponsor images are non-working, not just this one, in a local preview).

@mattstratton, what do you think? Thanks.


mattstratton:
This is hard to test in deploy preview, because the older events get redirected to www.

It looks okay to me. My suggestion is that we merge it in, test it once it is deployed, and we can easily roll it back.

PR #2979, opened 2017-08-18
specify min number of organizers

kmugrage:
Small text update to specify minimum number of organizers required


devopsdays-service:
@kmugrage, thanks! Based upon the content in this PR, we need a review from @mattstratton, @MikeRosTX, @phrawzty and @bridgetkromhout

kmugrage:
Umm, woah. Is this page part of the theme?

bridgetkromhout:
Perfect wording change! And no, it's not part of the theme - but @mattstratton flagged some pages to make sure we'd noticed if people tried to update them. :)

PR #2985, opened 2017-08-18
[DET-2017] Program updates

broamski:
Remove a speaker from the program, add information about evening event.


broamski:
Thanks!

kmugrage:
@broamski - FYI, the extra dash thing I mentioned above could be removed by using "custom" instead of "talk" for those entries. See the comments on the referenced bug.

PR #2989, opened 2017-08-18
[LON-2017] Updating pics and profiles and adding a sponsor

kramos:
Update Oli and Sasha's pics, Sasha's bio, and added Ticketmaster


MikeRosTX:
Merging because code looks good despite **deploy/netify** preview failed.

PR #3043, opened 2017-08-26
[Auckland-2017] Add sponsor logo for Datacom and Xero

flaviafm:
Adding sponsors Datacom as bronze level and Xero as silver level


kmugrage:
Hello. New sponsors should not have a year in their yml / logo file names anymore. Would you mind renaming these? Thanks!

PR #3058, opened 2017-08-29
[Auckland-2017] Fix Datacom Logo for square

flaviafm:
Making sponsor logo square to match others sponsors


bridgetkromhout:
For what it's worth, this is not required. We can handle non-square logos.

PR #3074, opened 2017-08-30
[TLV 2017] Correcting speaker and program layout

bridgetkromhout:
This is for @gilzellner to approve. I'm correcting some of the issues that were preventing their keynotes from showing up entirely as expected. Also, the images should be in static (speeds up rendering).


gilzellner:
Looks great! thanks!

PR #3075, opened 2017-08-30
Adding front-page squares for events missing them

bridgetkromhout:
Adding couple of logos for the front page that are present in the repo but weren't on the front page. /cc @electrofelix


electrofelix:
@bridgetkromhout thanks, sorry I missed that was needed.

PR #3082, opened 2017-08-31
[GAL-2017] Add image for logo to appear with links

electrofelix:
Adds a sharing version of the event logo


phrawzty:
I <3 that logo so much. :)

PR #3086, opened 2017-08-31
[GAL-2017] Fix sponsors display

electrofelix:
Missing tag was resulting in the style being applied incorrectly on the sponsors page.


electrofelix:
Seems that something else might be broken now, please don't merge.

electrofelix:
Must have been a lag between me making the correction and it showing up in my local preview of the site, hence missed that it broken part of the sponsors page. All looks good with the updated commit.

PR #3099, opened 2017-09-03
[2017-brasilia] Add sponsors

adrianovieira:
- Rancher Labs
- Configr


bridgetkromhout:
Hi, @adrianovieira - please do not put "2017" at the beginning of a sponsor name - we're moving away from using that.  Also, "rancher" with the image at https://github.com/devopsdays/devopsdays-web/blob/master/static/img/sponsors/rancher.png is the same company as the Rancher Labs you're trying to add. If Rancher asked you to use a different logo, do you know if they want it updated everywhere going forward, used only in .br, or something else?

adrianovieira:
Ok! I can update the name cuting the 2017 at the begining. I noticed that Rancher already has one logo there, but it is not the correct brand logo. Than I thought I could update it according to their brand rules (http://rancher.com/brand-guidelines/) which tell us to use "Rancher Labs logo'.

But I'll update all our logos according to your tip Bridget.

adrianovieira:
I hope everything is ok now.

bridgetkromhout:
Thanks. We'll merge this for now, and I'm hoping @kmugrage can follow up on the logo updating  issue, as I know he knows how to properly update logos if need be.

PR #3100, opened 2017-09-03
[Cape Town 2016] Add YouTube videos for each speaker

adrianmoisey:
Backfilling the old event with their videos. I got a little jealous of
the other past events looking cooler than ours :)

I'm not 100% sure if this is the best way to do this, so let me know if
I should change my approach.

Thanks web admins!


bridgetkromhout:
Spot-checked the first couple and LGTM.

PR #3105, opened 2017-09-04
Allow shell scripts to be called from anywhere

adrianmoisey:
Since the paths are relative in these scripts, it's assumed you are
calling them from `utilities/`. If you try calling them from somewhere
else, they will fail.

This change allows you to call them from anywhere, as it `cd`s you into
the right directory.


devopsdays-service:

bridgetkromhout:
Overall seems like a good plan. However I'm heading out of town without a laptop and cannot test this right now.

adrianmoisey:
No worries, there's no rush.

MikeRosTX:
Just curious, but shouldn't this PR be closed out and opened under [https://github.com/devopsdays/devopsdays-web/issues](https://github.com/devopsdays/devopsdays-web/issues)?

mattstratton:
@mikerostx why?

bridgetkromhout:
@MikeRosTX These scripts live in devopsdays-web so it's fine here.

I ran through and this works. (I assumed it would, but I was away from a laptop last week and so I couldn't test.) Thanks, @adrianmoisey!

PR #3106, opened 2017-09-04
Make hugo stop if you send Ctrl+C

adrianmoisey:
While running hugo under Docker, sending a Ctrl+C doesn't make it stop,
only the shell script stops.

(even thought Hugo says "Press Ctrl+C to stop")

Adding the `-i` to Docker changes this behaviour. Hitting Ctrl+C now
causes Hugo to stop, and the shell script to exit.

### DIFF
--- a/hugoserver.sh
+++ b/hugoserver.sh
@@ -9,4 +9,4 @@
 
 docker stop hugo-server
 docker rm   hugo-server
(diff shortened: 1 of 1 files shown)


bridgetkromhout:
Good call!

PR #3117, opened 2017-09-05
[RIGA-2017] Added Google Analytics tracking code

arthur-kovalenko:
Added Google Analytics tracking code


phrawzty:
The netlify preview is stuck but the PR looks good.

PR #3130, opened 2017-09-06
Fix sed commands to use correct placeholder for file

electrofelix:
Remove argument after '-i' as it is being recognized as the sed script to execute
rather than as an argument to the option flag.

### DIFF
--- a/utilities/add_sponsors.sh
+++ b/utilities/add_sponsors.sh
@@ -2,6 +2,16 @@
 
 set -e
 
+# Detect OS for correct 'sed' syntax
+OSNAME=`uname`
+SEDCMD(){
+  if [[ $OSNAME == 'Linux' ]]; then
+    sed -i "$@"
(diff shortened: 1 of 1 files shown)


devopsdays-service:
@electrofelix, thanks! Based upon the content in this PR, we need a review from @mattstratton and @bridgetkromhout

electrofelix:
Was this script perhaps written for Mac? The change I've made allows it to work on Ubuntu, but I'm not sure about cross platform compatibility with sed.

adrianmoisey:
I think it makes sense to copy what the other scripts are doing, take a look at https://github.com/devopsdays/devopsdays-web/blob/c2e36cb117669f27ff9282e76c40855acf6b112a/utilities/add_new_event.sh#L5-L13 as an example.

mattstratton:
Yeah, the -i is for x-plat to Mac. I want to get a +1 from @bridgetkromhout before merging and I am pretty sure she is AFK for a bit.

PR #3131, opened 2017-09-06
[HFD-2017] Update navigation

jpadilla:
- Remove propose since CFP is closed
- Move navigation to second position


jpadilla:
@MikeRosTX thanks! ✨

PR #3150, opened 2017-09-08
[HFD-2017] Update logo

jpadilla:


MikeRosTX:
That new log looks like **_The Heartbeat of DevOps in Hartford_**. 😉

PR #3159, opened 2017-09-10
Berlin 2017 Preliminary program online

range:
This adds the preliminary program and all already confirmed speakers to the berlin 2017 website.


bridgetkromhout:
Hi, @range - the "2x" images are generated in the build pipeline. Please remove those and add only one image for each speaker (as explained in https://github.com/devopsdays/devopsdays-theme/blob/master/REFERENCE.md#speaker-page-fields) - "It must be square, and 300px square, 600px square, or 900px square.".

range:
I hope that was enough?

PR #3170, opened 2017-09-11
[CHI-2017] Update Chicago location to have the proper floor

mattstratton:
Matt had put in the wrong floor. This fixes it.


kmugrage:
Preview failure was Netlify unable to allocate memory to run git. I can't imagine this could be a breaking change so merging.

PR #3239, opened 2017-09-19
[CHI-2017] updating Jerry's bio

jerrycattell:


mattstratton:
I approve this because I probably should have updated it before the event.

PR #3243, opened 2017-09-19
[TLV-2017] Sponsors

gilzellner:
[TLV-2017] Sponsors


data/sponsors/walkme.yml
```diff
@@ -0,0 +1,2 @@
+name: Walk Me
+url: https://walkme.com/
```
kmugrage: can you add a newline here please?

gilzellner: done

gilzellner: also, stuff like that I don't mind if you add on your own, its totally fine

PR #3249, opened 2017-09-20
[EDI-2017] Update sponsor details to the local site and changing my image to a jpg.

kevinhcross:


kmugrage:
@kevinhcross the NCR logo is going to look pretty bad at that resolution. It should be a minimum of 200px wide, 600px recommended. I don't _think_ it's a hard rule so let us know if you can't get a replacement and want us to merge as is.

kevinhcross:
@kmugrage I have replaced the image with a larger one.

PR #3257, opened 2017-09-21
[Cape Town 2017] Make DevOpsDays a bit more consistent

adrianmoisey:
DevopsDays vs DevOpsdays vs DevOpsDays

Our images all have `DevOpsDays`, so I've gone for that one.


phrawzty:
> Our images all have DevOpsDays, so I've gone for that one.

A good a reason as any. 😄

PR #3280, opened 2017-09-26
[DET-2017]Opening day preparation

broamski:
- Add missing presenter from the schedule
- Remove organizers employer
- Close registration


broamski:
Thanks!

PR #3286, opened 2017-09-26
[INDY-2018] Add C. Chalfant as team member

chalfant:
[INDY-2018] Add C. Chalfant as team member @ghoneycutt


ghoneycutt:
+1

bridgetkromhout:
Adding team members requires you to email info@devopsdays.org with their email address so we can sync Slack, email lists, and the website. We'll merge when we can do all three.

ghoneycutt:
Thank you!

PR #3294, opened 2017-09-27
Adding missing quote

bridgetkromhout:
Adding missing quote that was leading to Indy not showing up.


MikeRosTX:
Thanks for catching that @bridgetkromhout!  Need to nickname you EagleEye. 😉

PR #3314, opened 2017-09-29
[GAL-2017] Sponsor and organizer updates

electrofelix:


electrofelix:
email send with organizer email details

PR #3325, opened 2017-10-01
Remove reference to file that doesn't exist.

adrianmoisey:


devopsdays-service:

adrianmoisey:
Sorry, that's incorrect. The file does exist, github just links to the wrong path.

PR #3326, opened 2017-10-01
Fix sponsor example

adrianmoisey:
`add_sponsors.sh` expects this string to exist, because it replaces it
with the correct twitter account.
See
https://github.com/devopsdays/devopsdays-web/blob/9e96ad2f7990150a5b14f91975312aaa0d1cb0c3/utilities/add_sponsors.sh#L64

### DIFF
--- a/utilities/examples/data/sponsors/sponsorname.yml
+++ b/utilities/examples/data/sponsors/sponsorname.yml
@@ -1,3 +1,3 @@
 name: "SPONSORNAME"
 url: "URL"
-twitter:
+twitter: "TWITTER"


devopsdays-service:

bridgetkromhout:
Hmm. iirc this is data some people wanted to collect in case we started doing something with it... which we do not yet. Which suggests to me that we shouldn't collect it before we have a plan for it, because hilariously bad things happen when sponsors get acquired and don't secure their previous Twitter handles... https://twitter.com/govdelivery for example.

adrianmoisey:
Should `add_sponsors.sh` then be updated to not ask for twitter details?

mattstratton:
It's possible that we don't need to hack on this script too much more; I've made quite a bit of progress on `devopsdays-cli` which will replace all the scripts. In fact, I almost don't want to make the scripts change too much, as I'm working at feature parity for them. If something is totally busted in one of the scripts, by all means, let's fix it...but if they are "nice to have" things, let's make sure we include them in [devopsdays-cli](http://github.com/devopsdays/devopsdays-cli).

To be clear, while I said I have made quite a bit of progress, at this exact instant that codebase in a huge mess, so don't judge me :)

adrianmoisey:
Oh wow, I didn't know about `devposdays-cli`! That looks neat.

Here are the options I see with regards to this pull request:
1. The twitter section of `add_sponsor.sh` removed, along with the reference in the example. This solves what Bridget brought up, and brings this script in line with `devopsdays-cli`, and removes broken functionality.
2. Fix the twitter example template (ie, this pull request) and add that functionality to `devopsdays-cli`

By the sounds of it, option #1 satisfies both of your concerns. Let me know what you would like me to do.

mattstratton:
I would suggest option #1.

bridgetkromhout:
Went with option 1 - closing this.

PR #3349, opened 2017-10-05
[PHILLY 2017] Update sponsor logos.

peterjshan:
On the tin.

Also please note: I've namespace for a new googlecloud logo as per the sponsors request.

cc: @mattstratton


bridgetkromhout:
Temporary namespace. I'll follow up.

PR #3356, opened 2017-10-07
[Berlin-2017] add silver sponsor 'BurdaForward'

nmesstorff:
add silver sponsor 'BurdaForward'


phrawzty:
Hello! The sponsor image is of a pretty low resolution. I'm happy to merge, but if you'd like to update it with something a bit larger later (say [600x600 for retina displays](https://github.com/devopsdays/devopsdays-web/blob/master/utilities/README.md#sponsor-logos)), that'd be OK. :)

PR #3358, opened 2017-10-07
[Nashville-2017] Add Jeff Davis as organizer

ebpitts:


phrawzty:
Hello! Has this person also been added to the mailing list and Slack? If yes, great - if not, please make mention of their email address so that we can add them as well. This PR can be merged once that's cleared up.

bridgetkromhout:
Added to email and Slack.

PR #3367, opened 2017-10-08
Ohio 2017: The Schedule

kjenkins19:
The initial schedule as drawn up. Bios, abstracts, and schedule.
Note: I am currently pointing at existing images for speakers, as I know there has been much discussion in the past about not uploading too many images into the repo, and it seemed silly to not reuse what already exists.
I will be uploading the images for the rest of the speakers as I can find them. Saving to a later date. I can also upload images for the existing speakers during that time also if you don't want me pointing at the existing ones.

### DIFF
--- a/content/events/2017-ohio/program.md
+++ b/content/events/2017-ohio/program.md
@@ -1,0 +1,5 @@
++++
+Title = "Program"
+Type = "program"
+Description = "Program for devopsdays Ohio 2017"
++++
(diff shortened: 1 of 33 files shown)


bridgetkromhout:
Hi, @kjenkins19 - I understand your desire to reuse images! We can't introduce dependencies between events like this, though. Please correct this before we merge. Put images in your own namespace - thanks.

bridgetkromhout:
Ah, I was able to correct this PR in the online editor! Merging.

PR #3370, opened 2017-10-09
[GAL-2017] Add additional speaker to talk

electrofelix:
Add second speaker to the Reliability At Scale talk


bridgetkromhout:
You might want to add speaker bios and/or pictures to the speakers for this talk, and you also might want to rephrase the talk description to say "we" instead of "I", but this works!

PR #3374, opened 2017-10-09
[Nashville-2017] Add Advizex sponsor

ebpitts:
Rename references from 2015-Advizex to Advizex and add to Nashville sponsors.


bridgetkromhout:
Looks right.

PR #3394, opened 2017-10-12
[LA 2018] Add Chris Webber to team members.

irabinovitch:
[LA 2018] Add Chris Webber to team members.


bridgetkromhout:
Added back to Slack and email as well.

PR #3400, opened 2017-10-13
[AMS2018] - Set dates for CFP and event

yvovandoorn:
* Sets CFP date for early Jan
* Sets close for mid April
* Sets conference dates for end of June (back at the same spot!, things never change)
* Opens up sponsorship, but no active recruiting yet.


bridgetkromhout:
hey @yvovandoorn, I'll merge this without the (optional) "location" page, but since you do have your location you might want to also set that page up as usual. thanks!

PR #3403, opened 2017-10-13
[ZRH-2018] Adding Sponsors

Roderick-Jonsson:
Adding Pontine and Puzlle ITC as sponsors as well as the sponsor levels

There seems to be something off with the styling of the sponsor levels on our start page (Browser: Safari 10.1.2)
devopsdays 2018 zurich sponsor display problem


kmugrage:
@Roderick-Jonsson there is definitely something off, and it's not just Safari. Chrome looks the same.

bridgetkromhout:
The problem is coming in with this line: https://github.com/devopsdays/devopsdays-web/blame/master/content/events/2018-zurich/welcome.md#L107 - I'll merge and then fix it.

PR #3410, opened 2017-10-15
[2017-Salvador] Adding links & front-page image

bridgetkromhout:
Hi, @gomex - I'm updating the devopsdays.org page so it links better to your stand-alone site. That way, if people start out on devopsdays.org they can find your event. Please let me know if this is okay with you - thanks!


bridgetkromhout:
The link will look like this: https://deploy-preview-3410--devopsdays-web.netlify.com/events/2017-salvador/

bridgetkromhout:
Haven't heard back, but I think this is an overall improvement, so I'm merging - please submit a PR if you'd prefer it to be different. Thanks!

PR #3424, opened 2017-10-16
[TLV-2017] change sponsor logos

gilzellner:
[TLV-2017] change sponsor logos


kmugrage:
@bridgetkromhout @mattstratton - can you confirm this is the correct namespacing change?

bridgetkromhout:
I'll merge, then correct.

PR #3512, opened 2017-10-30
[Cape Town 2017] Program shuffle

adrianmoisey:
And turn off sponsors, since we're not going to sign up any more.
Also name the location of our event.


manicminer:
LGTM

PR #3516, opened 2017-10-30
[TLV-2017] updated speaker info for talk

yeshess:


yeshess:
sorry for the ugly PR

PR #3550, opened 2017-11-03
[Cape Town 2017] Schedule update

adrianmoisey:
Overlap and extend the workshops.

They were too short in the previous schedule.


manicminer:
LGTM!

PR #3554, opened 2017-11-03
[GAL-2017] Update organizers

electrofelix:


MikeRosTX:
Team email and Slack have been updated as well.

PR #3576, opened 2017-11-07
[RIGA-2017] Adding presentations

arthur-kovalenko:
Publishing presentations


bridgetkromhout:
Hi, @arthur-kovalenko - instead of using the embed code provided by slideshare, we recommend using a frontmatter element as shown in https://github.com/devopsdays/devopsdays-theme/blob/master/REFERENCE.md#talk-page-fields. Here's an example of it in use: https://github.com/devopsdays/devopsdays-web/blame/master/content/events/2017-minneapolis/program/eryn-oneil.md#L8

It would look like this: https://www.devopsdays.org/events/2017-minneapolis/program/eryn-oneil/ but more importantly, if you add it to the frontmatter, we can programmatically flag it in other ways (adding icons, etc).

PR #3612, opened 2017-11-15
[RIGA-2018] Setting up a page for DevOpsDaysRiga 2018

arthur-kovalenko:
Setting up a page for DevOpsDaysRiga 2018 / booking dates


data/events/2018-riga.yml
```diff
@@ -0,0 +1,83 @@
...
+event_twitter: "devopsdaysriga"
+friendly: "2018-riga"
+description: "DevOpsDays is coming back to Riga!"
+ga_tracking_id: UA-105774008-1
+
+#The image, relative to static/events/YYYY-CITY that you want to be the background of the header on your page.
+#masthead_background: ""
+
+# All dates are in unquoted 2017-MM-DD, like this: variable: 2016-01-05
+startdate:  2018-09-27
```
phrawzty: Do you have a signed contract with the venue for these dates?

data/events/2018-riga.yml
```diff
@@ -0,0 +1,83 @@
...
+# Location
+coordinates: "56.946695, 24.076940" # The coordinates of your city. Get Latitude and Longitude of a Point: http://itouchmap.com/latlong.html
+location: "Park Inn by Radisson Riga Valdemara" # Defaults to city, but you can make it the venue name.
+location_address: "Krogus iela 1, Riga LV-1048 Riga Latvia"
+
+# Team members
+team: ["Edgars Letinskis", "Uldis Karlovs-Karlovskis", "Laura Arajuma", "Arturs Kovalenko", "Liene Jansone"]
+
+# Emails
+organizer_email: "organizers-riga-2017@devopsdays.org" # Put your organizer email address here
```
phrawzty: This should be `-2018`, which doesn't exist yet. 😉 Could you please do one of the following:
1. Send an email to info@devopsdays.org with the names and email addresses of the organisers listed here, or
2. Confirm here that the desired membership of the list is the same as from 2017.
Then I'll set up the 2018 list.

data/events/2018-riga.yml
```diff
@@ -0,0 +1,83 @@
...
+proposal_email: "organizers-riga-2017@devopsdays.org"
```
phrawzty: (see above)

data/events/2018-riga.yml
arthur-kovalenko: Yep, does not exist, but for now it was more important for us to book dates :) 
I confirm that the list of desired members is the same as before.

data/events/2018-riga.yml
arthur-kovalenko: We have agreement with the venue. It's not signed, but we did the same last year.

data/events/2018-riga.yml
phrawzty: Super. I've made the `-2018` mailing list, so please update this PR.

data/events/2018-riga.yml
arthur-kovalenko: Changes made

PR #3623, opened 2017-11-19
[Brasilia-2017] debugging why it's not showing up

bridgetkromhout:
I noticed this shows locally but not on the built site - checking to see if it's this character in the URL.

### DIFF
--- a/content/blog/2017-Brasilia-inreview.md
+++ b/content/blog/2017-Brasilia-inreview.md
@@ -1,7 +1,7 @@
 +++
 Tags = ["2017"]
 Author = "Adriano Vieira"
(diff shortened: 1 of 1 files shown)


bridgetkromhout:
Hi, @mattstratton - I thought the problem here might lie in the special characters in the frontmatter, but that doesn't appear to be correct (or at least, it's not the entire problem). What's happening is that this blog post is showing up locally but not in the netlify builds (and therefore, not on the production site). I'm at a bit of a loss as to why. Adding verbose mode to the local build doesn't help show me any additional errors which might shed light on what's happening. Thoughts?

bridgetkromhout:
Note: I was running an old version of hugo locally.

PR #3624, opened 2017-11-19
[Stockholm 2018] Replacing twitter handle

bridgetkromhout:
Pretty sure @jessabe wants her public twitter profile to be the one linked. Let me know if this is wrong, @lsc!


lsc:
Yeah, that's my bad. I picked first in my search over at tweetbot

--
Lowe Schmidt | +46 723 867 157

PR #3626, opened 2017-11-20
[São Paulo 2018] Moving SP to 2018

bridgetkromhout:
Hi @gomex - I'm moving São Paulo event from 2017 to 2018, since as far as I can tell it didn't happen in 2018. Please let us know if the organizer list needs to change - thanks!


gomex:
Thanks!

PR #3634, opened 2017-11-21
[Philly 2018] Adding landing page for 2018.

peterjshan:
On the tin.


MikeRosTX:
@peterjshan before we can merge this PR since you're specifying a date, you need to add a **Location** nav_element to yaml if you have a confirmed venue.

PR #3637, opened 2017-11-28
[NYC-2018] Improve sponsor colors and text

TomOnTime:


MikeRosTX:
Since there are no **Pay Now** buttons for Gold, I'm assuming it's **SOLD OUT**.
If it indeed **Sold Out**, you should mention it on the **Sponsor.md**, if not then I would recommend having a **Pay Now** button.

PR #3641, opened 2017-11-30
[Atlanta 2018] Correction

bridgetkromhout:
The frontmatter wasn't based on the template we're using now (and future dates will make things not show up).


bridgetkromhout:
When I dug into the discrepancies in this event, I noticed a few issues - we have "speaker" vs "program" mixed up for one speaker, and we have a title in the name of the other speaker (causing a consistency issue). After corrections, both speakers show up correctly. /cc @cacorriere

PR #3643, opened 2017-12-01
Update hugo version in docker to v0.30.2

mattstratton:
This change updates the version of hugo in the docker script to v0.30.2


mattstratton:
I tested this locally, and it works!

PR #3655, opened 2017-12-05
add_program.sh should be more flexible in parsing yml

TomOnTime:
add_program.sh was breaking on Darwin. This fixes the problem.


devopsdays-service:

adrianmoisey:
Some context on this:
https://github.com/devopsdays/devopsdays-web/blob/e28803da9f2746918e8adcc3796e0298e3e7bb86/data/events/2018-new-york-city.yml#L9

There are 2 spaces between `startdate:` and `"2018-01-18"`. Since the `cut` uses spaces as a delimeter, it's grabbing the second space, instead of the date itself.

PR #3659, opened 2017-12-07
[PDX-2018] We have a venue!

BillWeiss:
Also, we added an organizer, and I lost an employer.


MikeRosTX:
Waiting for email on organizers changes before merging this PR. @BillWeiss is aware.

PR #3673, opened 2017-12-11
add_program.sh: Improve MacOS compatibility

TomOnTime:


devopsdays-service:

PR #3685, opened 2017-12-12
utilities/add_* should read defaults from env

TomOnTime:
* Instead of always prompting for year and city, check if $DOD_YEAR or $DOD_CITY is set.
* Future-proof the default year setting


devopsdays-service:

PR #3697, opened 2017-12-14
[VAN-2018] Updating JPEG with white background

tylermauthe:
The previous logo-square image was a PNG and on iOS showed up with a black background - I'd like to replace with this JPEG. Thanks!


phrawzty:
@mattstratton I don't think we're against this (using jpgs for logos), but if transparent PNG backgrounds are showing up black on iOS, that might be something we should look into. 😄

mattstratton:
IIRC the docs specify a jpg for logo-square in the first place.

tylermauthe:
I'm curious why we are required to use JPEG?

mattstratton:
Because that’s how i coded it. There is logic that looks for a specific file name. If this is a major show stopper for you we can prioritize it into the backlog. But jpg is a better format for most logos from a performance perspective

mattstratton:
Anyway, this is all documented. I spent a fair amount of time writing the docs which are all linked in the readme. Please read them rather than wonder why it works or doesn’t. Thanks!

PR #3709, opened 2017-12-16
[ZRH-2018] add nine.ch as gold sponsor + new logo for Atlassian

drbokko:


MikeRosTX:
Recommend using jpg instead of png format. P
Why is there a new logo for Atlassian being upload if you don't even have them listed a a sponsor in the yaml file?

drbokko:
Hello Mike:
- Atlassian requested a logo change for the 2017 event; We are finalising the contract so I thought it would be nice to perform the change as they request.
There is a unique logo file per sponsor unless one would create sponsor profiles with the year in the prefix (as was in the past)

- nine.ch also requested a logo change

- About the png I use the same format as the former logo I found.

bridgetkromhout:
I'm preserving the logo history for old events in https://github.com/devopsdays/devopsdays-web/pull/3713 - when that's in place I'll merge this.

MikeRosTX:
OK, it's all fine.  But Atlassian's logo is not appearing in the preview.
**https://deploy-preview-3709--devopsdays-web.netlify.com/events/2018-zurich/welcome/**

Only reason I suggest changing png to jpg, is because we've seen this behavior before.

bridgetkromhout:
>> We are finalising the contract 

>  Atlassian's logo is not appearing in the preview.

This would imply to me that they'll list the logo when the contract is final.

bridgetkromhout:
Hi, @drbokko - if you rebase off master to resolve the conflict this should be okay to merge now.

drbokko:
@MikeRosTX @bridgetkromhout pages now look as they should

bridgetkromhout:
Well, the new atlassian file has the wrong name, but i'll just merge and then fix. thanks.

PR #3713, opened 2017-12-17
Update sponsors

bridgetkromhout:
Sponsor logo changes (and preserving history).


drbokko:
@bridgetkromhout should I do the same for all sponsors with updated logo?

PR #3729, opened 2017-12-21
[Cape Town 2018] Add Cape Town event for 2018

adrianmoisey:
Placeholder page for Cape Town's 2018 event.

What do I need to do to get `organizers-cape-town-2018@devopsdays.org` and `proposals-cape-town-2018@devopsdays.org` created?


MikeRosTX:
Emails created.

PR #3773, opened 2018-01-04
Columbus 2018: Base creation (from script), reserving dates.

kjenkins19:
This is just the kick off for the Columbus 2018 to reserver the dates, 9/19 to 9/20.


bridgetkromhout:
Removed Andy Dyrcz from Slack and email, based on not being in this PR.

PR #3781, opened 2018-01-04
[HFD-2018] Adds TBD page and current organizers for Hartford 2018

crayzeigh:
Initial page updates for Hartford 2018 with unconfirmed dates and confirmed organizers.


bridgetkromhout:
Do the organizers with only one name listed want to be listed that way? I ask because usually people want public credit for the work they're doing.

crayzeigh:
This is an excellent question and one I planned on asking, as well. 

My initial thought was to publish as is so all organizers could see what the info provided currently looked like online and provide updates to the information they'd prefer to list.

If you think it better to ask for updates before our initial PR I have no problem with that either.

PR #3830, opened 2018-01-16
[CHINA-2018] add 3 China events in 2018

martinliu:
We add Beijing, Shanghai and Shenzhen and all oragranrizer. This is a initial PR for the year, will add more detail later.


martinliu:
I will make a new PR here. Close it for now.

PR #3831, opened 2018-01-16
[BEIJING-2018] add Beijing 2018 event

martinliu:
We added Beijing event information, will continue to add more detail later.


bridgetkromhout:
Looks good! Thanks.

PR #3834, opened 2018-01-16
[IST-2018] Add 2018-istanbul event dates

srhtcn:


bridgetkromhout:
In order to list dates, we require you to also list the location (as you need to have confirmed those dates with your venue). Thanks!

srhtcn:
Thanks @bridgetkromhout , I've also added the location info :)

PR #3835, opened 2018-01-16
[SP-2018] Fixed mobile view

fike:
Rollback event filenames to original in "content/events/2018-sao-paulo" because broke mobile icons when use filenames in portuguese.


bridgetkromhout:
Hi, @fike - sorry for the annoying problem. I agree that mobile icons should work correctly! Fortunately, we have another option that doesn't require you to revert to English. I've fixed this problem with https://github.com/devopsdays/devopsdays-web/pull/3836 - now you should see that the mobile icons are fixed.

We have an open issue for internationalization (https://github.com/devopsdays/devopsdays-theme/issues/47) and clearly we need to build this in (so that if your site is using the expected words in any supported language, you don't need to explicitly name the icons).

PR #3838, opened 2018-01-17
[VAN-2018] new sponsors added to van2018 and updating treasuredata sponsor image

rmkbow:
adding realtor.com sponsor
updating treasuredata logo image (less wasted horizontal whitespace)
adding sponsors to vancouver 2018


kmugrage:
@rmkbow - we avoid changing logos for events which have happened in the past, which this PR would do. 

Please see https://github.com/devopsdays/devopsdays-web/pull/503 and https://github.com/devopsdays/devopsdays-web/blob/master/utilities/README.md for context.

rmkbow:
@kmugrage doesn't that only apply to rebranding? this one is like fixing an image that has been needsmorejpg?

if it applies should I rename the old one and keep the new one with the base name, and update old references to use the updated old sponsor name?

or would best course of action to just create copies sponsor data/images of existing sponsors and not reuse existing sponsor data/images them at all then so there isn't this historical dependency? like each event keeps it's own 2018-Vancouver-sponsorname data/image

bridgetkromhout:
Ah yes, this update of the image doesn’t change it visually. So it’s fine in this case. We want to keep a sponsor not event or year specific - there is often re-use.

PR #3878, opened 2018-01-25
[bmore-2018] Adding a number of sponsors; updating parking cost

bushong1:


bridgetkromhout:
@bushong1 - we put our podcast in as a default because we talk about devopsdays every episode, but it's in no way required to list us or make a discount code available. Totally optional. :)

PR #3911, opened 2018-01-31
[SEA-2018] upgrade sumologic to gold

kmugrage:
Also updated JFrog logo


kmugrage:
Not merging myself since it's a logo change to existing events. Seems to me like it's "close enough" but would like someone else to merge to make sure.

bridgetkromhout:
Yeah, it's not dramatically different.

PR #3918, opened 2018-02-02
[ZRH-2018] Add Redgate as gold sponsor

Roderick-Jonsson:
Add Redgate as gold sponsor


Roderick-Jonsson:
How do I remove the "become a sponsor" links on the welcome page? We just need bronze
screenshot_17

mattstratton:
You need to set the max level for the ones you don't want to keep to the amount you already have; I think I got the numbers right but check it in the deploy preview once it's done building :)

PR #3922, opened 2018-02-03
[CLT-2018] adding sponsors

jmaggio:


bridgetkromhout:
Hi, @jmaggio - 

The logo you're trying to add for Dell EMC violates their brand guidelines by changing the color of "EMC": http://brand.delltechnologies.com/dell-emc/logo-library/

And the logo you're trying to add for New Relic is one they use on Twitter, but it's not the one they've been using with devopsdays events for years. 

In both cases, do you have an email from the sponsor saying they want that to be what we use?

We're careful about such choices because it can lead to confusion for other events.

Thanks.

bridgetkromhout:
I should learn never to use github's web-based editor because it just leads to messy merge conflicts! 

@jmaggio, I've got all your sponsors listed on your event page now, with their last-used logo renamed to be the most current. You'll want to `git pull upstream master` in particular to sync the `data/events/2018-charlotte.yml` file from master, as I've re-named a number of sponsor files. We try not to give sponsors a year-specific name, because then many events won't think to look for it and will end up using the old one.

I didn't merge in your two proposed new logos (for Dell EMC and for New Relic) because of the reasons I stated earlier. Please send separate PRs for those changes if you have a communication from the vendor asking for those edits.

Thanks.

jmaggio:
Ah thank you @bridgetkromhout

PR #3923, opened 2018-02-03
[SP-2018] Add organizers details and updated index page

wsilva:
- Added details of each organizer
- Updated index page with some more informations.


bridgetkromhout:
Hi, @wsilva - if you are going to add and remove organizers, please email us at info@devopsdays.org. We need the email address of the new organizer.

wsilva:
Hi @bridgetkromhout, I have just made a quick fix and we are adding Dani Vieira to organizers. Also I have just sent the e-mail as requested.

PR #3934, opened 2018-02-05
[DSM-2018] Remove silver sponsorship

timgifford:
We don't have any more space for sponsorships. Don't want anymore sponsorships to sneak in.


mattstratton:
@timgifford You probably want to set the max for silver sponsors in your data file to the number that you have...since the "become a silver sponsor!" link is still active on all your pages.

timgifford:
Thanks for the tip @mattstratton!

PR #3947, opened 2018-02-08
[DEN-2018] Add Contrast Security sponsorship.

mattk42:
Also removing DynaTrace for now as their payment has not come through and removing the link to paypal for platinum sponsorship.


bridgetkromhout:
Yeah, in Minneapolis we tell people "all benefits of sponsorship start after payment is received".

PR #3995, opened 2018-02-19
Initial commit for Boise 2018

tresni:


data/events/2018-boise.yml
```diff
@@ -41,29 +42,31 @@ team_members: # Name is the only required field for team members.
...
-sponsors_accepted : "no" # Whether you want "Become a XXX Sponsor!" link
+sponsors_accepted : "yes" # Whether you want "Become a XXX Sponsor!" link
 
 # In this section, list the level of sponsorships and the label to use.
 # You may optionally include a "max" attribute to limit the number of sponsors per level. For
 # unlimitted sponsors, omit the max attribute or set it to 0. If you want to prevent all
 # sponsorship for a specific level, it is best to remove the level.
 sponsor_levels:
+  - id: presenting
+    label: Presenting
```
bridgetkromhout: Hi, folks - while we encourage you to be creative and individual about what you call your sponsorships, "Presenting" will confuse sponsors because it sounds like it includes them giving a conference presentation. Selling talk slots is not allowed. Perhaps "Exhibiting" is more along the lines of what you mean by this?

bridgetkromhout: I read through your prospectus, and I think I see what you mean by "presented by". I'd still suggest changing your "Presenting" level to "Platinum" to avoid sponsor confusion. You also might want to change your "Booth" level to "Bronze", as "Booth" will confuse people who may think that's the only choice if they want a booth (and they additionally may try to ship their booth components, when you only have room for an 8-ft display table).

tresni: Updated in b5ee30f.  Removed prospectus (and reference to) until it gets rev'd.

bridgetkromhout:
Hi @tresni! Thanks, and sorry for the delay - I was at an event and I didn't make it clear that you could have this merged with these changes. I'm merging now.

PR #4025, opened 2018-02-23
[CPH-2018] corrected-new-sponsors-added

mariuszbury:
Hey. I applied _rebase_ and fixed all the issues - everything should bang now. The other "don't merge" branch can be deleted.


mattstratton:
Why was the theme added as a submidule?

PR #4027, opened 2018-02-23
[DET-2018] Initial PR for Detroit 2018

joearnet:
PR for Detroit's 2018 event. Our prospectus is pending as well as a few more commitee members. Otherwise all is as it should be, as far as I know


joearnet:
Asked by Mike Rosado to list the organizers from DET-2017 whom are leaving; here they are:
Ben Chess
Brian Nuzskowski
Zach Zeid

PR #4031, opened 2018-02-24
Seattle speakers

bridgetkromhout:
Submitting for @kmugrage's review. I'll mark Do Not Merge and he can decide if he wants this.

- Added "Speakers" page
 -- Note that some images are very small but they'll be enlarged in the production build
 -- Note that I removed "image" lines that are blank so the default image will show up, and added a missing one so Aaron's picture shows up.

- Fixed a couple speaker images that aren't square - the build will make them square and it might not look great, like cropping their heads:

screen shot 2018-02-24 at 1 18 33 pm

Also, for discussion:

1) I'm curious if you flagged those afternoon talks as "workshop" simply because you wanted a different color? or are they actually workshops? 

2) Your front page still says "Our Call for Proposals is now closed. We are busy going through more than 200 great submissions and will announce the schedule soon!") - even if that refers to how you're still selecting Ignites, you might want to clarify.


kmugrage:
The afternoon talks are “tech talks”, but not actually workshops. I’m planning on finding a way to add clarity about that but didn’t want to delay the posting more.

PR #4033, opened 2018-02-24
[Victoria 2018] Adding Victoria logo to front page

bridgetkromhout:
Adding the same Victoria logo from the event page so it's also on the front page. /cc @jrstarke


jrstarke:
LGTM

PR #4072, opened 2018-03-01
[OSL-2018] Creating Oslo 2018 event

bruvik:
creating page for Oslo 2018, with the organizers that so far have confirmed they will join this year


bridgetkromhout:
Welcome back!

PR #4092, opened 2018-03-05
Blacks in Technology already existed as a sponsor

bridgetkromhout:
Hi, @ghoneycutt - Blacks in Technology already existed as a sponsor, so I'm fixing it so we don't have a second sponsor that's essentially the same. See https://github.com/devopsdays/devopsdays-web/pull/4076 for reference.


ghoneycutt:
@bridgetkromhout 

We had a higher res graphic than what is currently there. Are the logos ever shown where this would matter?

ghoneycutt:
Looks like the text is smaller.. I'll submit a new PR.

PR #4132, opened 2018-03-11
[Beijing-2018]Open registration.

martinliu:
Add registration link.


bridgetkromhout:
"registration_closed: 2018-05-04" this doesn't take a date - just true/false reflecting current state, so I'd suggest not setting it.

martinliu:
I will change that, but I did not see any error when I build locally. Thanks for your help!

martinliu:
Ok, I got it fixed.

bridgetkromhout:
I think you are right; we don't validate all the yaml. I just didn't want you to think that would do something that it will not.

PR #4155, opened 2018-03-13
[Maringa-2018] Changing title and description of my talk

b-dantas:


mattstratton:
@b-dantas Are you one of the organizers of the Maringa event? If not, we will need approval from an organizer.

b-dantas:
@mattstratton it was requested by organizers to change it. They are too busy to do that because the event is close. @souzaxx could you confirm this with linuxmen?

MikeRosTX:
Fixed in #4160

PR #4173, opened 2018-03-15
[CHI-2018] Add Bluth company as sponsor

mattstratton:
Added the Bluth company as a sponsor for Chicago - it's also a brand new sponsor in the system.


mattstratton:
Whoops testing mistake.

PR #4206, opened 2018-03-20
[TOR-2018] Speaker updates

WhatsChrisDoing:


kmugrage:
@WhatsChrisDoing - I noticed when reviewing this that you have the @ symbol in front of the twitter handle for Kris. That is not required. Also, when testing to see if it works anyway (it does) I noticed that she asks that people not use the account you linked. 

I suggest changing "Twitter = "@kris__nova"" to "Twitter = "krisnova""

WhatsChrisDoing:
Good catch. Let me get a PR in.

PR #4218, opened 2018-03-22
[DET-2018] Adding new organizers

joearnet:
[DET-2018] Adding new organizers


MikeRosTX:
Please send email to **info@devopsdays.org** with list of new organizers to be added.

**full name - email**

PR #4226, opened 2018-03-23
[JKT-2018] Update sponsor logo

streichsbaer:
This PR updates the Progrez sponsor logo


bridgetkromhout:
Note: I'm updating this without preserving the old one because there are no past events where it's used.

PR #4229, opened 2018-03-23
[HFD-2018] Set the date and venue for Hartford 2018.

jaxzin:


MikeRosTX:
Need confirmation of venue location.

PR #4247, opened 2018-03-26
[LON-2018] Adding more organisers

kramos:
Adding more organisers.  Just one other person still TBC so I haven't added them yet.


kramos:
Eeek, very sorry @mattstratton and thanks!  I'd expected hugo server to fail locally if there was an error.  I'll always visually check from now on.

mattstratton:
Huh. It should be failing locally for you too! Weird. 

The problem is it returns a non-zero edit code but doesn’t crap out. So you don’t see it died unless you look at the Hugo output lines carefully

PR #4267, opened 2018-03-29
Devopsdays Warsaw 2018 add location , data, organizators

APiwowarczyk:
i add location , change data & organizators


MikeRosTX:
COMPLETED:
organizers-warsaw-2108@devopsdays.org was created.

ACTIONS NEEDED:
Need to troubleshoot some more, because it is not previewing via deploy/netlify.
Slack invitations need to be matched up and sent out

PR #4290, opened 2018-04-01
Changing Denver from jpeg to jpg

bridgetkromhout:
Changing Denver from jpeg to jpg. Also increasing the size of a couple that were below the 300x300 minimum.

This is intended to fix this:

screen shot 2018-04-01 at 3 03 00 pm


bridgetkromhout:
Heads up to @tavisto - you'll want to `git pull upstream master` or some such, to get these changes into your branch.

PR #4292, opened 2018-04-01
[Van 2018] Vancouver's CFP is closed

bridgetkromhout:
Hi, @tyler-mauthe-hs - I suggest you 👍 this PR because your CFP is closed, and this makes that clear. Please let us know - thanks.


tylermauthe:
Good call - thank you for the update.

PR #4295, opened 2018-04-01
[Den 2018] Removing propose link

bridgetkromhout:
Since the Denver CFP is closed, you probably don't want this link anymore. Looking for a 👍 from @tavisto or other Denver organizer.


tavisto:
Looks good to me

PR #4321, opened 2018-04-04
[Des Moines 2018] Updating appdynamics logo

bridgetkromhout:
Hi @jessiepuls - I noticed that AppDynamics has been asking other events to use their new logo. You probably want to 👍this (unless you've made other arrangements with them). Thanks.


jessiepuls:
Looks good to me 👍

PR #4323, opened 2018-04-04
Validate shell scripts

ghoneycutt:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,4 +17,5 @@ before_script:
 
 script:
   - ls
+  - for i in $(find . -type f -name "*.sh"); do bash -n $i; done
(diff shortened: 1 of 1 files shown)


bridgetkromhout:
Since this affects the build, I'm going to defer a decision on it to @mattstratton (as he's the one with the most insight into the build).

bridgetkromhout:
> Started monkeying with shell scripts and though we should add some simple validation.

I don't disagree. I want to make sure this doesn't have unanticipated effects, and I know Matt will have more insight into it than I will.

bridgetkromhout:
Upon further consideration, this sort of syntax validation is unlikely to have side effects.

PR #4324, opened 2018-04-04
Add ability to specify twitter handle when adding sponsors

ghoneycutt:

### DIFF
--- a/utilities/add_sponsors.sh
+++ b/utilities/add_sponsors.sh
@@ -55,6 +55,9 @@ fi
 read -p "Enter sponsor url: " url
 [ -z "${url}" ] && url=''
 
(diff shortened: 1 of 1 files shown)


devopsdays-service:
@ghoneycutt, thanks! Based upon the content in this PR, we need a review from @mattstratton and @bridgetkromhout

bridgetkromhout:
Good addition. Though we aren't displaying these programmatically yet, it doesn't hurt to collect them.

bridgetkromhout:
And just for reference, here's some clarity on where we've gone back and forth with this in the past: https://github.com/devopsdays/devopsdays-web/pull/3326

PR #4328, opened 2018-04-05
Fix extra space in yaml

ghoneycutt:
Without this patch the add_sponsors.sh script will add an extra space to
the yaml output which can cause the sponsor logo to not appear.

### DIFF
--- a/utilities/add_sponsors.sh
+++ b/utilities/add_sponsors.sh
@@ -74,7 +74,7 @@ else
 fi
 
 echo "Add this to ../data/events/"$event_slug".yml under sponsors:"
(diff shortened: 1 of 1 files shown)


devopsdays-service:
@ghoneycutt, thanks! Based upon the content in this PR, we need a review from @mattstratton and @bridgetkromhout

ghoneycutt:
After the patch there is only one space between the `id` key and its value.

```
Add this to ../data/events/2018-indianapolis.yml under sponsors:
  - id: foo
    level: theirlevel
```

bridgetkromhout:
Yep, checks out.

PR #4363, opened 2018-04-10
[ZRH-2018] Add color legend for program

Roderick-Jonsson:
We had a request regarding what the colours mean on the program site, so I implemented a color legend for clarification.


Roderick-Jonsson:
@MikeRosTX Awesome! Thank you Mike for beeing o fast 👍

Issue #4127, opened 2018-03-09
Change to use common docker image


mattstratton:
There is an updated docker setup that we won't have to maintain here - https://hub.docker.com/r/jojomi/hugo/

We should update our docs and scripts (and travis build) to use this instead.

espoelstra:
This should probably include updating the CONTRIBUTING.md to recommend Docker instead of installing Hugo directly.

mattstratton:
I don’t recommend that. It’s an option but not the recommended solution. We provide it for people who want it, but frankly I don’t support it myself.

PR #4401, opened 2018-04-14
Use existing docker image

adrianmoisey:
Fixes https://github.com/devopsdays/devopsdays-web/issues/4127

I still need to figure out how Travis is setup and replace that one.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ services:
   - docker
 
 before_install:
-  - docker run --name "hugo-server" -P -v $(pwd):/src jojomi/hugo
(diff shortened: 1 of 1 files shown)


kmugrage:
"works on my machine"

mattstratton:
My only concern is that the baseURL doesn't get set...if you build this, what are the paths to the CSS, etc, in the html?

bridgetkromhout:
Hmm, Travis doesn't seem to like this: `The command "docker run --name "hugo-server" -P -v $(pwd):/site jojomi/hugo" failed and exited with 1 during .`

adrianmoisey:
Replaced by https://github.com/devopsdays/devopsdays-web/pull/4536

PR #4428, opened 2018-04-17
[NYC-2019] Adding Team Members

russAIC:
adding committee team members, not sure why this is thinking that we're adding 8 new files :/


MikeRosTX:
@russAIC - Before we can merge this PR, please send an email to **_info@devopsdays.org_** with Khader & Jerome emails so we can add them to team's email and to Slack.

PR #4603, opened 2018-05-16
[DC-2018] Update USDS logo per their request

markcornick:


MikeRosTX:
Marked **_do-not-merge_** until we get confirmation.
In the title it should as **_[cityabbr-yyyy]_** so that it's not so confusing.
All indications from email, is that this new logo should have been for DC but it's not reflected as a Platinum sponsor in the preview. Upon further review of the code, it looks like the **_yaml_** modified was for **_Baltimore_** and not **_DC_**.

markcornick:
USDS is a platinum sponsor for DC, see https://github.com/devopsdays/devopsdays-web/blob/master/data/events/2018-washington-dc.yml#L263 (this did not require modification in this PR)

When previewing locally I see the new USDS logo on the DC 2018 page and the old logo on the Baltimore 2018 page. What else do you need here?

MikeRosTX:
Thanks for the confirmation!  It's clear now.

PR #4630, opened 2018-05-22
[HFD-2108] Update CFP and registration

jpadilla:


MikeRosTX:
Labeling **_do-not-merge_** until all checks pass.

PR #4670, opened 2018-05-30
[SP-2018] Added new speaker and sponsor

fike:
* Added Igor Souza as speaker
* Added Vmware as sponsor


fike:
@MikeRosTX, I have been updated Igor image. Is it fine, now?

PR #4715, opened 2018-06-05
[SP-2018] Last  add ignites

fike:
* Added Alexandre Stratikopoulos as Ignite Speaker
* Added BrownBagBrasil as Community Sponsor
* Moved Igor Souza talk to 07/06
* Move Launa Linares talk to "normal" talk


data/events/2018-sao-paulo.yml
```diff
@@ -167,6 +167,8 @@ sponsors:
     level: community
   - id: easytaxi
     level: community
+  - id: vbrown  bagbrasil
```
kmugrage: This doesn't look like a valid sponsor id. Should be vbrownbagbrasil based on other info in this PR

PR #4720, opened 2018-06-06
[INDY-2018] Change sponsor levels - Platinum includes after party

ghoneycutt:
Platinum sponsorship includes after party sponsorship from a la carte
section.


ghoneycutt:
netlify timed out. I think the job needs to be restarted.

PR #4728, opened 2018-06-08
[IST-2018] update Istanbul sponsor snapbytes logo as requested

serhatcan:
update sponsor logo as requested


bridgetkromhout:
This is the first time this is used, so no history needs to be preserved.

PR #4733, opened 2018-06-09
[MSP 2018] Adding Stripe as break sponsor

bridgetkromhout:
Adding Stripe as break sponsor


bridgetkromhout:
Merging anyhow since it was just a netlify timeout.

PR #4781, opened 2018-06-19
[ATX-2019] Adding skeleton site with dates for DoDATX 2019

richardboydii:
We are approved for UT Stadium again for 2019 and the core organizers (Ernest, Karthik, and James) are all read in and approved.

Created a base event config and site using the add_new_event.sh.


kmugrage:
@richardboydii are there any changes to organizers this year? Need to create the mailing lists and verify slack access. 

If there are changes, please email info@devopsdays.org with a complete list of full names and email addresses for the organizing team. 

Thanks!

richardboydii:
@kmugrage looks like I stepped on your toes but the changes you made are the same I was going to make.

For now I'd like to keep the organizer list for 2019 the same as 2018. We'll update it sometime in the Fall / Winter but I want to keep everyone in the loop as we move towards the end of Q3.

Should I email the core team to get the new addresses for 2019 made? What's the process there? Same as creating a new conference?

kmugrage:
@richardboydii please do email info@devopsdays.org with the names / emails for this year's crew. I was going to just copy over 2018 but there are more than 20 people on that one and only 14 on this year's contact page (so far). We need to make sure they match.

Thanks

PR #4822, opened 2018-06-27
[CHA-2018] Adding logo, removing Ethan as organizer, adding sponsorship tier

robscott:
Unfortunately Ethan has stepped down as an event organizer due to a busy schedule.


phrawzty:
I've also removed Ethan from the mailing list and Slack.

PR #4828, opened 2018-06-28
Rio de Janeiro 2018 - adding initial content & organizers

amandabatista:
Rio de Janeiro 2018 - adding initial content & organizers


phrawzty:
Mailing list created; Slack invites sent.

PR #4841, opened 2018-06-29
[NYC-2019] adding more organizers 

russAIC:
*Please title your pull request in this format: The event name and year in the title of the PR, along with a description of what is being changed, i.e., `[CHI-2018] Add Bluth Company as a sponsor`*

PR #4842, opened 2018-06-29
looks like my name was missing from the org list?

ceren:
Ceren wasn't on the organizers list


MikeRosTX:
Closing out because I'm adding to #4841

PR #4882, opened 2018-07-06
[LDN-2018] Add speakers and program

chriscoffee:
I've added all chosen speakers who've confirmed that we've selected. This includes both 30 minute talks and ignite talks but does not include backups which are currently on a seperate branch should we need it.

---

[LDN-2018] Remove speakers who aren't speaking

I messed up slightly when adding speakers after being confused by the
spreadsheet. This commit removed those speakers who aren't actually
speaking...

[LDN-2018] Remove .jpg from link title

I'd managed to add .jpg to a link title reference for Louise.


phrawzty:
Looks like a _great_ line-up so far. Fantastic!

PR #4886, opened 2018-07-06
[INDY-2018] Added Jared Speno, removed Robert Fox and Ian Alford

ghoneycutt:


bridgetkromhout:
Did you intend to remove the `Cats In A Thunderstorm` talk? Thanks.

ghoneycutt:
@bridgetkromhout netlify died, or we could double check. The talk is still there under Ian Alford. Found that when you list the program twice, one for each speaker, and click on a speaker, you see double program entries. This is meant to fix that.

ghoneycutt:
Thanks for checking!!

PR #4932, opened 2018-07-15
[JKT-2018] Delete a speaker image for privacy concern

mappuji:
As the owner of the image, I want to delete the image for privacy concern and also delete the twitter handle since I am not using the Twitter account months ago and it already used by another person.


mappuji:
Thanks @MikeRosTX

PR #4970, opened 2018-07-24
[SIN-18] Updating organizer info

muthupl:
Updated organizer details with my name


MikeRosTX:
Whenever a new organizer is added to the website, an email needs to be sent to **_info@devopsdays.org_** indicating a new organizer is coming onboard with the new organizer's name and email so we can add that person to Slack and team email.

PR #4985, opened 2018-07-26
[Riga-2018] 4 new speakers added

ayartsev:


phrawzty:
Confirmation received from Arturs Kovalenko via Slack.

PR #5004, opened 2018-07-29
[BER-2018] Preliminary Program

range:
This is the preliminary program for the 2018 Berlin event.


MikeRosTX:
Are you sure you want to use the old style program and not the new template that has been created to generate a program page? An example can be seen at https://dev.devopsdays.org/events/2017-ponyville/program

MikeRosTX:
Ralph Angenendt [4:48 PM]
Yes. And I don't seem to have ponyville in my tree?

I'd like to get the program out as fast as possible and then maybe look at the new style during the next week. There will be changes anyway.

Mike Rosado [4:49 PM]
OK, merging AS IS. :+1:

PR #5072, opened 2018-08-08
[CAI-2018] Adding InfoGlobe as gold sponsor

aameen79:
[CAI-2018] Adding InfoGlobe as gold sponsor

PR #5076, opened 2018-08-08
[CAI-2018] move infoglobe logo to correct location

phrawzty:
#5072 was merged but the sponsor logo was not in the correct location; this PR fixes that issue.


phrawzty:
@aameen79 

image

PR #5131, opened 2018-08-16
[JHB-2019] Added additional organiser

bdelate:
Added an additional organiser.


phrawzty:
Email received as well; I'll modify the ML and send out the Slack invite shortly.

PR #5150, opened 2018-08-18
[Raleigh 2018] Fixing links for Raleigh

bridgetkromhout:
Hi, @normalfaults - there are a couple places in the welcome template that aren't driven by the yaml. I've fixed these broken links for you, and made your location map show up correctly.


normalfaults:
thank you!

PR #5154, opened 2018-08-18
[Berlin 2018] Adding speaker updates (pics, bios)

bridgetkromhout:
I noticed that there was some missing info for your speaker page. Here's my suggested additions/changes. We need someone from the Berlin team such as @range or @nmesstorff to approve this update.


bridgetkromhout:
Great!

PR #5156, opened 2018-08-18
[London 2018] Image resize; removing proposal link

bridgetkromhout:
Hi, @chriscoffee - I suggest removing the proposal link since the CFP is closed, and moving those two pics to jpg because jpeg extensions don't get resized correctly. Please let us know if this is okay for the London site. Thanks.


chriscoffee:
Thanks for the commit and the suggestion @bridgetkromhout. I didn't actually know there was a 'jpeg' image on there haha, slightly fail but great spot.

I will let others know on our call tomorrow too.

PR #5159, opened 2018-08-19
[BHE-2018] New organizers and others

devopsdaysbh:
Changelog:
- New organizers
- Removal of 2 former organizers
- CFP is now open!
- Event and event venue!


devopsdaysbh:
Note: E-mail with some requests for new and old members sent

PR #5186, opened 2018-08-21
[CU-2018] Move 2018 event to 2019-Cuba TBD

kikicarbonell:
We need to decline the event 2018-Cuba to start planning next event in 2019 Cuba with venues: Santa Clara, Villa Clara, Cuba.


kikicarbonell:
Thanks @MikeRosTX

PR #5210, opened 2018-08-24
VictorOps logo look update

bridgetkromhout:
This updates a number of events in order to keep the past orientation of the logo on those pages. Boston will need to pull in this update for their yaml.

Note to @technovangelist and @MikeRosTX - we don't add a year-delimited version of logos anymore. If a company asks to have their logo updated, we datestamp the old logo and events using it.


MikeRosTX:
Duly noted.

PR #5212, opened 2018-08-24
Attempt to make it clearer where the reference documentation lives

adrianmoisey:
This seems more clear to me, but I'm not 100% sure if it really is.


adrianmoisey:
It might make sense to move "Reporting issues" down to above or below "Contributing".
But I think I won't do that just yet.

PR #5213, opened 2018-08-25
adds TZ info to all CFP start/end/announce dates

tgross:
For https://github.com/devopsdays/devopsdays-theme/issues/476 and https://github.com/devopsdays/devopsdays-theme/issues/487.

Data changes to take advantage of the template updates in https://github.com/devopsdays/devopsdays-theme/pull/641


tgross:
Currently a merge conflict b/c I need to update a few dates that got changed since I started the PR. Fix coming shortly.

tgross:
Merged conflict fixed.

tgross:
cc @bridgetkromhout 

I have another half-finished commit waiting in the wings to add this data to start/end dates and registration dates, but I want to verify that we really want to do that for those before I wrap it up.

bridgetkromhout:
> I want to verify that we really want to do that for those

@tgross, thanks! indeed, I'm not sure about that.

I do have a possible-confusing "what even is timezones" question, though - specifically, when an event says their CFP "closes" on a specific "day", I think the common interpretation is "23:59:59 in that timezone". That is to say, if a CFP closes on 2018-01-01, you have all day long on the 1st of January to get your submission in. If I am understanding time correctly (not entirely a given!), I think what we're seeing here in this very welcome and helpful PR is "close the CFP the _instant_ it becomes January 1st!" which is _probably_ not a desirable default, as it may confuse the organizers who are thinking of the day as being "end of day" and not noticing the change. Thoughts?

tgross:
Yeah from a user experience that's definitely something I hadn't considered. We could change the end times to be 23:59?

Also, how does that actually work given this is a static site? Is it getting periodically re-rendered on a timer?

tgross:
> Let's assume that anyone who wants their CFP to close at 7pm or whatever would change it from the default, but I think that in the absence of clarifying details, people proposing talks would assume 23:59 conference-location-local time.

One quick `sed` later and that's done.

bridgetkromhout:
> I want to verify that we really want to do that for those

I think we came down at "yes" for this (when we do the next theme release).

Looking at this work-in-progress I see some cities missing from your "let's update all the 2018/2019 events"? You can see this if you check out the `/speaking` page on a local build of this branch or on the netlify preview: https://deploy-preview-5213--devopsdays-web.netlify.com/speaking/

screen shot 2018-08-27 at 2 57 55 pm

So, before we merge this I think we may want to make sure they're all up to date (I realize this is a moving target!)

In terms of timing for merging this, we also may want to wait until after the code release that supports this, because the current codebase just shows the entire string which will look a little confusing.

tgross:
> So, before we merge this I think we may want to make sure they're all up to date (I realize this is a moving target!)
> In terms of timing for merging this, we also may want to wait until after the code release that supports this, because the current codebase just shows the entire string which will look a little confusing.

Yeah, just to avoid a bunch of rework I'll hold off on making that data fix until the theme update gets deployed. But once that's been done I'll knock that out.

mattstratton:
Just want to confirm that we can roll out the theme changes you've done so far without this PR?

tgross:
Yes. :+1:

PR #5249, opened 2018-08-28
[YUL-2018] Add schedule

purpleidea:
This is our schedule and timetable!


purpleidea:
I think I understood the format from looking at other entries. Hopefully this will generate the correct schedule. If there are any technical changes I got wrong, or if someone has some tips on testing this locally, please let me know! Thanks!

purpleidea:
So please merge this, but also if someone could explain how I could get the schedule to actually generate, that would be appreciated. The "netlify" preview doesn't seem to show it :/

purpleidea:
Please merge ASAP because a lot of people are waiting to see it. Thank you!

bin/preview-site.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
```
mattstratton: What is the purpose of this additional script?

content/events/2018-montreal/program/.gitignore
```diff
@@ -0,0 +1 @@
+template.md
```
mattstratton: If you have files locally that are needed for your site to build they will likely fail. What caused the need for this ignored file?

purpleidea:
@mattstratton What comments??

bin/preview-site.sh
purpleidea: Running this script builds the site locally and makes it available in your web browser at localhost:1313

This helped reduce the cycles of pushing to Git and waiting for the builders.

content/events/2018-montreal/program/.gitignore
purpleidea: Understood. I have (as do other organizers) local template files here which we use as the basis and then populate with speaker info. We don't want to commit the template, but we don't want git to show a dirty tree. I don't know why this is an issue as it's private within our `2018-montreal` namespace so it shouldn't hurt anyone else. :)

bin/preview-site.sh
mattstratton: The `hugo watch` command does the same thing as specified in the repo’s readme files. I’m uneasy about adding additional scripts that I don’t mantain into a repo used by large numbers of folks. This script isn’t documented and all other events have been able to work without it.

content/events/2018-montreal/program/.gitignore
mattstratton: If that’s working for your use case, it should be fine. Just wanted to make sure it wasn’t needed for the build.

bin/preview-site.sh
purpleidea: It's a small one-line script which helped me out and will help out others. I work by the principle of "do-ocracy" to add helpful things where I see fit. If you don't agree, then I'll remove it, but I'd much rather merge this now and discuss that elsewhere.

mattstratton: I’ll merge this but please ask next time before adding things globally. 

@bridgetkromhout please review this as well as we have other utility scripts.

mattstratton:
Basically, for future reference please don’t add your scripts with content merges. That would allow a discussion of the script without blocking your content update.

bin/preview-site.sh
purpleidea: > I’ll merge this but 

Great, thanks!

> please ask next time before adding things globally.

Don't worry, I won't try and be helpful outside of my own corner again ;)

purpleidea:
w00t! Thanks for the fast merge. It took much longer than we expected to go through all the speakers, get acceptances, etc, and a lot of people we're constantly pinging us to know who's coming. Great to get it done!

PR #5251, opened 2018-08-29
[KC-2018] Sponsor logo maint

ThePVD:
add elastic, swap to newer logo


phrawzty:
Note to repo maintainers: This is currently the _only_ event that points at `disruptops` so the logo change is OK. 😄

PR #5255, opened 2018-08-29
[Tokyo-2019] Add placeholder content.

BenLubar:
The non-commented information is correct. The icons are temporary copies of the 2018 logos.

/cc @apapadimoulis


apapadimoulis:
👍 

NOTE we wanted to get this up soon, since we already reserved venue and the date is fixed. Content will come asap.

bridgetkromhout:
Hi - sorry for the delay - this happened to be a week that many of the site maintainers were onsite working at events. Question: can you add the other organizers? We do like to list at least three.

PR #5268, opened 2018-08-31
[NTL-2018] Adds REA and Hays as sponsors

krutisfood:


data/events/2018-newcastle.yml
```diff
@@ -14,14 +14,14 @@ cfp_date_start: 2018-07-16 # start accepting talk proposals.
 cfp_date_end: 2018-08-20 # close your call for proposals.
 cfp_date_announce: 2018-08-15 # inform proposers of status
 
-cfp_open: "true"
+cfp_open: "false"
```
phrawzty: The subject and commit message only refer to adding 2 sponsors - not to closing the CFP. I'd just like to verify that this is correct before merging. 🙂

krutisfood: Yes thanks, CFP also closed :)

PR #5299, opened 2018-09-03
Update new program script with new frontmatter

mattstratton:
This can be merged at any time, as the frontmatter in the template is backwards-compatible. However, since we aren't sure what the field will be named, let's hold off until the PR listed below is merged.

Ref: https://github.com/devopsdays/devopsdays-theme/pull/605


bridgetkromhout:
> Build cloud 'lw-host-14-e' does not support 'Visual Studio 2015' image.

Chatting with @mattstratton and this may be corrected in our build now.

PR #5310, opened 2018-09-04
[PHILLY-2018] Adding speakers.

peterjshan:
On the tin, but there is a bug it seems where the program is not rendering correctly. I'll mark as do not merge until it's resolved.


mattstratton:
Confirmed it looks good to go

PR #5332, opened 2018-09-06
[PHX-2018] Add first round of dodphx 2018 speakers

skuenzli:
Add first round of dodphx 2018 speakers and session titles


skuenzli:
Thank you, Bridget!

PR #5333, opened 2018-09-06
[Boston 2018] Adjusting for resizing

bridgetkromhout:
Note to @technovangelist and @MikeRosTX - the gulp resizing doesn't work on .jpeg (only .jpg and .png) which leads to problems like this:

screen shot 2018-09-05 at 11 58 12 pm

This PR fixes this and sizes a couple pics down to the max of 900x900px.


MikeRosTX:
Ah OK. Thanks for the confirmation @bridgetkromhout! I thought that might have been fixed, but I guess I was wrong. 😉

technovangelist:
interesting. thnaks




*Matt Williamsm@technovangelist.com
<m@technovangelist.com>http://technovangelist.com
<http://technovangelist.com>*

PR #5398, opened 2018-09-13
adds more TZ info to event data

tgross:
This PR continues the work discussed in https://github.com/devopsdays/devopsdays-web/pull/5213#issuecomment-416575633 by adding more TZ data:

@tgross wrote:
> Yeah, just to avoid a bunch of rework I'll hold off on making that data fix until the theme update gets deployed. But once that's been done I'll knock that out.

This adds all the stashed changes I had, after reconciling with the changes to events merged in since then. It only covers events from 2018 and not historical events. cc @bridgetkromhout

### DIFF
--- a/data/events/2018-amsterdam.yml
+++ b/data/events/2018-amsterdam.yml
@@ -6,19 +6,19 @@ description: "devopsdays is coming to Amsterdam!" # Edit this to suit your prefe
 ga_tracking_id: "UA-38891729-3" # If you have your own Google Analytics tracking ID, enter it here. Example: "UA-74738648-1"
 
 # All dates are in unquoted 2018-MM-DD, like this:   variable: 2018-01-05
-startdate: 2018-06-27
-enddate: 2018-06-29
+startdate: 2018-06-27T00:00:00+02:00
+enddate: 2018-06-29T23:59:59+02:00
 
 # Leave CFP dates blank if you don't know yet, or set all three at once.
 cfp_date_start: 2018-01-15T00:00:00+01:00  # start accepting talk proposals.
-cfp_date_end:  2018-04-01T00:00:00+02:00 # close your call for proposals.
+cfp_date_end:  2018-04-01T23:59:59+02:00 # close your call for proposals.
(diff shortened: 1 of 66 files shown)


bridgetkromhout:
Thanks! There are a few places you're removing recently-added sponsors - can you double-check that?

tgross:
Looks like I had an improper merge. I've fixed those and rechecked all the dates too to make sure I didn't accidentally change a date too. I think I've got it all now.

bridgetkromhout:
Ah - can all the enddates be 23:59 instead of 00:00?

tgross:
Ok, I think I've got all those now. I found a few places where folks had quotes around the dates or extra spaces that had broken some `sed` I'd done before too.

bridgetkromhout:
Daylight saving time is so evil! Thank you so much for your care on that.

A peek at the build on https://deploy-preview-5398--devopsdays-web.netlify.com/ shows we've hit a corner case for the single-day events (we want to display them with just the one day - it probably looked like duplicated code).

Here's what I mean: 

screen shot 2018-09-13 at 3 02 53 pm


screen shot 2018-09-13 at 3 03 06 pm

tgross:
Ah I see, that popped up because when we made the enddate not the same time of day this (existing) conditional fails https://github.com/devopsdays/devopsdays-theme/blob/master/layouts/index.html#L26. I'll need to change that to compare just the days, not the times.

bridgetkromhout:
> Fix for that in devopsdays/devopsdays-theme#661

Thanks, @tgross! We'll want to hold off on merging this PR until we get the theme released. /cc @mattstratton

bridgetkromhout:
(sorry, I mis-clicked and accidentally closed this - "close and comment" is too close to "comment"!)

mattstratton:
> "close and comment" is too close to "comment"!

Protip - the [Refined Github extension](https://github.com/sindresorhus/refined-github) makes those buttons far away from each other!

bridgetkromhout:
With the theme release, I'm merging this.

PR #5431, opened 2018-09-17
[WLG-2018] Welcome page, sponsor, and speaker updates

judymarcelo:


bridgetkromhout:
Hi, Judy - Are you sure you want to switch back to defaulting to UTC? Given your time zone that means your event will scroll off the front page during the event. (We recently fixed this, but you need to set the full TZ string to take advantage of the fix, which is why we set it for you.)

judymarcelo:
Thanks for updating this for us Bridget, much appreciated.

PR #5432, opened 2018-09-17
[ZRH-2018] Correct images of organizers

Roderick-Jonsson:
Just noticed that I mixed up 2 images


Roderick-Jonsson:
@bridgetkromhout Thank you Bridget!

PR #5439, opened 2018-09-18
[Bhe-2018] new sponsors

devopsdaysbh:
[Bhe-2018] new sponsors


devopsdaysbh:
Need to add the last keynote of the day. We just got our name for it =D

PR #5460, opened 2018-09-19
[Cape-Town-2018] Temporary notice on 2017 welcome page

manicminer:
Add a temporary notice to our 2017 event welcome page, due to a paste error in a sent Mailchimp campaign.


manicminer:
Thanks!

PR #5467, opened 2018-09-20
[SLC 2019] - Initial Commit for 2019

lindleywhite:
We are submitting our official date for 2019!


bridgetkromhout:
Thanks! We've set up the email aliases and deactivated the organizer who's not returning.

PR #5480, opened 2018-09-20
Adding event group for SP

bridgetkromhout:
Adding an event group for São Paulo (as the first year filename's not that).


bridgetkromhout:
(In most places we compare city name, but just to be safe.)

PR #5486, opened 2018-09-21
[GAL-2018] - add Avaya as Silver sponsor

Bryji:
Adding Avaya as a Silver level sponsor for DoDGalway 2018


bridgetkromhout:
Merging without archiving the old image, as they're identical other than whitespace.

PR #5491, opened 2018-09-21
[BH 2018] Resizes images

bridgetkromhout:
Attention: @devopsdaysbh - some of your images are too small (under 300x300 px) so they aren't resizing correctly. I'm fixing that.


devopsdaysbh:
Thanks a lot @bridgetkromhout. Yeay, I was facing a bad time trying to fit these images on our sponsors board. That's why then.

Thanks again for your help!

PR #5501, opened 2018-09-24
[PAN-2018] Fix bad html in sponsor file for panama

mattstratton:
Incorrect html tag causing build failure


mattstratton:
Dupe of #5500

PR #5521, opened 2018-09-27
[ZRH-2019] Add Tobias and Julia to organizers

Roderick-Jonsson:
Also sent mail to info@deovpsdays.org


Roderick-Jonsson:
@MikeRosTX Help, it says that it is missing an expected key at line 54, but I don't see what should be missing there

bridgetkromhout:
> @MikeRosTX Help, it says that it is missing an expected key at line 54, but I don't see what should be missing there

I suspect the unquoted `"` characters might have been the problem. Let's see if this rebuild fixes it.

Roderick-Jonsson:
@bridgetkromhout YAY thanks Bridget!

PR #5536, opened 2018-09-29
[2018 Fortaleza] Renaming speaker image correctly

bridgetkromhout:
Correcting the speaker image name, after the workaround in https://github.com/devopsdays/devopsdays-web/pull/5535.


bridgetkromhout:
Heads up to @gb8may - make sure you pull in the changes from upstream - I've renamed this file for you. Thanks.

PR #5553, opened 2018-10-03
[GAL-2018] Correct date of informing proposers due to delays

electrofelix:
Informing proposers is behind schedule, so update to let anyone checking
know that it'll be a few more days.

Reached out to most, but yet inform all.


mattstratton:
Build is failing - date format might need a preceding zero?

electrofelix:
Gah! OOps, that is a blindingly obvious mistake, sorry, will fix

PR #5570, opened 2018-10-05
[Blog] Add devopsdays 101 (pt-br) post

somatorio:
Add a "companion" blog post to the organizing page to help brazilian organizers about some basic questions at the begining


bridgetkromhout:
Thanks! Quick question about this - it sounds (from google translate) like you're saying it's a good idea for the organizers to be from some different companies. We actually _require_ that. It is mandatory to have organizers from at least three different employers. Can you make sure that is clear? It might also be a good idea to make sure it is clear that people cannot buy talk slots or attendee contact info.

gomex:
Thanks @somatorio  ;)

PR #5577, opened 2018-10-08
[KC-2018] KaizenOps logo update

ThePVD:
update with logo provided by Mark Williams 9/25/18


kmugrage:
@ThePVD we don't allow changes to logos for events which have happened in the past. For the context, see #503 where I tried to do it for my company a couple years ago ;)

I'd do it for you but don't have the time in the next couple days unfortunately.

PR #5581, opened 2018-10-08
[Cape Town 2018] Add some more slides

adrianmoisey:
Add a few more slides to some of the speaker's talks.


bridgetkromhout:
Despite the timeout, this should work.

PR #5582, opened 2018-10-08
[PT-2019] First PR of devopsdays portugal 2019 edition

eduardopiairo:
*Please title your pull request in this format: The event name and year in the title of the PR, along with a description of what is being changed, i.e., `[CHI-2018] Add Bluth Company as a sponsor`*

### DIFF
--- a/data/events/2019-portugal.yml
+++ b/data/events/2019-portugal.yml
@@ -1,0 +1,154 @@
+name: "2019-portugal" # The name of the event. Four digit year with the city name in lower-case, with no spaces.
+year: "2019" # The year of the event. Make sure it is in quotes.
+city: "Portugal" # The displayed city name of the event. Capitalize it.
+event_twitter: "devopsdayspt" # Change this to the twitter handle for your event such as devopsdayschi or devopsdaysmsp
+description: "devopsdays is coming to Portugal!" # Edit this to suit your preferences
+ga_tracking_id: "" # If you have your own Google Analytics tracking ID, enter it here. Example: "UA-74738648-1"
+
+# All dates are in unquoted 2019-MM-DD, like this:   variable: 2016-01-05
+startdate:  # The start date of your event. Leave blank if you don't have a venue reserved yet.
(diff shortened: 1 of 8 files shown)


bridgetkromhout:
Let's chat about how to link events - I think that feature might not be clear? It's how events move around in a country (like New Zealand for example - https://www.devopsdays.org/events/2018-wellington/welcome/). What you want to do is use an "event_group".

data/events/2019-portugal.yml
```diff
@@ -0,0 +1,154 @@
+name: "2019-portugal" # The name of the event. Four digit year with the city name in lower-case, with no spaces.
+year: "2019" # The year of the event. Make sure it is in quotes.
+city: "Portugal" # The displayed city name of the event. Capitalize it.
+event_twitter: "devopsdayspt" # Change this to the twitter handle for your event such as devopsdayschi or devopsdaysmsp
+description: "devopsdays is coming to Portugal!" # Edit this to suit your preferences
+ga_tracking_id: "" # If you have your own Google Analytics tracking ID, enter it here. Example: "UA-74738648-1"
```
bridgetkromhout: This is how other countries do it: https://github.com/devopsdays/devopsdays-web/blob/master/data/events/2018-wellington.yml#L7

So you would add this here:

`event_group: "Portugal"`

bridgetkromhout:
Please identify an individual event with a city in the URL. You can group cities with an event group if you're going to move it to other cities over time. 

- Sometimes events think they will move, but never actually do.
- Sometimes a new city in the country wants to run their own event and not feel excluded or feel like they are second-class.
- It's confusing if we have scope all over the place, so we're trying to focus on cities that the event will happen in.

Thanks!

eduardopiairo:
Hello Bridget,

We already had the discussion around the event name with other core organizers through email. This PR with the name "DevOpsDays Portugal" was made following the last instructions from @KrisBuytaert .

Thank you.

PR #5585, opened 2018-10-09
[AMS-2019] Add Sheela as local organizer

dpnl87:
Please let me know if we need any additional information to add Sheela to the local AMS org team.


MikeRosTX:
Before we can merge this PR, we need an email sent to info@devopsdays.org requesting Sheela Nistala be added a a new organizer along with her email so we can add her to Slack and to AMS dl email.

PR #5590, opened 2018-10-10
2018-montreal: Fix schedule timing

purpleidea:
This entry was put in as an inside joke, but we should probably update
it so attendees see the right time =D

*Please title your pull request in this format: The event name and year in the title of the PR, along with a description of what is being changed, i.e., `[CHI-2018] Add Bluth Company as a sponsor`*


purpleidea:
Thanks Mike!

PR #5602, opened 2018-10-11
[Kiev-2019] add Provectus and Percona sponsors

delgod:


bridgetkromhout:
This looks like an intermittent failure:

```
[NuGet] Could not connect to the feed specified at 'https://chocolatey.org/api/v2/'. Please verify that the package source (located in the Package Manager Settings) is valid and ensure your network connectivity.
hugo not installed. The package was not found with the source(s) listed.
 If you specified a particular version and are receiving this message, it is possible that the package name exists but the version does not.
 Version: ""
 Source(s): "https://chocolatey.org/api/v2/"
Chocolatey installed 0/1 packages. 1 packages failed.
 See the log for details (C:\ProgramData\chocolatey\logs\chocolatey.log).
Failures
 - hugo - hugo not installed. The package was not found with the source(s) listed.
 If you specified a particular version and are receiving this message, it is possible that the package name exists but the version does not.
 Version: ""
 Source(s): "https://chocolat
```

PR #5615, opened 2018-10-13
[GVA-2019] Move event to 2019 and add dates, location & logo

5ika:


bridgetkromhout:
I've corrected errors; please send another PR if you want changes.

PR #5620, opened 2018-10-14
[EDI 2018] Adding some speaker images & corrections

bridgetkromhout:
Hi, @andyburgin - I knew which ones to use for a couple of the images you're missing, and I corrected a few typos.


kevinhcross:
Hi,

It was me that added Louda using this script: https://github.com/devopsdays/devopsdays-web/blob/master/utilities/add_speakers.sh

I did notice the name came out "louda-pea" but I assumed that was because something on the site would not deal with anything non-ascii so I left it as was. In hindsight I probably should have changed `ñ` to `n` but it was displaying OK in the site and did not notice that the URL was "louda-pea".

bridgetkromhout:
Hi, @kevinhcross - I wrote that bash script and yes, it doesn't handle special characters that shouldn't be in a URL (and we could definitely improve it to give a more explicit error about that).

@andyburgin, good catch! I missed updating the yaml and that link. Fixed and the program page is now linked properly. Please let me know if https://deploy-preview-5620--devopsdays-web.netlify.com/events/2018-edinburgh/welcome/ looks right to you now and if we should merge this.

andyburgin:
Hi Bridget, looks good, please go ahead and merge.

PR #5630, opened 2018-10-16
[SLC-2019] Update Organizers

lindleywhite:
Add Heather Walker and Curtis Salisbury as organizers for SLC


bridgetkromhout:
Please email info@devopsdays.org to send us their email addresses. We update the website, email list, and Slack all together. Thanks.

PR #5637, opened 2018-10-16
[KC-2018] Replace Cory with ADO

bksteiny:
We've had a speaker back out, so we are improvising.


phrawzty:
Netlify failure is unrelated to PR.

PR #5641, opened 2018-10-17
 [Kiev-2019] add ab-soft sponsor

delgod:


bridgetkromhout:
Hi - We can make the official name in the URL spelled Kyiv if you like (and redirect the Kiev URLs) - is that what you prefer?

delgod:
@bridgetkromhout, both links work fine now, I have added an alias to `welcome.md` some time ago.
I am ok with the current default link.

bridgetkromhout:
Okay. We have a feature called "event_group" that would let you group together events with the city names spelled differently, so if you decide to change the official URL for 2019, your previous 2018 event will still be grouped with it.

PR #5710, opened 2018-10-29
Add Serhat and Nathen to core.toml

kmugrage:


devopsdays-service:
@kmugrage, thanks! Based upon the content in this PR, we need a review from @mattstratton and @bridgetkromhout

bridgetkromhout:
Ah, thanks! Indeed, I got busy and forgot to do this.

PR #5724, opened 2018-11-01
[GAL-2018] Update logo with final version

electrofelix:
Update the logo to match the final version that will be used during the
event and for any printed materials.


phrawzty:
Not sure why netlify failed here. @electrofelix, can you confirm that this works on local?

electrofelix:
@phrawzty yeah it was display correctly for me locally so I wasn't sure how to re-trigger to see if it was just a transient error

PR #5728, opened 2018-11-01
[INDY-2019] We're back!

ghoneycutt:


mattstratton:
You won’t get it locally since it comes from running `html-min` but this will break in prod. 

Here’s the problem - some bad html on your sponsor page:

```
    Parse Error: <td style="padding: 5px" bgcolor="silver"</td>
```

The opening TD tag is missing the `>`

mattstratton:
I pushed in a fix - let’s see if that does it.

ghoneycutt:
Thanks @mattstratton !!

PR #5729, opened 2018-11-01
Update Hugo to 0.50

ghoneycutt:


mattstratton:
Can i convince you to also update the Hugo version reference here? 

https://github.com/devopsdays/devopsdays-web/blob/master/CONTRIBUTING.md

bridgetkromhout:
@mattstratton can we make sure this is tested and present in the themes repo first, so we see it working in the example site?

ghoneycutt:
@mattstratton Got it updated with a pointer to `.travis.yml`, so it does not have to be updated in unison again.

PR #5731, opened 2018-11-02
[GAL-2018] Add IBM as gold sponsor

j3p0uk:
Also turn off sponsors as we need to get printing done - just 2 weeks to go!


phrawzty:
Why a specific 2018 sponsor profile for IBM?

electrofelix:
@phrawzty I think there is uncertainly that it's exactly the same logo as  https://www.devopsdays.org/img/sponsors/ibm.png or  https://www.devopsdays.org/img/sponsors/ibmcode.png but maybe it is with just a white background?

phrawzty:
Imho the only difference is whitespace. Is this a specific request from IBM?

j3p0uk:
So we replied to the person in IBM with a pointer to the existing logo and they came back with the new one.  There was no specific mention about the existing one being wrong, but there was a lot of talk about getting the logo to use ok'ed by the legal department.  IMHO it is a slightly different shade of blue, and is also cropped tight top and bottom instead of square.  I also exported from the hi-res eps as a transparent png.  That covers all of the differences, but I would agree that they are largely the same.

j3p0uk:
Also I haven't validated the URL that the sponsor wishes to use, the various yaml files have various URLs.  (Note to self, add some questions to tickets for next year)

PR #5735, opened 2018-11-03
[DC-2019] Prepare the site

nathenharvey:
Event date and location are TBD.


devopsdays-service:
@nathenharvey, thanks! Based upon the content in this PR, we need a review from @mattstratton and @bridgetkromhout

bridgetkromhout:
Hey, @nathenharvey - sorry for the delay! As you may have guessed, this got flagged because you updated a utility script in the same pull request as your event announcement. While the change you made is fine, you'd be surprised/horrified at how late in the year people sometimes instantiate new events for the same year (and conversely, how frequently they sometimes are preparing the following year's event much earlier than we'd think!)

PR #5739, opened 2018-11-04
[NYC 2019] Fix image

bridgetkromhout:
heads up, @TomOnTime - fixing this so the image shows up.


bridgetkromhout:
This is what happens if you use a png for an organizer image:

screen shot 2018-11-04 at 1 13 56 pm

PR #5757, opened 2018-11-07
[INDY-2019] registration and cfp dates still under discussion, removing for now

apoland:
registration and cfp dates still under discussion, removing for now


phrawzty:
Confirmed via Slack; merging.

PR #5787, opened 2018-11-12
[TLV-2018] adding Sabine, change fiverr link, fixing program for continued workshop

gilzellner:
[TLV-2018] adding Sabine, change fiverr link, fixing program for continued workshop

we added fiverr as a sponsor recently and nobody else has had them yet.

### DIFF
--- a/data/sponsors/fiverr.yml
+++ b/data/sponsors/fiverr.yml
@@ -1,2 +1,2 @@
 name: "Fiverr"
-url: "https://www.fiverr.com/jobs/offices/tlv"
+url: "https://www.fiverr.com/"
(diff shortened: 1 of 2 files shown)


data/sponsors/fiverr.yml
```diff
@@ -1,2 +1,2 @@
 name: "Fiverr"
-url: "https://www.fiverr.com"
+url: "https://www.fiverr.com/jobs/offices/tlv"
```
phrawzty: Suggest leaving this as the generic URL (so that it can be used by other events) and setting a local override for your event ([example](https://github.com/devopsdays/devopsdays-web/blob/master/data/events/2018-paris.yml#L108-L110)).

gilzellner: done

PR #5800, opened 2018-11-14
[FLN-2018] Change program, speakers and sponsor informations

jhonatanmorais:
This pull request:

- Add new talks and speakers and Fix event program
- Fix sponsor logo: Softplan
- Add new sponsor: InfoQ Brasil
- Remove cancelled talks and speakers


bridgetkromhout:
As this is the only time softplan's logo appears so far, changing it in this PR is fine (we don't need to save the old version).

PR #5807, opened 2018-11-15
[DC-2019] Adds more organizers for DC 2019

nathenharvey:
Signed-off-by: Nathen Harvey <nathen.harvey@gmail.com>


nathenharvey:
Email list and slack invites are complete

PR #5837, opened 2018-11-20
date -v doesn't work on most Linux distros,  rewrote so it works on L…

KrisBuytaert:
date -v doesn't work on Linux , so I rewrote this to work on Linux, this has been tested on OS also.


devopsdays-service:
@KrisBuytaert, thanks! Based upon the content in this PR, we need a review from @mattstratton and @bridgetkromhout

PR #5880, opened 2018-12-02
[LON-2018] add videos of talks and slides

andyburgin:
Added video links
Ignite pdfs added and already merged into devopsdays-assets


andyburgin:
Hi, just checked the preview and it looks like the hugo theme layout borks when there's both a video on pdf deisplayed. Will look at changing the pdf to a download link instead.

Please decline this PR

PR #5885, opened 2018-12-03
[AMS-2019] Re-add Gerard to the local organiser team

dpnl87:
Gerard was wrongly deleted, added him back to the list.


phrawzty:
Hello,

I also re-added this person back to the AMS-2019 mailing list (based on the the registered email address in Slack). In the future, for org changes, please also transmit an associated email address as well. Thanks!

PR #5902, opened 2018-12-06
[HOUSTON-2019] Adding proposal link, adding bio for one of the organi…

skryukova:
…zers

[HOUSTON-2019] Adding proposal link, adding bio for one of the organizers


data/events/2019-houston.yml
```diff
@@ -10,12 +10,12 @@ startdate:  # The start date of your event. Leave blank if you don't have a venu
 enddate:  # The end date of your event. Leave blank if you don't have a venue reserved yet.
 
 # Leave CFP dates blank if you don't know yet, or set all three at once.
-cfp_date_start:  # start accepting talk proposals.
-cfp_date_end:  # close your call for proposals.
-cfp_date_announce:  # inform proposers of status
+cfp_date_start: 
+cfp_date_end: 
+cfp_data_announce:
```
bridgetkromhout: Please fill these dates in - that's what lists your event on the "speaking" page.

skryukova: Filling in those dates without filling in the start/end date seems to cause compilation error. 
Since we don't have a venue secured yet, I was hesitant to put the end/start date of the event.

MikeRosTX: I'll add tentative dates to match what is posted in PaperCall.io.
The venue options are in discussion.

MikeRosTX: Not sure why these dates are failing. Maybe @bridgetkromhout or @mattstratton can give it a quick glance tomorrow if they have some free time.

bridgetkromhout: Ah, right, I didn't notice you haven't published an event date. You can't list CFP dates in that case, because the site logic is showing CFPs that are open for events that haven't happened yet. We can merge without dates, but I strongly suggest adding them as soon as you have event dates, since you won't be on the single CFP page without dates.

MikeRosTX: OK, that makes a lot of sense. I was suspecting that was going to be the case.
Thanks @bridgetkromhout! Proceed with merge without CFP dates until event dates are available..

PR #5921, opened 2018-12-10
[LDN-2019] Add Dominique to the Organisers

chriscoffee:
We've added Dominique as an organiser so get Dominique on the site.

Signed-off-by: Chris Mills <me@christophermills.co.uk>


chriscoffee:
This is a duplicate PR. Apologies maintainers

PR #5957, opened 2018-12-17
[LDN-2019] Enable accessibility and input CFP dates.

chriscoffee:
We've got some CFP dates now so I've added those but not enabled the
page as of yet. I've also enabled the accessibility page as it's
worthwhile having this after discussing in our weekly meeting.

Signed-off-by: Chris Mills <me@christophermills.co.uk>

### DIFF
--- a/content/events/2019-london/accessibility.md
+++ b/content/events/2019-london/accessibility.md
@@ -1,0 +1,17 @@
++++
+title = "accessibility"
+type = "event"
+description = "Accessibility information for devopsdays London 2019"
++++
+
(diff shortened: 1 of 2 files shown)


chriscoffee:
Thanks a lot for this @bridgetkromhout. I just wrongly assumed it was hidden until `cfp_open` was true. :smile: 

For those reading this not sitting in the #website channel on the devopsdays slack there's some discussion around this here: https://devopsdays.slack.com/archives/C07T0BHHN/p1545056948058800

bridgetkromhout:
I'm not positive cfp_open works that way. @mattstratton may know offhand or we can test, but it was a definite worry when I saw it.

PR #5954, opened 2018-12-17
[STHLM-2019] Add new members to stockholm team

JasmineCrozier:
I added 3 persons to our team

PR #5958, opened 2018-12-17
[Stockholm-2019] Add organizers for Stockholm

bridgetkromhout:
Adding the organizers initially tried in https://github.com/devopsdays/devopsdays-web/pull/5954


JasmineCrozier:
Thank you for the help on this @bridgetkromhout :)

PR #5959, opened 2018-12-17
[BNA-2019] Added event organizer

notsureifkevin:


phrawzty:
Please send an email to info@devopsdays.org with the full name and email address of the new organiser, so that they can be added to the mailing list and Slack.

PR #5975, opened 2018-12-20
[SP-2019] Add new organizers

camilafrasan:


camilafrasan:
I'll close this PR because it has changes that are not part of the event (São Paulo).

PR #5993, opened 2018-12-23
[Los Angeles 2019] Adjusting CFP opening

bridgetkromhout:
Hi, @irabinovitch - it appears that the CFP opening date was set to January when this was first listed, months ago. This prevents you from showing up (yet) in https://www.devopsdays.org/speaking/ - so I'm adjusting it to a few days ago when you recently updated the site. Please let us know if you actually don't want to be listed. Thanks!


irabinovitch:
@omargohan and @cwebberOps can you all confirm the CFP dates and update?

PR #6007, opened 2018-12-30
[SP-2019] Add new organizers

camilafrasan:


MikeRosTX:
@camilafrasan - Please send an email to **_info@devopsdays.org_** with all 4 new organizer's **_full name_** and their **_emails_** so that we can add them to the team dl alias and invite them to join Slack.

PR #6034, opened 2019-01-07
[HOU-2019] Adjusting speaker photo to have square dimensions

skryukova:
[HOU-2019] Adjusting speaker photo to have square dimensions


skryukova:
I didn't see that Bridget already replaced the image.

PR #6070, opened 2019-01-10
[NYC-2019] Add Matthew Walter as an organizer

ohaiwalt:
Already a member of the Slack team.

I'll ask another organizer to send the email to info@devopsdays.org with my full name and email so I can be added to the email alias.


kuritz:
Dropping this here in case approval is needed from current organizer. Thanks!

PR #6095, opened 2019-01-15
[CPH-2019] Add Keynote Speaker Matt Stratton

JasmineCrozier:


bridgetkromhout:
The "matt-stratton@2x.png" file shouldn't have the 2x already in it. I'll correct.

PR #6105, opened 2019-01-16
Use latest Hugo - v0.53

ghoneycutt:


ghoneycutt:
We'll see what Travis has to say.. though I'm using v.053 on my workstation to preview the site and it works as expected.

ghoneycutt:
CC @mattstratton

PR #6139, opened 2019-01-21
[MSP 2019] Added full datetime

bridgetkromhout:
I think this error is from having incomplete datetime in the CFP fields:

<i>
5:15:01 PM: ERROR 2019/01/21 23:15:01 theme/partials/events/cta.html template: theme/partials/events/cta.html:54:60: executing "theme/partials/events/cta.html" at <time $e.cfp_date_end>: error calling time: Unable to Cast <nil> to Time
</i>


bridgetkromhout:
The line it's referencing:
<i>
      {{- if and (ge now (time $e.cfp_date_start)) (le now (time $e.cfp_date_end)) -}}
</i>

PR #6147, opened 2019-01-22
[BMORE-2019] Adding Fearless rebrand to sponsor; archiving old logo a…

bushong1:
…nd updating old events to retain previous logo

I followed the pattern of the Victorops logo update.  Please let me know if there's something else I need to do in the case of logo updates.


phrawzty:
This looks correct. Thanks for respecting the old logo preservation policy. :)

Issue #6157, opened 2019-01-22
investigate jpg/jpeg mismatch in build


bridgetkromhout:
We appear to have an issue with the build where front-page images were broken/not loading after a normal full build (because of linking to jpg when jpeg was what existed). We noticed this at the same time as a probably-unrelated issue: https://github.com/devopsdays/devopsdays-web/pull/6139#issuecomment-456562609 (that is in theory now resolved via https://github.com/devopsdays/devopsdays-web/pull/6152).

Temporary patch for the image issue: https://github.com/devopsdays/devopsdays-web/pull/6145/files

When I test locally to run the full build (without that patch), we get links with .jpg:
```
<a href="/events/2019-new-york-city"><img src="/events/2019-new-york-city/logo-square.jpg" class="img-fluid" alt="devopsdays New York City" srcset="/events/2019-new-york-city/logo-square.jpg 1x, /events/2019-new-york-city/logo-square@2x.jpg 2x"></a>
```
However, the (smaller than original, so, processed) files generated which appear post-gulp have .jpeg extensions:
<i>
dist/events/2019-new-york-city/logo-square.jpeg
dist/events/2019-new-york-city/logo-square@2x.jpeg
</i>

I think some of the tasks in https://github.com/devopsdays/devopsdays-web/tree/master/gulp/tasks must be what's rewriting these files, but unsure exactly what is setting the output filename.

While it _sounds_ like this could be related to the image minimization in https://github.com/devopsdays/devopsdays-web/issues/5748, I tested locally setting it explicitly to 5.0.2 and that didn't seem to change the behavior.

bridgetkromhout:
My guess it's something we don't have the version pinned on, or via some other way the build is doing something unexpected. But I think I can rule out our content as being the problem, in that the problem manifests when we build in this specific pipeline, but not without the image pipeline.

mattstratton:
For what it's worth, this is the code that does the logo resizing/saving:

https://github.com/devopsdays/devopsdays-web/blob/master/gulp/tasks/responsive-images.js#L12-L38

mattstratton:
Bingo - https://github.com/mahnunchik/gulp-responsive/issues/113

PR #6162, opened 2019-01-23
Adjusting npm version limits

bridgetkromhout:
Because of the "breaking change in a minor release" drama chronicled in https://github.com/devopsdays/devopsdays-web/issues/6157, I'm proposing adjusting all these versions to where they're at right now, with the following changes:

- instead of `^` which allows changes as long as they aren't major version changes, switching to `~` which keeps the minor version the same and only allows for point-release changes to be applied automatically

- In some cases we don't use the `~`, like for gulp-imagemin which has a breaking change according to https://github.com/devopsdays/devopsdays-web/issues/5748.

If we merge this PR, we may need to be more hands-on with updates (and if we don't, it might be challenging to catch up all at once if we have to). The upside would be way less in the way of mysterious breakage without us knowing what changed. I think the positive outweighs the potential downside, but this is a @mattstratton call.


mattstratton:
👍

Ship it!

PR #6184, opened 2019-01-25
[ZRH-2019] Sold out coffee and entrance

Roderick-Jonsson:
Adding sold out to coffee and entrance only. For some odd reason git also is adding the last commited changes.


bridgetkromhout:
Okay, so now that I have time I'll explain what's happening. The problem you're running into is that you're quite far behind, and this is a problem because from time to time, the core team needs to make corrections to cities (like when you input a sponsor incorrectly, or when a sponsor logo changes).

screen shot 2019-01-26 at 3 35 56 pm

Notice how far behind you are. That's why this is happening. To fix this, you need to do a `git pull upstream master` into your fork's master branch.

In an attempt to make this easier for you, I'm removing and editing files I can via this PR. (The image, I'll need to delete after merging.)

PR #6201, opened 2019-01-29
[BER-2019] Change in local Organizers

range:
Added a new organizer for the Berlin event.


phrawzty:
(confirmed via Slack that ML and Slack are in alignment)

PR #6208, opened 2019-01-30
[ATX-2019] Updating organizers list

scottsbaldwin:


phrawzty:
Hello! Please send an email to info@devopsdays.org with the names and email addresses of those who have been added/subtracted/modified, so that we can keep the mailing list, slack, and site in sync. Thanks!

PR #6265, opened 2019-02-06
[CPH-2019] Add thoughtworks

JasmineCrozier:
I added thoughtworks as silver sponsor.

I could not view the local website, it gave me an error message that pointed to {{< param year >}} in the Boise event. So I temporarily removed it and put it back. I did not really change anything, but this is why it shows in my commits :)

If anyone can suggest a solution to that problem, I'm interested :)


phrawzty:
Concerning the error message, let's follow up on Slack. 🙂 I'll merge this now.

PR #6313, opened 2019-02-12
[Cape Town 2019] Adding logo to front page

bridgetkromhout:
The logo needs to be a jpg and named this way to show on the front page. /cc @adrianmoisey in reference to https://github.com/devopsdays/devopsdays-web/pull/6312.


adrianmoisey:
Ah right! Thanks for making the PR! 👍

PR #6327, opened 2019-02-13
[Cape Town 2019] Use the propose page

adrianmoisey:
To explain to people what the CFP is about, before sending them to
papercall


manicminer:
LGTM!

PR #6328, opened 2019-02-13
Buffalo-2019 - Adding dates for the event

aspleenic:
Adding the dates of the event, the location, and map information.


bridgetkromhout:
Hi - is this a one-day event? That's fine, of course - I just want to make sure you understand that the timestamps you picked means it will be removed from the front page of the website before it's actually over, if it's a two-day event.

aspleenic:
This is a two day event - September 26th and 27th I will update this in a separate PR

PR #6359, opened 2019-02-19
[AMS-2019] Update AMS 2019 dates

sheela-nistala:
We decided to switch from a 4 day event to 3 day event. Hence the change in dates.


phrawzty:
I’d just like to double check. The PR is good, and I’ve no reason to distrust you (hehe), but whenever dates get changed for an event, it’s good policy to _make sure_. :) Checking in on Slack…

PR #6360, opened 2019-02-19
[GENEVA-2019] update Daniel's workshop title & description

phrawzty:
The title and description were taken from the original presentation proposal (which was a traditional _talk_). This PR modifies the title and description to more accurately match the _workshop_.


phrawzty:
👍 from Matteo on Slack.

PR #6389, opened 2019-02-21
[CHI-2019] Add dates, CFP, sponsorship, and more

mattstratton:
We have a contract! This PR will enable the dates, CFP, sponsorship, etc for Chicago 2019


mattstratton:
Need one final review before merging, so please hold off right now.

bridgetkromhout:
This was breaking because of a misplaced image we've corrected, so you'll want to merge master in again.

mattstratton:
@bridgetkromhout I merged master back in and now we are passing all checks. We have a final thing to check before merge though :(

PR #6422, opened 2019-02-27
[ZURICH-2019] Add Zuehlke as bronze sponsor

Roderick-Jonsson:
YAY we got Zühlke as bronze Sponsor :D


phrawzty:
The Netlify check is being stinky today. This is a small change and it looks fine (plus it builds local, so…).

Roderick-Jonsson:
@phrawzty Thanks for doing a local test and then merging it so fast. Merci beaucoup 🏆

PR #6428, opened 2019-02-27
[Ist 2019] Trying setting all three dates

bridgetkromhout:
All three dates need to be set. I picked an arbitrary date later than the close date for the reveal (as they aren't displaying that on the site anyhow).


bridgetkromhout:
Heads up to @mehmettaskiner - I edited your config because you need start, end, and release dates for the CFP. Please go ahead and edit the program release date if desired; I just guessed.

PR #6429, opened 2019-02-27
[PT2019] claranet and laredoute as bronze sponsors

eduardopiairo:
* claranet as bronze sponsor
* laredoute as bronze sponsor


bridgetkromhout:
Despite the timeout, this is likely to build correctly.

PR #6436, opened 2019-02-28
[ATX-2019] CFP announce date is now 3/11

scottsbaldwin:
* Also mark sponsorships as sold out.


bridgetkromhout:
Some netlify builds have been timing out so there's no preview, but this LGTM.

PR #6438, opened 2019-02-28
[STH-2019] Add propose menu item

krakan:
Tomorrow we open the CFP. The menu item is not controlled by the date as the button above the share links is. But as it opens at midnight now is a good  time to ad the ink.


MikeRosTX:
Code looks good, so I merged despite Netlify have some hiccups. Just make sure to confirm later that your pages got updated.

PR #6444, opened 2019-03-02
testing netlify revert

bridgetkromhout:
Not positive that this change worked since everything since this times out - figured we could try this.


mattstratton:
No no no. This will totally break. The `dist` output only gets created if gulp runs.

bridgetkromhout:
Ah, okay!

PR #6450, opened 2019-03-02
[BOG-2019] Updating location, propose, and navbar

MikeRosTX:


MikeRosTX:
@devopscolombia, just want to let you know of some changes.

PR #6457, opened 2019-03-04
[RDU-2019] change propose page and remove direct link to form

normalfaults:
[RDU-2019] change propose page and remove direct link to form


bridgetkromhout:
Merging despite timeout.

PR #6465, opened 2019-03-05
[2019-HOU] Added more speakers, talks and updated program

skryukova:
Copied data about speakers (Aaron Aldrich) from other cities, added "blank" image for speakers without images, added talk details based on papercall.io information.


mattstratton:
@skryukova 	for what it's worth, you don't need to add blank images - if you don't set an image, the site displays a default.

mattstratton:
Hmm. I think using the blank instead of default makes the page look confusing:

image

PR #6473, opened 2019-03-05
[Columbus 2019] Creating event

kjenkins19:
This is the initial creation of the Columbus 2019 Event. This is just a bare minimum to get it in there with the dates (which were very limited unfortunately). Will need the 2019 emails created, please copy the organizer email list from last time and keep slack the same.


bridgetkromhout:
Set up email; Slack already set up.

PR #6492, opened 2019-03-07
[Vancouver 2019] Removal

bridgetkromhout:
removal of unconfirmed speaker /cc @Velutas


Velutas:
Whoops! There are three others that should be removed as unconfirmed too. I'll make a PR to take them out too.

Velutas:
Thanks!

PR #6499, opened 2019-03-07
[CPH-2019] adding matthias and redhat

JasmineCrozier:
adding a speaker, a sponsor and mob programming sessions


phrawzty:
Netlify timed out, but the PR looks good (I verified locally).

PR #6504, opened 2019-03-07
[INDY-2019] Extend CFP two weeks

ghoneycutt:

### DIFF
--- a/data/events/2019-indianapolis.yml
+++ b/data/events/2019-indianapolis.yml
@@ -13,7 +13,7 @@ enddate: 2019-07-26
 # Leave CFP dates blank if you don't know yet, or set all three at once.
 cfp_date_start: 2018-12-06 # start accepting talk proposals.
 cfp_date_end: 2019-03-22 # close your call for proposals.
-cfp_date_announce: 2018-12-06 # inform proposers of status
+cfp_date_announce: 2019-04-07 # inform proposers of status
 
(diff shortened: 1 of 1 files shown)


ghoneycutt:
the deploy/netlify error seems unrelated

ghoneycutt:
@MikeRosTX would like the announce date change?

bridgetkromhout:
> @MikeRosTX why would the announce date change?

Hi, @ghoneycutt - the announce date as it stands is invalid - it's already in the past. Please pick a different one in the future if you don't want to use the one Mike picked - thanks!

bridgetkromhout:
Perhaps this is a confusion in what the terms mean?
```

cfp_date_start | YYYY-MM-DD | No | The date you will start accepting talk proposals. Can be a blank value. | 2016-01-05
cfp_date_end | YYYY-MM-DD | No | The date you will close your call for proposals. Can be a blank value. If you set the full datetime string with correct TZ, CFP will display as open until local close time. | 2016-01-05T23:59:00-06:00
cfp_date_announce | YYYY-MM-DD | No | The date you will inform proposers of status. Can be a blank value. | 2016-01-05
```

So technically you can leave the announce date blank although we may use it to populate information in https://www.devopsdays.org/speaking/ at some point, so it's better to set it.

ghoneycutt:
Aha! Looks good to me. Thank you!

PR #6505, opened 2019-03-07
[STH-2019] Add first sponsor and remaining organizer photos

krakan:
Adding the rest of our pretty faces plus our first confirmed sponsor.


bridgetkromhout:
Because Emily Bache is being removed from the website, I also removed her from email and Slack.

PR #6521, opened 2019-03-10
[DSM-2019] Add Aureon as lanyard sponsor

timgifford:
Also updated to a Granular vertical logo


MikeRosTX:
Please rename the old Granular logo like this before we can merge it:
**_granular-before-20190309.png_**

PR #6534, opened 2019-03-11
[STH-2019] change Johan's photo to JPEG

krakan:
The PNG doesn't display on the website although it does locally.


bridgetkromhout:
Right, it's a little confusing, but as is says in https://github.com/devopsdays/devopsdays-theme/blob/master/REFERENCE.md#organizer-fields - "The name of the image for this user, located in static/events/YYYY-CITY/organizers/. This image must be a JPEG, either 300px square or (optimally) 600px square."

PR #6557, opened 2019-03-12
[CPH-2019] Changing xebialabs logo

JasmineCrozier:
Rose from Xebialabs asked me to change the logo for our event. I was not sure how to modify it in the repo because there was two xebialabs files (xebialabs-bv). I made 2019-xebialabs in order not to mess up with previous events. Please modify if it doesn't match your system.


bridgetkromhout:
The correct way to adjust sponsor logos is to preserve history. I'll submit a PR that does it.

PR #6580, opened 2019-03-13
[PHILLY-2019] Add sponsors and prospectus.

peterjshan:
On the tin mostly. This also adds some images of our prospectus we want to use later.


bridgetkromhout:
I've removed Sam from Slack and email, per this PR.

PR #6591, opened 2019-03-14
Datetime datafile and template fix

bridgetkromhout:
We have a number of sort-order bugs in the theme that will be much easier to address when events have correct, full, usable datetime strings.

In this PR I am ensuring all 2019 events use full datetime strings with correct TZ offsets. I'm adding a clarifying comment to every event yaml file so hopefully people will submit usable datetime strings going forward.

I've corrected incorrect offsets and erroneous `"`s in a couple places as well. I've taken care not to change anyone's meaning.

I'm also changing the template so that new events are guided in the right direction.


bridgetkromhout:
Spot checking the deploy preview, I don't see anything looking _worse_ than it does now: https://deploy-preview-6591--devopsdays-web.netlify.com/

PR #6593, opened 2019-03-14
Remove references to absURL to help deploy previews

mattstratton:
This is a fairly major change to the theme, but it will stop us using absolute URLs which should help build times in deploy previews.


bridgetkromhout:
I spot-checked and this looks like it works.

PR #6604, opened 2019-03-15
[CHN-2019] Add reg link and chang log file

martinliu:
* registration is open, add reg link
* change Beijing logo file

Thanks and regards,
Martin Liu


martinliu:
need more test.

PR #6607, opened 2019-03-15
[VIC-2019] Add Semaphore as sponsor, update organizers

avandendorpe:
Updated the organizers list for this year and added Semaphore as a Bronze sponsor


BGnoinski:
Can confirm Alain is with us and not just adding himself to the event.

phrawzty:
> Can confirm Alain is with us and not just adding himself to the event.

Please send an email to info@devopsdays.org with the email address of this person so that we can add them to Slack and your mailing list.

PR #6617, opened 2019-03-16
[Denver 2019] CFP is closed

bridgetkromhout:
Since the CFP date has passed - /cc @endzyme


endzyme:
Thanks Bridget! Sorry missed that!

PR #6620, opened 2019-03-16
[DEN-2019] Adjusting organizers

endzyme:
Thanks!


bridgetkromhout:
Thanks; I also made those changes on Slack and the email list.

PR #6628, opened 2019-03-18
[CPH-2019] Add Kenneth and Microsoft

JasmineCrozier:


bridgetkromhout:
Hi, can you change the ø character in the file names to o or another ascii character? We are using ascii in urls to avoid problems. You can use any characters you like in the descriptions, but not the file names.

RandomSort:
Hi Bridget,
files should be renamed now. Cheers!

JasmineCrozier:
@bridgetkromhout Johan fixed it :)

Issue #6625, opened 2019-03-17
investigate speaking page issue


bridgetkromhout:
I'm not sure why this started happening.

<em>
Bridgets-MacBook:devopsdays-web bridget$ hugo --theme=devopsdays-theme --buildDrafts=false --baseURL="https://www.devopsdays.org/"
Total in 84512 ms
Error: Error building site: failed to render pages: render of "page" failed: execute of template failed: template: speaking/single.html:9:4: executing "main" at <partial "speaking.ht...>: error calling partial: "/Users/bridget/github/devopsdays/devopsdays-web/themes/devopsdays-theme/layouts/partials/speaking.html:12:34": execute of template failed: template: partials/speaking.html:12:34: executing "partials/speaking.html" at <dateFormat "2006-01-...>: error calling dateFormat: unable to cast <nil> of type <nil> to Time
</em>
<p>

There is probably a data problem in some CFP dates. But then, I'm unsure why it didn't start happening when I put all the full datetime strings in. It started happening with a build that didn't touch that, afaict.  I wonder if I broke it when I added datetimes and then we were operating off cache or something like that?

I put a temporary patch in (https://github.com/devopsdays/devopsdays-web/pull/6624) to alleviate the symptoms (which consisted of "site builds would fail with the above error"). I can duplicate the error locally, so it's debuggable. This patch allows the site to build, so our production builds are fine. (The CFP listing just doesn't appear on the speaking page until we solve this.)

jakubjakubik:
https://github.com/devopsdays/devopsdays-web/blob/master/data/events/2019-montevideo.yml#L9

the start date is NIL, but 

https://github.com/devopsdays/devopsdays-web/blob/master/data/events/2019-montevideo.yml#L14

cfp_date_end is defined - thus iterating over events that have the cfp_date_end defined tries to format a nil startdate

bridgetkromhout:
Thanks, @jakubjakubik! In this case, this is behaving as desired (because we do not want to list a CFP when the event dates are unknown). We just hadn't had time to figure out where the problem was. In this case, it was bad data, which I am reverting in https://github.com/devopsdays/devopsdays-web/pull/6639.

bridgetkromhout:
Upon further review, as this will allow CFP dates without event dates only on an event's own pages, and will not allow the event to appear on the `/speaking` page, I think it's wiser not to break the build (even if it allows events to list their CFPs before their event dates, which we do not want). So I have merged the fix. Thanks!

[...]

PR #6637, opened 2019-03-18
FIX issue #6625 - investigate speaking page issue

jakubjakubik:
Fix for the speaking page issue


devopsdays-service:
@jakubjakubik, thanks! Based upon the content in this PR, we need a review from @mattstratton, @MikeRosTX, @phrawzty and @bridgetkromhout

bridgetkromhout:
Ah! Thank you for surfacing this; this let me find the data problem!

We actually do not want to merge this, because we do not _want_ events to be able to list an open CFP if they have not yet published their event dates. So in this case I'm going to document this decision more clearly.

Appreciated!

bridgetkromhout:
I've thought some more, and I am going to merge this because it isn't going to list events that are missing a start date. This means we won't get the early warning (from this breaking) if someone announces a CFP date without a start date. This is a policy choice, but given that this is a rare case that will both confuse people and break the build, I agree with you that we should protect the site from bad data breaking the build.

PR #6608, opened 2019-03-15
[MVD-2019] CFP is Open, Sponsorship packages added, logo contest

andresantoniuk:
*Please title your pull request in this format: The event name and year in the title of the PR, along with a description of what is being changed, i.e., `[CHI-2019] Add Bluth Company as a sponsor`*

PR #6639, opened 2019-03-18
[MV 2019] no cfp dates without event dates

bridgetkromhout:
Hi, @andresantoniuk and @jantoniuk - we made an error by merging https://github.com/devopsdays/devopsdays-web/pull/6608 - we do not list CFP dates until the event dates are published. The reason for this is people cannot realistically determine whether they can submit a talk until they know when your event is.

We discovered this when @jakubjakubik worked on troubleshooting the issue I opened: https://github.com/devopsdays/devopsdays-web/issues/6625

Although the failure mode was a bit confusing, we do not want to display CFP dates when there is no event date. So, please add your event dates when they are confirmed, and then we will be happy to have your CFP dates added back to the site.

I hope that makes sense. Thanks!

### DIFF
--- a/data/events/2019-montevideo.yml
+++ b/data/events/2019-montevideo.yml
@@ -10,9 +10,9 @@ startdate:  # The start date of your event. Leave blank if you don't have a venu
 enddate:  # The end date of your event. Leave blank if you don't have a venue reserved yet.
 
 # Leave CFP dates blank if you don't know yet, or set all three at once.
-cfp_date_start: 2019-03-17T00:00:00-03:00 # start accepting talk proposals.
-cfp_date_end: 2019-06-01T23:59:59-03:00 # close your call for proposals.
-cfp_date_announce: 2019-06-17T00:00:00-03:00 # inform proposers of status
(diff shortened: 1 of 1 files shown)


andresantoniuk:
@bridgetkromhout I'm so sorry that our PR causes all this problems.

We understand the rationale of not publishing CFP before event date, but as date is almost confirmed we were encouraged by @MikeRosTX to start promoting CFP while we confirm venue.

I noted there were some problems with dates and made a brief note on  https://github.com/devopsdays/devopsdays-web/pull/6608/commits/c1b2d56a876af32319977254146eb32734c5d1ff

bridgetkromhout:
The problem we have seen happen before: CFP submissions come in. Then the event date is published. Then some of the CFP submitters have to change their minds (because the event date does not work well for them to speak). And if they don't tell the event until the CFP closes, then the event maybe thinks they have the talks they want, but it turns out, they do not. It can lead to a lot of extra work for the event.

I hope your date is confirmed soon! Thanks.

PR #6644, opened 2019-03-19
Bogotá-2019 Firts Speakers Added

devopscolombia:
Firts Speakers Added and also sponsor DevOps Institute
Thanks

PR #6650, opened 2019-03-19
[Bogota 2019] Fix URLs

bridgetkromhout:
Please only use ASCII, not unicode, in URLs. Fixing from https://github.com/devopsdays/devopsdays-web/pull/6644. /cc @MikeRosTX


MikeRosTX:
Good catch @bridgetkromhout! 👍

PR #6656, opened 2019-03-20
[PT2019] outsystems as gold sponsor

eduardopiairo:
outsystems as gold sponsor.


phrawzty:
Beyond adding the indicated sponsor, this commit appears to _delete_ `static/img/sponsors/edp.png` as well. Are you sure about that?

PR #6657, opened 2019-03-20
[ATL-2019] updating sponsors

cacorriere:
added diamanti and sysdig as gold, improving and eliassen as bronze, updated convox and electric cloud logos at request


MikeRosTX:
@cacorriere - Please rename the old logos being replaced with the 3 new ones being updated in this PR using the following format: **_<logo_name>yyyymmdd.jpg_** or **_png_**

bridgetkromhout:
It's the first time the convox logo is being used, and the "improving" image hasn't changed - they're just uploading a higher-res version. So electric cloud is the only one that needs changing.

PR #6659, opened 2019-03-20
[ATX-2019] Add embedded sched

scottsbaldwin:


MikeRosTX:
@scottsbaldwin - Was this what you were expecting to see? Because it doesn't look correct to me and I'm wondering if attendees will be confused.

image

scottsbaldwin:
@MikeRosTX , I have unlocked the sched event so that it is now public.

PR #6746, opened 2019-03-29
[LDN-2019] Replace Cloud Native London Logo

chriscoffee:
Cloud Native London have a better/new logo and Dom has asked for it to
be replaced.


bridgetkromhout:
Fine to merge without preserving the old one, as this is the first time it's going to be used.

PR #6755, opened 2019-03-30
[BOS-2019] Update event location and organizer info

adzuci:


MikeRosTX:
Spoke to @adzuci via Slack and explained that whenever organizers are added and removed, they need to send an email to info@devopsdays,org so we can remove them from Slack and email DL. They are removing 2 organizers in the PR, but Adam will send the core team an email confirming the 2 organizers to be removed.

adzuci:
Email sent!  I also added three other minor changes to the branch, which passed tests.  Tried to see if the GA logic goes into the Netlify previews, but didn't actually see our UA ID in there, so I may not have fully understood what's needed to get that to work.

PR #6777, opened 2019-04-02
[GHENT-2019] add mailchimp link

toshywoshy:
add a mailing list option through mailchimp


MikeRosTX:
It's OK to have the Email Subscription on the Welcome page.
But this is a personal preference, I would have placed it on the Registration place when it's time to buy tickets. Again just a personal preferences of mine, and maybe if the team would like to consider it.

PR #6815, opened 2019-04-05
[PDX-2019] Add a propose link to welcome page

BillWeiss:
I really wish I understood Hugo enough to do this right. Anyone who
wants to help me use the cfp_link content instead of duplicating it
here, that would be rad. :shrug:


bridgetkromhout:
You probably also want to uncomment "propose" and comment out the program and speakers: https://github.com/devopsdays/devopsdays-web/blob/master/data/events/2019-portland.yml#L38-L43

PR #6838, opened 2019-04-06
[CAMP-2019] Add estabilis and infoq as sponsors

Igorolivei:
Add estabilis and infoq as sponsors


Igorolivei:
@MikeRosTX now that we talked about the partnership with InfoQ and made everything clear, I reopened the PR.

PR #6840, opened 2019-04-07
Revert 8 master

xzilla:
/me leaving a trail of tears on the local repo, since apparently we are supposed to rebase now?


xzilla:
Hmm... github web is a magical plase. Was experimenting on the local DC branch, not sure how this ended up over here. I really do not want to merge this into upstream!

PR #6842, opened 2019-04-07
[RDU-2019] changing techtown logo by request from techtown aspe.

normalfaults:
[RDU-2019] changing techtown logo by request from techtown aspe.

PR #6843, opened 2019-04-07
Updating techtown to aspe

bridgetkromhout:
Update as requested in https://github.com/devopsdays/devopsdays-web/pull/6842.


normalfaults:
These changes work for us. thank you for all your help.

PR #6845, opened 2019-04-07
Finishing Sonatype update

bridgetkromhout:
Adding missing elements for update in https://github.com/devopsdays/devopsdays-web/pull/6835 - /cc @MikeRosTX


MikeRosTX:
Oops! Sorry.... Great catch! 👍

PR #6865, opened 2019-04-09
[DC 2019] add Red Hat Public Sector to platinum sponsors

robbkidd:
Added Red Hat's Public Sector group as a sponsor for DC.

Trimmed the campaign ID from their URL because this entry ought to be reusable for different events.


MikeRosTX:
@robbkidd - Is there something we should be waiting on from you before merging? WIP gives us that impression, that we should be waiting on something.

robbkidd:
@MikeRosTX Yes, this is intentionally WIP waiting for approvals from other DC organizers.

@nathenharvey @xzilla  👀❓

data/events/2019-washington-dc.yml
```diff
@@ -310,5 +310,7 @@ sponsors:
     level: platinum
   - id: excella
     level: platinum
+  - id: redhatgov
+    level: platinum
```
bridgetkromhout: You can override the URL right here if you want to include the campaign ID.

robbkidd: SHUT THE FRONT DOOR! Nice! I'll open another PR for that. Thanks!

PR #6883, opened 2019-04-11
[AMS-2019] Add Google as Gold sponsor

dpnl87:


bridgetkromhout:
Hi, @dpnl87 - you selected the 2011 logo for Google. I'm positive you shouldn't use that one - that one has their old font. I looked at the email thread and saw the present-day Google Cloud logo in the email signature of the person you were working with, so I'm going with that. If you have correspondence from Google asking for their 2011-era logo, please forward it to info@devopsdays.org and then raise a new PR to change it. Thanks!

PR #6934, opened 2019-04-15
[AMS-2019] Fix the Valtech logo

dpnl87:
Valtech contacted us about the logo not being the official version. As they are also a sponsor of the Kiev event, this will also change it on the Kiev page. This is expected.


kmugrage:
@dpnl87 it appears this would also change an event in the past, AMS 2017. This means that the event in the past has to be updated so that it doesn't change. That's done by creating a new sponsor called valtech-before-20190415, changing the old logo to match that name, modifying AMS 2017 to use the older version, and then using the new version on future events (2019 AMS, Kyiv, Stockholm). 

You can see https://github.com/devopsdays/devopsdays-web/pull/503 for more context and info, and of course let us know if you have questions / need help.

dpnl87:
@kmugrage It appears the logo has always been wrong. With them now being a sponsor for more events they started to notice. So with it updating the 2017 event as well ( AMS was their first sponsored event ) would be a good thing.

kmugrage:
I understand @dpnl87 - but we're not allowed to rewrite history. If you check out the original thread it was actually me as a sponsor wanting to change our old one but got overruled.

bridgetkromhout:
Screen Shot 2019-04-15 at 3 05 39 PM

@kmugrage is correct - we don't change history - we keep the time-period-appropriate past logos.

A google search shows that that past logo was in fact used historically. (In some places, like glassdoor, it's still being used to this day.) 

The good news is that it's similar enough to the new logo that it shouldn't lead to much confusion having it on an old page. We'll happily change it for all future events.

bridgetkromhout:
When the build finishes we should be able to resolve this. Thanks.

bridgetkromhout:
The update is complete (with a scaled-down image). Thanks!

dpnl87:
Thank you @bridgetkromhout and @kmugrage!

PR #6942, opened 2019-04-16
[Columbus 2019] New Organizers, Registration and CFP.

kjenkins19:
Need Bill Schwanitz slack reactivated
Need slack account created for Carmen Deardo
Opened both CFP and Registration.


bridgetkromhout:
Added both to Slack and email.

PR #6943, opened 2019-04-16
[Columbus 2019] Adding logos

bridgetkromhout:
Hi, @kjenkins19 - if you're planning to use the same logo as last year, this adds it to the main list of logos and to your welcome page.


bridgetkromhout:
@kjenkins19 you can see the proposed changes here: https://deploy-preview-6943--devopsdays-web.netlify.com/

Screen Shot 2019-04-16 at 8 54 13 AM

PR #6951, opened 2019-04-16
Revert "[HFD-2019] Set CFP announce date"

bridgetkromhout:
Reverts devopsdays/devopsdays-web#6945

"cfp_date_announce" means when an event plans to inform proposers of status, not when they told people the CFP was open (which is cfp_date_start).


jpadilla:
@bridgetkromhout thanks

PR #6955, opened 2019-04-17
[PAN-2019] Remove unnecessary info

kikicarbonell:


bridgetkromhout:
Good idea about google translate! It would be cool if we could get that built into the site.

PR #6961, opened 2019-04-17
[POZ-2019] Sponsor logo update, speaker added.

mazek:


bridgetkromhout:
Before merging this, since it has a sponsor logo update, we need to preserve sponsor logo history.

mazek:
Is there anything I should do to preserve logo history?

mazek:
I added dell previous dell logo to preserve history of it.

bridgetkromhout:
Ah, okay, so looking at this some more, I'm realizing that "dell" and "dell technologies" should be different sponsors. "dell" is a business unit under "dell technologies". This is much like "dell emc", which exists as a unit under "dell technologies"

So in this case I think we should make a new sponsor entry called "dell-technologies". Sponsors still might want "dell" or "dell emc". In this case they want "dell technologies".

bridgetkromhout:
The sponsor stuff looks good now (although since you're the first event using tomtom, you don't have to preserve old versions of it). However, since you added a new staff member, you now need to wait for a merge until someone has time to look and see if you sent an email with their email address, then add them to the email list and Slack. So, if you want this merged asap, remove the new staff member and put that in its own PR.

mazek:
Info about new organizer already sent to info@devopsdays.org. I will wait to merge it all in one piece:)

PR #6970, opened 2019-04-18
[Kiev-2019] update Zone3000 logo

delgod:


bridgetkromhout:
It's fine in this case to change the logo without preserving it, since this is the first time it's being used.

PR #6986, opened 2019-04-19
Update sharing image for fqdn

### LINKED ISSUE #6967: twitter sharing image not working
Twitter requires a FQDN; we had removed this for build reasons. Not quite sure how to solve for this. We could theoretically hardcode it to www.devopsdays.org (the problem with using baseurl is that it makes every build longer)

mattstratton:
fixes #6967


mattstratton:
Tested this with twitter and facebook debuggers. Seems good!

PR #7008, opened 2019-04-23
[Kiev-2019] minor updates

delgod:


phrawzty:
The appveyor failure is not related to the modifications made in the PR; however, your commit message states that Microsoft is a _travel_ sponsor when you've added them as Bronze. Is this correct?

delgod:
@phrawzty  yes, we add travel sponsors as Bronze
thanks for attention!

PR #7017, opened 2019-04-25
[DFW-2019] Adding Sponsorship+Welcome+sponsors

MikeRosTX:


MikeRosTX:
WIP - Waiting on PayPal links from @ConferenceOps, also have to add list of individual items.

data/events/2019-dallas.yml
```diff
@@ -204,6 +204,18 @@ sponsors:
 #  level: gold
 #  url: http://mysponsor.com/?campaign=me # Use this if you need to over-ride a sponsor URL.
   - id: arresteddevops
+    level: media
+  - id: devopslive
+    level: community
+  - id: addo-2019
```
bridgetkromhout: We don't namespace sponsors with a date.

data/events/2019-dallas.yml
```diff
@@ -2,7 +2,7 @@ name: "2019-dallas" # The name of the event. Four digit year with the city name
...
 event_twitter: "devopsdaysdfw" # Change this to the twitter handle for your event such as devopsdayschi or devopsdaysmsp
-description: "The same group of seasoned organizers that brought you DevOpsDays DFW several years in a row will be back again another year in 2019 planning what they believe will be an even more awesome event. They were encouraged and motivated by the feedback received from previous year's events, which is why they are determined to bring an even better experience this year and improve the DevOps conversation throughout the DFW Metroplex." # Edit this to suit your preferences
+description: <div style="font-weight:bold;color:#ffffff;text-align:left;vertical-align:top;vertical-align:middle">"<i>The same group of seasoned organizers that brought you DevOpsDays DFW several years in a row will be back again another year in 2019 planning what they believe will be an even more awesome event. They were encouraged and motivated by the feedback received from previous year's events, which is why they are determined to bring an even better experience this year and improve the DevOps conversation throughout the DFW Metroplex.</i>"</div> # Edit this to suit your preferences
```
bridgetkromhout: HTML in this field could break the site in the future; please don’t put it in the yaml.

data/events/2019-dallas.yml
MikeRosTX: Can we try once I'm done and if it breaks, I can quickly revert the changes?
Main reason I did it, is because it makes the test legible in white.

image

data/events/2019-dallas.yml
MikeRosTX: Can we try once I'm done and if it breaks, I can quickly revert the changes?
Main reason I did it, is because it makes the test legible in white.

image

data/events/2019-dallas.yml
MikeRosTX: Agreed. Since I do't do this very often, let me try the replacing logo thing we're suppose to do.

data/events/2019-dallas.yml
bridgetkromhout: It won't break right this second in this specific case, but there is a strong possibility that it will break with future updates, and it will almost certainly not work if slightly different, so we don't want a visible example of "what not to do".

mattstratton: I agree with Bridget. We will not merge this as is. Please revert it.

PR #7023, opened 2019-04-25
[SLC-2019] Add Fidelity as a sponsor

lindleywhite:
Add Fidelity as a sponsor and add the sponsor image and yml


MikeRosTX:
I would suggest renaming everything to **fidelitycareers**. Because someday Fidelity as a company may decide in the future to sponsor some other devopsdays event.

PR #7029, opened 2019-04-26
[Baku-2019] Added event in Baku

NasehBadalov:


phrawzty:
👋 Please send along the names and email addresses of the people listed as _Organizers_ to info@devopsdays.org . That way we can add those people to the mailing list and Slack. Once that's confirmed, we will review and merge this PR. Thanks!

MikeRosTX:
Everything looks great and we have all the emails for organizers, but we next to remove the date of the event until you have confirm from the venue location where the event will occur.

If venue location is confirmed, please the location to the navbar and in yml pluss the location.md.
If you do not have a venue location, please remove the date from the yml file so it can be listed under TBD.

PR #7071, opened 2019-04-30
[ZRH-2019] Add Applause as Sponsor

Roderick-Jonsson:
We got a new sponsor :)


data/events/2019-zurich.yml
```diff
@@ -220,6 +222,8 @@ sponsor_levels: # In this section, list the level of sponsorships and the label
   label: Bronze
 - id: community
   label: Community
+- id: applause
```
phrawzty: This item is probably a mistake. You've effectively created a new sponsorship tier called `applause`. Suggest removing this line and the one below it. 😄

Roderick-Jonsson: Sorry didn't notice that. Thanks for the change 👍

PR #7094, opened 2019-05-03
[Pdx-2019] registration is up, cfp is down

BillWeiss:
I'm doing this via a comical stack of connectivity and hardware, so I haven't seen the render. I'll look at the netlify output and leave a +1 or -1 to say if I'm happy with it. :/


BillWeiss:
+1

PR #7095, opened 2019-05-03
[Pdx-2019] sponsor adds

BillWeiss:
Like my last commit, I need to see Netlify work for this. Cloudbolt gave us a special image, and told Bridget that it was just for this event, so I forked their sponsor entry.


BillWeiss:
Bad file upload, -1 . I'll fix it when I'm at a real computer.

BillWeiss:
+1 , figured out my file.

PR #7106, opened 2019-05-03
[LDN-2019] Use capitalone.co.uk rather than .com

chriscoffee:
A request from CapitalOne to use the UK version rather than the US
version of the site.

This requires there to be another file with a different ID and another
logo.

Signed-off-by: Chris M <me@christophermills.co.uk>


bridgetkromhout:
The logo looks the same to me, and it's possible to override the URL without adding a new sponsor, but if you say the logo is different according to them, perhaps there's something subtle. If it turns out it does have the same image, we can collapse this complexity. Thanks.

chriscoffee:
@bridgetkromhout yeah it's the same! I didn't know it was possible, though I didn't really look that hard or give any explanation to what I tried, Apologies!

How do you use the id of one sponsor but override the URL? Would love to learn! 😄

bridgetkromhout:
Answered elsewhere (sorry for the delay) but for posterity the answer lies here: https://github.com/devopsdays/devopsdays-web/blob/master/themes/devopsdays-theme/REFERENCE.md#sponsor-fields

PR #7117, opened 2019-05-05
[STH-2019] add f5 and iZettle as sponsors

krakan:
also EA/Dice were mislabeled as silver sponsors while they want to be community sponsors


phrawzty:
Hi @krakan: Community sponsorships are generally for small associations, not-for-profits, and other such organisations. EA/Dice is, well, not that. Are you sure about that status change?

krakan:
Hello @phrawzty,
Yes, we'll discuss this in the org team, but for now their status should be "community".

data/events/2019-stockholm.yml
```diff
@@ -108,9 +108,13 @@ sponsors:
   - id: b3init
     level: silver
   - id: ea
-    level: silver
+    level: community
```
phrawzty:
```suggestion
    level: silver
```

krakan:
Yes, please do that until we get the potential bronze level sorted out.

PR #7205, opened 2019-05-15
[PT-2019] add Rhys Evans interview to speaker page

MiguelAlho:
We have added the first of the speaker interviews to help promote them and the event and are using the speakers page to add this content (instead of a specific landing page). no CSS changes have been made for this (though that was considered a possibility in the website slack channel discussion)

### DIFF
--- a/content/events/2019-portugal/speakers/rhys-evans.md
+++ b/content/events/2019-portugal/speakers/rhys-evans.md
@@ -6,8 +6,37 @@ type = "speaker"
 linktitle = "rhys-evans"
 +++
 
-#### Speaker
+### Speaker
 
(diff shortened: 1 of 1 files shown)


bridgetkromhout:
Sounds good. I checked out https://deploy-preview-7205--devopsdays-web.netlify.com/events/2019-portugal/speakers/rhys-evans/ and it does show the interview content (albeit in the left column, so it does show up as somewhat narrow).

MiguelAlho:
thanks @bridgetkromhout ! I think that, with the <hr>s , there is a clearer divsion of content. and with the limited time we have, exploring other options might set us back a bit. But thanks again for the info you sent on the slask channel. I want to explore that further for future events.

PR #7296, opened 2019-05-26
[SYD-2019] kickoff

m1keil:
Adding sydney 2019 event to the list. :fireworks:

### DIFF
--- a/content/events/2019-sydney/conduct.md
+++ b/content/events/2019-sydney/conduct.md
@@ -1,0 +1,35 @@
++++
+Title = "Conduct"
+Type = "event"
+Description = "Code of conduct for devopsdays sydney 2019"
++++
+
+## ANTI-HARASSMENT POLICY
+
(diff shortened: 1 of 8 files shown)


bridgetkromhout:
While the UTC +2 time was definitely wrong, I think the +10 may also be wrong for your event dates: https://www.timeanddate.com/time/zone/australia/sydney - by that time in October, will it be +11?

m1keil:
good catch @bridgetkromhout :-)

bridgetkromhout:
@m1keil can you email info@devopsdays.org with your organizer team's names and email addresses? We need that info to set up the email list and Slack. Thanks.

m1keil:
done!

could you please also add devopsdays-sydney-2019-organisers@googlegroups.com
as a target to the email alias?

Cheers

bridgetkromhout:
> could you please also add devopsdays-sydney-2019-organisers@googlegroups.com
as a target to the email alias?

As you can imagine, this ends up being way too complex in terms of one-offs, so we don't do them. But you are of course welcome to have your own mail filters set any way you like.

bridgetkromhout:
Fixed a few other things - when the build finishes this should be good to merge.

bridgetkromhout:
Hmm, the propose link 404s. Either fix it or remove it?

m1keil:
@bridgetkromhout I was hoping link will be active by today but we need some financials to figure out. commented out for now. Thanks for the review.

bridgetkromhout:
Great! I’ll merge this now and your page will be up. You will likely want to add an event image and possibly organizer images, when you get a chance.

m1keil:
Yep, going to do tweaks during the week. thanks a lot

PR #7312, opened 2019-05-28
[PT-2019] bosch as bronze sponsor

eduardopiairo:
* bosch as bronze sponsor


bridgetkromhout:
Let's name this `bosch-pt` as this is country-specific.

bridgetkromhout:
Ah, since the image isn't country-specific, I've set a URL override that is. We can merge post-build.

PR #7314, opened 2019-05-29
[MVD-2019] Add sponsors and change redhat icon site-wide

andresantoniuk:
we are adding redhat as sponsor of montevideo-2019; as logo changed following https://github.com/devopsdays/devopsdays-web/pull/503 we updated redhat to redhat-before-20190528 in previous events


data/events/2019-boise.yml
```diff
@@ -91,7 +91,7 @@ sponsors:
     level: silver
   - id: microsoft
     level: silver
-  - id: redhat
+  - id: redhat-before-20190528
```
bridgetkromhout: devopsdays Boise 2019 is after May 28th, 2019 and thus should say `redhat`, not `redhat-before-20190528`.

data/events/2019-minneapolis.yml
```diff
@@ -132,7 +132,7 @@ sponsors:
     level: community
   - id: signalfx
     level: platinum
-  - id: redhat
+  - id: redhat-before-20190528
```
bridgetkromhout: devopsdays Minneapolis 2019 is after May 28th, 2019 and thus should say `redhat`, not `redhat-before-20190528`.

bridgetkromhout:
Because you have a protected master branch and I cannot push changes to it, I will mark this "do not merge" until you make the corrections I mentioned - thanks.

data/events/2019-boise.yml
andresantoniuk: @bridgetkromhout it's a good point, but the fact is that when Boise team added RedHat as sponsor they decided to use the previous available logo so if we change this we'll "rewrite history".

Actually they included Redhat as sponsor in https://github.com/devopsdays/devopsdays-web/commit/4b23b8a04581032c4de420c6db9bff539efae24d#diff-ba2e1bcc3f60b941342a653255d2a641 (Apr 2)

What do you suggest to do?

data/events/2019-minneapolis.yml
andresantoniuk: same as Boise; we can revert to "redhat" in this case, but we'll rewrite originally intended logo from this organizers.

Please let us know how to proceed.

data/events/2019-boise.yml
bridgetkromhout: As discussed on Slack, sponsor logo changes take effective for all future events as of the logo change date.

data/events/2019-minneapolis.yml
bridgetkromhout: As discussed on Slack, sponsor logo changes take effective for all future events as of the logo change date. Thanks!

andresantoniuk: done.

andresantoniuk:
References to Redhat logo already fixed

PR #7332, opened 2019-05-30
[PT-2019] man digital hub as silver sponsor

eduardopiairo:
* man digital hub as silver sponsor


eduardopiairo:
It was a test that was wrongly committed. Already reset to the previous size.

PR #7394, opened 2019-06-07
[DC-2019] add STS as a gold sponsor

robbkidd:
👋 @nathenharvey @xzilla


robbkidd:
🚧 Need to add a Twitter handle to the sponsor data.

PR #7429, opened 2019-06-13
[NYC-2020] - adding nyc 2020

jaydestro:
Initial integration for New York City 2020 event.


mattstratton:
Pending email alias creation :)

MikeRosTX:
Pending on @jaydestro to send a list of NYC organizers to the core team.

bridgetkromhout:
I didn't see an email, but since it's the same as last year, minus three people, I just set it up like last year, removing those people.

PR #7431, opened 2019-06-13
[AMS2019] Add Indigo as wine sponsor

yvovandoorn:
There will be wine at a BBQ, add the sponsor.


MikeRosTX:
@yvovandoorn, when the **_indigo.yml_** in this PR is same as old one. In reality this PR is to replace Indigo's old logo with a new one. Since this is the case, please reference the following on replacing logos with new ones:  https://github.com/devopsdays/devopsdays-web/pull/503#issuecomment-223454015

yvovandoorn:
@MikeRosTX addressed the issue as they are different indigo's and different URLs (and updated company name).

PR #7454, opened 2019-06-17
[PRG-2020] Initial event info

a-vorobiev:
Initial info about Devopsdays Prague 2020 organizers.


MikeRosTX:
@a-vorobiev, we need you to send an email to infor@devopsdays.org and include a list of all the local organizers for Prague that has their full name and their email before we merge this PR.

PR #7468, opened 2019-06-19
Fix grammar for becoming a sponsor.

don-code:
Boston recently added an a-la-carte sponsorship type. The framework duly creates a grammatically incorrect "Become a a la carte sponsor" link on our page.

This change makes vowels form grammatically correct sentences. The solution is not perfect - it will screw up on things like "Become an one-day sponsor," but is generally more often correct than not.

Before:
Screenshot from 2019-06-19 01-41-44

After:
Screenshot from 2019-06-19 01-41-55


MikeRosTX:
Waiting on @mattstratton's approval.

PR #7470, opened 2019-06-19
[GAL-2019] Add Aspect as sponsor.

scanepa:
* New silver sponsor
* New sponsor logo


electrofelix:
Ah, I'm thinking that we shouldn't have changed the old logo file... sorry

PR #7478, opened 2019-06-20
[Columbus-2019] Updating sponsor to a different level, adding Organizer

kjenkins19:
Adding Jack Maher as an organizer. Ping on Slack to get his email to sign him up for Slack.


kmugrage:
@kjenkins19 please send his full name and email address to info@devopsdays.org so we can add him to Slack and the mailing lists. We try to avoid passing info on Slack since it's temporary.

PR #7483, opened 2019-06-20
[DFW-2019] Adding Verizon as a sponsor

MikeRosTX:


MikeRosTX:
Waiting on firm confirmation from ConferenceOps that payment was received.

PR #7519, opened 2019-06-25
[STH-2019] remove diabol as sponsor

mattgruter:
Diabol has pulled out as a sponsor for DevOpsDays Stockholm 2019.


phrawzty:
Confirm that this sponsor was invoked for _this_ event only and thus can be removed.

PR #7521, opened 2019-06-25
[CHI-2019] activecampaign and better logos

margaretvaltie:
added active campaign as lanyard sponsors. 2 newer betterer logos for zoro and bcbs


bridgetkromhout:
Okay to change the logos without preserving old ones since this is the first use of each.

PR #7529, opened 2019-06-26
[SM-19] Cancellation DevOpsDays Santa Maria 2019

tonybolzan:
It is with regret that we inform you that the DevOpsDays Santa Maria 2019 is being canceled.

After an internal deliberation of the organization members, we are sorry to say that we have decided not to hold the event this year. Although we have received many promises of support and partnership, few have been fulfilled and it is not financially feasible to maintain the event. Moreover, due to all the effort involved and the small team of organizers in the project we are not able to accommodate all needs to carry out the event in a satisfactory manner.


bridgetkromhout:
Sorry to hear this - we hope to get your city's event back in the future!

PR #7530, opened 2019-06-27
[LDN-2019] Add 2019 Speakers

chriscoffee:
We've added the speakers for 2019 to the website. This does not include the program. 

Images are in png format and haven't been ran through an optimiser, hoping that gulp is still running the optimisation.

If not I can do this...


chriscoffee:
Please don't merge this for the time being, I thought we were putting ignites live, we're not.

chriscoffee:
Right, that's fixed now, feel free to merge. Thanks

PR #7559, opened 2019-07-01
[DC-2019] add Comcast as Silver sponsor

robbkidd:
OK to reuse existing logo and website URL.


robbkidd:
@xzilla @nathenharvey 👀

PR #7562, opened 2019-07-01
[GAL-2019] Embedded ticket widget tweak and register interest form

electrofelix:
Adding a form that can handle registering user and email via ti.to interested list without leaving
the website.


electrofelix:
@mattstratton is this approach safe to use?

mattstratton:
Seems to work and is acceptable. Bear in mind that this is *unsupported* and may break without notice (i.e. we can't/won't regression test this in changes we make to the site/hugo version)

electrofelix:
@mattstratton understood and perfectly reasonable

PR #7618, opened 2019-07-09
[Edi-2019] New organisers

mjslane:
Please could we add the following new organisers to the slack and organiser lists

cristiano.balducci@gmail.com
leonardo.t@gmail.com


MikeRosTX:
Please send an email to info@devopsdays.org with  the new organizer's full name and their email addresses. Thanks!

bridgetkromhout:
A bit of detective work and I think Leonardo Trotti is the name we're missing.

PR #7627, opened 2019-07-10
[FOR-2019] Add Sponsor AWS and Hostweb

gb8may:


data/sponsors/aws.yml
```diff
@@ -0,0 +1,2 @@
+name: "aws"
+url: "https://www.amazon.com.br/"
```
phrawzty: There is already an established AWS sponsor. If you want to use a different URL, you can specify it using an over-ride for your event. See [the docs](https://github.com/devopsdays/devopsdays-web/blob/master/utilities/README.md#sponsor-url-override) for more details.

p.s. That's not the AWS URL in any case.

phrawzty: Concerning the image; please don't over-write sponsor images. If you want to use a different image, you can do that, but you need to set up a new sponsor YAML. If the sponsor has requested their logo be updated specifically, then you need to back-date the previous sponsor profile. This has [already happened](data/sponsors/aws-before-20180313.yml) for AWS; you can grep for the string `aws-before-20180313` to see how it was handled.

If the Brazilian version of AWS is actually different from AWS in the rest of the world, you should set up a sponsor just for that. I don't believe this to be the case, however.

I suspect that simply using the existing AWS sponsor profile is what you want to do.

gb8may: I don´t found an AWS img on directory, but if there´s one we can use it.. no problem. How can i proceed?

phrawzty: The logo: `/static/img/sponsors/aws.png`

It's used by `data/sponsors/aws.yaml`

You can just put "aws" as a sponsor and it will work. :)

gb8may: Ow.. so.. i´ll do it! Thanks and sorry about that!

PR #7650, opened 2019-07-12
[MSP-2019] Add microsoft as silver sponsor

DebbieGillespie:


bridgetkromhout:
Subtle difference, but I added them as "microsoftazure" instead. We may change it (I emailed them to discuss) but often the cloud unit is what gets represented.

PR #7660, opened 2019-07-13
[BEN-2019] Add the 2019 event

vjdhama:


bridgetkromhout:
I've set up the email list for this year and made sure the correct people are on Slack. I'll add the team pictures after merging, as you link them but don't include them.

PR #7692, opened 2019-07-19
[MVD2019] Add new sponsor and update sponsor logo

andresantoniuk:


bridgetkromhout:
Okay to update logo - it's only used this time thus far.

PR #7708, opened 2019-07-22
[BUF-2019] Adding Verizon Media as Sponsor

aspleenic:
- Adding a bronze Sponsor: Verizon Media
- Updating Stark & Wayne sponsor logo


data/events/2019-buffalo.yml
```diff
@@ -90,6 +90,9 @@ sponsors:
   - id: bcbswny
     level: gold
     url: https://www.bcbswny.com/content/wny/careers.html
+  - id: verizonmedia
+    level: bronze
+    url: https://www.verizonmedia.com
```
bridgetkromhout: Fwiw, you don't need to put the URL here unless you're overriding the one in the sponsor data file.

bridgetkromhout:
Since Stark & Wayne is used here for the first time, logo changes don't require preserving old images.

bridgetkromhout:
Since Stark & Wayne is used here for the first time, logo changes don't require preserving old images.

PR #7721, opened 2019-07-23
Bogota-2020 initial information from DevOpsDays 2020-bogota

devopscolombia:
Added information to publish the event to TBD

Thanks


bridgetkromhout:
I have emailed to explain that we need the new organizers' email addresses so we can properly configure Slack and mailing list before we merge this.

PR #7761, opened 2019-07-27
[RDU-2019] added speakers

normalfaults:
[RDU-2019] added speakers


bridgetkromhout:
sanket-godebole.png is a typo but we can fix it.

PR #7763, opened 2019-07-28
[TLV-2019] new organizer added

shar1z:
I will send an email as well - thanks!!


bridgetkromhout:
Added to email and slack too.

PR #7787, opened 2019-07-31
[PRG-2020] Event updates.

a-vorobiev:
Dates, venue, sponsor information and social media links has been added.


bridgetkromhout:
We should have ensured https://github.com/devopsdays/devopsdays-assets/pull/71 was merged, as this PR references it - it's merged now.

a-vorobiev:
> We should have ensured devopsdays/devopsdays-assets#71 was merged

Actually, not. It is obviously better to have them merged both, but we would survive a couple of days with 404 on the PDF. This one was really important, that's why I've asked only for that. Anyway, thanks for merging the assets also.

PR #7842, opened 2019-08-07
[SHA-2019] Add 2019-Shanghai event

overloadzy:
*Please title your pull request in this format: The event name and year in the title of the PR, along with a description of what is being changed, i.e., `[CHI-2019] Add Bluth Company as a sponsor`*

PR #7865, opened 2019-08-10
[Shanghai 2019] Adding images for Shanghai 2019

bridgetkromhout:
Since we have images for the existing organizers I'm adding them.


bridgetkromhout:
These images are in support of https://github.com/devopsdays/devopsdays-web/pull/7842.

PR #7950, opened 2019-08-20
Combine SEO templates

mattstratton:
This combines all of the SEO templates into one, which allows for a single data query instead of doing it twice (and does a little code simplification).

This cut about 30 seconds off of the hugo build on my local machine.


mattstratton:
Tested this with Facebook debugger, etc, and it all is happy.

PR #7953, opened 2019-08-20
[KC-2019] Publish initial schedule

bksteiny:
We'll slot speakers as we get their acceptances and we come up with the schedule


mattstratton:
I like how you added a legend! Nice!

PR #7955, opened 2019-08-21
[EDI 2019] talks

MichalBryxi:
- First wave of speakers & sponsors
 - Enabled both program & speakers sections

cc: @kevinhcross looks good?

### DIFF
--- a/content/events/2019-edinburgh/program.md
+++ b/content/events/2019-edinburgh/program.md
@@ -1,0 +1,14 @@
++++
+Title = "Program"
+Type = "program"
+Description = "Program for devopsdays Edinburgh 2019"
+Icons = "false"
++++
+
+<div class = "row">
+  <div class = "col">
+    <hr />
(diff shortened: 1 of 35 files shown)


kevinhcross:
@MichalBryxi I had a quick look and I think there is more info we can add at this stage. A few of the speakers profiles only have first names but we actually have full names and bios for all the speakers. I will send you the location of that info on another channel.

MichalBryxi:
@kevinhcross I think now it has all the possible data we have? Anything missing?

bridgetkromhout:
You need the speakers.md and program.md files for those links to work. Here are the templates:

https://github.com/devopsdays/devopsdays-web/blob/master/utilities/examples/templates/program.md

https://github.com/devopsdays/devopsdays-web/blob/master/utilities/examples/templates/speakers.md

MichalBryxi:
Sorry @bridgetkromhout. Added, pushed.

PR #7961, opened 2019-08-22
[CAI-2019] Fixing location and updating organizers

aameen79:


aameen79:
@phrawzty @MikeRosTX @bridgetkromhout 
Update is needed urget
Thanks for support :)

aameen79:
@phrawzty @MikeRosTX @bridgetkromhout 
Update is needed urget
Thanks for support :)

PR #7963, opened 2019-08-22
[Columbus-2019] Sponsor updates.

kjenkins19:
App Dynamics gold
VMWare gold
Ahead silver
Checkpoint bronze
CoverMyMeds requested the URL update.

### DIFF
--- a/data/sponsors/2015-covermymeds.yml
+++ b/data/sponsors/2015-covermymeds.yml
@@ -1,2 +1,2 @@
 name: "CoverMyMeds"
-url: "https://www.scriptscribe.org/"
+url: "https://www.covermymeds.com/main/careers"
--- a/data/events/2019-columbus.yml
+++ b/data/events/2019-columbus.yml
@@ -118,6 +118,10 @@ sponsors:
     level: gold
   - id: gitlab
     level: gold
+  - id: appdynamics
+    level: gold
+  - id: vmware
+    level: gold
   - id: tasktop
     level: silver
   - id: upstart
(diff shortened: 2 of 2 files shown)


data/sponsors/2015-covermymeds.yml
```diff
@@ -1,2 +1,2 @@
 name: "CoverMyMeds"
-url: "https://www.scriptscribe.org/"
+url: "https://www.covermymeds.com/main/careers"
```
phrawzty: As a general rule, we don't allow modifications to historical sponsor data. If this is a one-time change (i.e. specific to _this_ event), you can use the `url:` feature in the sponsor list ([example](https://github.com/kjenkins19/devopsdays-web/blob/0cf3021ca311338e6fb9179dd8e591e30bb89579/data/events/2019-washington-dc.yml#L309-L311)).

If this is a permanent change, please use the `before` pattern, of which there are many prior examples:
```
devopsdays-web (master) » find . -name '*before*' | wc -l
     204
```

If you need help with this, just let us know either here or on Slack, and we'd be happy to assist. :)

kjenkins19: While this would normally be true, I'd like to point out a few things:
* CMM has only ever sponsors our conference, so I'm not worried about side affects of other conferences
* https://www.scriptscribe.org/ is no longer valid
* https://www.scriptscribe.org/ was already an unusual spot to be forwarded (it was some kind of blog). I believe their careers place is a more permanent location, and would be their preference for anyone clicking on a logo for prior years as well.

As such, I would like to request this change be made as is, rather than the before change. If you don't find those good enough, I can do the before system.

mattstratton: URL's changing in a case like this are fine. This can be merged.

PR #7966, opened 2019-08-22
[SLC-2020] Initialize next year's conference

Clait:
Copy 2019 files and change dates and info for setting up our 2020 event.


bridgetkromhout:
I've updated email and Slack to match the contents of this PR.

Clait:
Thanks Bridget - we have new members for the board incoming. Will post
another PR in the coming weeks for that.

bridgetkromhout:
That's fine. We need an email with their names/email addresses at the same time as the PR.

Clait:
> That's fine. We need an email with their names/email addresses at the same time as the PR.

sounds good

PR #7973, opened 2019-08-22
[Columbus-2019] Schedule, Speakers, and everything to go with it.

kjenkins19:
Currently known schedule.
Speakers. There a couple missing images, which should be fine for now.
Updated menu to remove propose, add in speakers and schedule


mattstratton:
Couple comments:

For the speakers with missing images, if you remove the `image = "ed-legault.jpg"` (for example) from the frontmatter on the speaker page for that person, it will show the default image. It will look better than a broken image.

The image for Marisa Grayson is too small - it will be enlarged by the build process and the quality will probably be pretty degraded. Your call on that one.

PR #7985, opened 2019-08-24
[CHI-2019] Update Preed's talk info+bio

preed:


preed:
Paging @mattstratton, @mattstratton to a white courtesy telephone.

PR #7993, opened 2019-08-25
[2020] Updating SLC email scheme

bridgetkromhout:
The old email aliases will all still work, but we're standardizing on the city name so you won't need to get a new alias created each year.  @Clait please pull in from upstream/master before next time you update - thanks!


Clait:
Thanks, will do.

PR #8006, opened 2019-08-25
[HOU-2020] - Added location, date, and updated organizer info

madblkman:


data/events/2020-houston.yml
```diff
@@ -10,8 +10,8 @@ ga_tracking_id: "" # If you have your own Google Analytics tracking ID, enter it
 #   variable: 2019-01-05T23:59:59+02:00
 # Note: we allow 2020-MM-DD for backward compatibility, but it can lead to unexpected behaviors (like your event disappearing from the front page during your last day)
 
-startdate:  # The start date of your event. Leave blank if you don't have a venue reserved yet.
-enddate:  # The end date of your event. Leave blank if you don't have a venue reserved yet.
+startdate:  2020-04-14T00:00:00-05:00 # The start date of your event. Leave blank if you don't have a venue reserved yet.
+enddate:  2020-04-15T00:00:00-05:00 # The end date of your event. Leave blank if you don't have a venue reserved yet.
```
bridgetkromhout: Is this a one-day event? If so, you want to list the end date as `2020-04-14T23:59:59-05:00` because you don't want two dates listed. If it's a two-day event, then you will want to list it as `2020-04-15T23:59:59-05:00` because you don't want it to disappear from the front page at the start of the second day.

bridgetkromhout:
We'll need to resolve that date ambiguity before merging - thanks.

data/events/2020-houston.yml
madblkman: Ah, gotcha! I'll make that correction real quick. Thanks!

Issue #6542, opened 2019-03-11
PDF on talk page can run into sponsor row


mattstratton:
I'm not 100% sure if this ever worked (I didn't write the code for this) but on the shown example, the PDF viewer bumps into the sponsor row, and it is not usable.

![image](https://user-images.githubusercontent.com/1280666/44929568-53b12380-ad10-11e8-9f19-f1fa2f144bf1.png)

bridgetkromhout:
This is fixed in https://github.com/devopsdays/devopsdays-web/pull/8076 for cases of a simple-to-embed PDF. We have a corner case, it appears, if the PDF is in a github repo - github appears to block such an embed (see https://github.com/github/markup#github-markup).

bridgetkromhout:
test case: content/events/2018-chicago/program/quintessence-anx.md

-slides = "https://github.com/quintessence/presentations/blob/master/sensory-friendly-monitoring/sensory-friendly-monitoring.pdf"
+pdf = "https://github.com/quintessence/presentations/blob/82dab688031a5f6c338b15ca5b2aa9efa26febf3/sensory-friendly-monitoring/sensory-friendly-monitoring.pdf"

mattstratton:
If github blocks embedding, I don’t see how we can code around it. This seems like a documentation update.

PR #8076, opened 2019-08-29
PDF fix test (with test PDF re-added).

bridgetkromhout:
Possible partial fix for https://github.com/devopsdays/devopsdays-web/issues/6542. Minimally tested but seems to help in this case.

### DIFF
--- a/content/events/2017-minneapolis/program/kristen-womack.md
+++ b/content/events/2017-minneapolis/program/kristen-womack.md
@@ -5,7 +5,7 @@ Talk_end_time = ""
 Title = "Logs & Alerts, Not Just for Ops"
 Type = "talk"
 Speakers = ["kristen-womack"]
-pdf = "https://kristenwomack.io/DevOps-Minneapolis-Ignite-2017"
+pdf = "https://static1.squarespace.com/static/5389534ee4b037d53f461731/t/59790b69e58c62191bb7337e/1501105003965/DevOps-Ignite-Talk-2017.pdf"
 youtube = "hJZEwvoaymI"
 +++
 
--- a/themes/devopsdays-theme/layouts/talk/single.html
+++ b/themes/devopsdays-theme/layouts/talk/single.html
@@ -79,7 +79,7 @@
         {{- if ne .Params.pdf "" -}}
         <div class="row">
           <div class = "col" style="max-width:100%">
-            <iframe src="{{.Params.pdf}}" width="100%" height="200%">
+            <iframe src="{{.Params.pdf}}" width="100%" height="100%">
               This browser does not support PDFs. Please download the PDF to view it: <a href="{{.Params.pdf}}">Download PDF</a>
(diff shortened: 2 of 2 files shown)


bridgetkromhout:
Improved: doesn't run into sponsors
Question: can we make it have a better experience for a pop-out-to-see-fully link? https://deploy-preview-8076--devopsdays-web.netlify.com/events/2017-minneapolis/program/kristen-womack/

mattstratton:
Hmm. if it doesn't show the whole slide without scrolling, I don't see the value in embedding the PDF instead of just using the "slides" type link, which provides a link to the slides instead (see this example):

image

mattstratton:
> Question: can we make it have a better experience for a pop-out-to-see-fully link?

We could definitely put a "link to PDF" hyperlink above it for sure...

mattstratton:
I suspect that the true fix for this is going to be doing some CSS magic - we'll have to maybe add a class for a div for PDF embeds, and then we can reference that class in the code. 

I think this fix is okay for now (it makes existing ones work okay) but I would open a new issue to make it look more ideal (and to include the "link to the pdf" link)

mattstratton:
I found a slightly better fix:

```
           <div>
              <object data="{{.Params.pdf}}" type="application/pdf" width="300" height="200">
              alt : <a href="{{.Params.pdf}}">{{ .Title }}</a>
              </object>
            </div> 
```

The width and height need to be tweaked I think though.

mattstratton:
My fix (above) looks like this:

image

mattstratton:
Making it 600x400 looks pretty good:

image

mattstratton:
The hyperlink to the PDF isn't necessary, because if you hover over it in this fix, it gives you the ability to download, etc:

image

bridgetkromhout:
Okay, cool. Can you add changes to this PR?

mattstratton:
Ah crap. It looks stupid on a smaller desktop (looks fine on mobile though)

image

mattstratton:
I'll add the changes to the PR so you can play with them in the preview though (can always back them out)

mattstratton:
Try out your test URL with my fix and see how it looks on your screen, etc

https://deploy-preview-8076--devopsdays-web.netlify.com/events/2017-minneapolis/program/kristen-womack/

bridgetkromhout:
At scaled-down sizes (smaller than 600x400) it doesn't look right (missing the action links). However, given that it looks fine on my 12" MacBook in a full-sized Chrome window at the default font size, I think this is an incremental improvement.

mattstratton:
Given that nobody has brought this up but you and me, Bridget, I think that means we can merge this as fixed :)

:shipit:

bridgetkromhout:
Wellll. I am unconvinced it works for the github PDF link of Q's (see https://github.blog/2015-03-17-pdf-viewing/) but that's a corner case.

PR #8080, opened 2019-08-29
[CHA-2019] Updates TVA's logo

asachs01:
Per TVA's usage guidelines here https://www.tva.com/Newsroom/TVA-Logos, we have to update their logo.


bridgetkromhout:
I understand the need to match their guidelines. For some reason, this blue looks wildly different from the one I see at that link (whether eps or jpg). 

Theirs is on the left, and yours is on the right - are you positive this is correct?

Screen Shot 2019-08-29 at 4 43 12 PM

asachs01:
Hmmmm... That was pulled from their supplied logos. Maybe some weirdness
with what I used to resize.

bridgetkromhout:
Closing in favor of https://github.com/devopsdays/devopsdays-web/pull/8085.

PR #8085, opened 2019-08-30
Correcting sponsor logo

bridgetkromhout:
Per https://github.com/devopsdays/devopsdays-web/pull/8080 but I'm keeping the correct color.

This is not a logo "replacement" or "update" - we're correcting an error in which one to use.

/cc @asachs01


asachs01:
Ty! I'm just now getting stuff taken care of for the night and was about to hop on and tackle these.

PR #8092, opened 2019-08-30
[KC-2019] Fix broken image for Roger Noah

mattstratton:
There is no image for Roger Noah. This change replaces the broken image with the default speaker image.

@aaronblythe please check this out.


bksteiny:
Thanks for the fix 👍

aaronblythe:
Thank you.

PR #8110, opened 2019-08-31
[PDX-2019] Hashicorp co-sponsor for the event

BillWeiss:
Friends, let me know if you'd like me to rename this. It's certainly got a URL for just our event.


mattstratton:
@bridgetkromhout what do you think?

bridgetkromhout:
I think that while we don't usually add an `and` in these cases, it doesn't matter too much (and I'll refactor it later if it does). As for the year-specific blog post, that's fine in this case because there really isn't a canonical "these two together, disassociated from a year/event". 

There's a slight corner case of "but what if they partner on a different event later" but if this happens a lot, they'll probably get a non-event, non-year landing page, so I'm not going to worry right now.

PR #8185, opened 2019-09-07
Add Event and Dates to the Slack unfurl

### LINKED ISSUE #8189: Add more meta tags to improve slack unfurl
Improvements suggested by @adrianmoisey 

https://medium.com/slack-developer-blog/everything-you-ever-wanted-to-know-about-unfurling-but-were-afraid-to-ask-or-how-to-make-your-e64b4bb9254

First related PR: https://github.com/devopsdays/devopsdays-web/pull/8185

adrianmoisey:
https://medium.com/slack-developer-blog/everything-you-ever-wanted-to-know-about-unfurling-but-were-afraid-to-ask-or-how-to-make-your-e64b4bb9254

Fixes #8189

### DIFF
--- a/themes/devopsdays-theme/layouts/partials/head/seo.html
+++ b/themes/devopsdays-theme/layouts/partials/head/seo.html
@@ -20,7 +20,7 @@
   {{- if $e.event_twitter -}}
     <meta name="twitter:creator" content="@{{$e.event_twitter}}" />
   {{- end -}}
-  {{- $.Scratch.Set "event" (printf "devopsdays %s %s " $e.city (chomp $e.year)) -}}
+  {{- $.Scratch.Set "event" (printf "devopsdays %s %s" $e.city (chomp $e.year)) -}}
   <meta name="twitter:label1" value="Event" />
   <meta name="twitter:data1" value="{{ $.Scratch.Get "event" }}" />
   {{- if $e.startdate -}}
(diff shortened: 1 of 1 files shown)


adrianmoisey:
Before:

image


After:

image

adrianmoisey:
Testing each page type:

Welcome:
image

Speakers:
image

Program:
image

Talk:
image

adrianmoisey:
I'm happy that it works as intended. 
It seems to do nothing on twitter (the medium article explains why, and seems to br correct).

Happy to receive review/thoughts/comments/ideas!

bridgetkromhout:
LGTM - let's have @mattstratton review too!

mattstratton:
LGTM. This is great. I was bikeshedding over the label being called "Event" vs something else, but I don't have a better idea.

One thing - maybe make the title ("Speakers", etc) include the event name, so it would be "Speakers - DevOpsDays Ponyville 2020") to give a bit more context?

mattstratton:
I don't know if we can control that though - it might pull directly from the HTML for the page, and I *think* we don't want to modify the title on the page.

mattstratton:
OK, so thinking about this some more, this is more a thing where we want to fix the `<title>` for pages to be more descriptive in general. That is out of scope for this change (it's also weird, because we don't want to automatically add the event name and year, because some pages have that in their title already). So really it's just a thing that people should do better in general, not something we should fix in code.

adrianmoisey:
I agree with you. I think it would be nice to have more info in the title, but for the reasons you describe might be difficult.

mattstratton:
This looks good to me, so I'm going to merge it

mattstratton:
Oof. This isn't great. Now every page in the site changes on every production build, so they are all new files. The master builds have doubled in time since adding this. :/

mattstratton:
Disregard. That was a red herring. This change is fine!

adrianmoisey:
> Disregard. That was a red herring. This change is fine!

*phew*

PR #8187, opened 2019-09-07
[DET-2019] added speakers and program

archieco:
[DET-2019] added speakers and program


archieco:
@bridgetkromhout, we really appreciate those edits and getting this posted! Thanks.

Issue #8194, opened 2019-09-08
investigate why some logos aren't loading for old years


bridgetkromhout:
Example: https://devopsdays.org/events/2015-minneapolis/welcome/

The devopsdays in the upper corner, plus the sponsor logos, don't load, but the city image does.

<img width="760" alt="Screen Shot 2019-09-08 at 4 42 08 AM" src="https://user-images.githubusercontent.com/2104453/64482492-2173b500-d1f3-11e9-87be-3cc639d322eb.png">

PR #8203, opened 2019-09-08
Fix sponsor logos on archived events

adrianmoisey:
Remove `srcset`, as the @2x images don't seem to exist.

This should fix https://github.com/devopsdays/devopsdays-web/issues/8194

From my investigation I think this is the correct solution.


adrianmoisey:
https://deploy-preview-8203--devopsdays-web.netlify.com/events/2015-minneapolis/welcome/

mattstratton:
That gets you halfway there. A lot of the old legacy events are now hosted at https://github.com/devopsdays/devopsdays-legacy so you need to do the same "search and replace" on that repo too :)

mattstratton:
https://github.com/devopsdays/devopsdays-web/blob/master/static/_redirects#L1-L34 for more context

adrianmoisey:
> so you need to do the same "search and replace" on that repo too :)

There aren't any matches for `srcset` on that repo, the image URLs all look good to me.

mattstratton:
Fair. I guess the legacy site was updated :)

bridgetkromhout:
Super awesome, @adrianmoisey! Nice to get those legacy sites fixed.

PR #8209, opened 2019-09-08
Migrate from bower to npm

### LINKED ISSUE #8205: refactor css/js processing
Currently, the CSS files are built using SCSS, and the Javascript files are concatenated.

The process for generating these is dependent on the non-free, OS X only tool Codekit. 

We should move from Codekit to using `gulp` as that is free and multi-platform. There are several issues that need to be resolved as part of this though:

- Right now we are using Bower for the package management to pull in Bootstrap and some jquery utilities. Bower is deprecated (and has been for a while). We need to migrate this to npm.
- A gulp task needs to be written that will compile the CSS and concatenate the JS
- Some of the packages we use are in Bower only, and npm packages don't exist. For those, we will need to copy their files directly into the theme and not pull them in from npm
- The SCSS file imports need to be updated to pull from `node_modules` and not `bower_components`
- We should update the versions on Bootstrap (among others) as part of this anyway

mattstratton:
This change migrates from the use of bower (deprecated) to npm for dependencies.

We also update Bootstrap to 4.3

Several packages do not exist in NPM so they are copied into the static directory for the theme

Fixes #8205 


mattstratton:
Note that this does not remove the codekit dependency, but it's a start.

mattstratton:
Latest commit adds the `gulp dev` task which compiles the SCSS and concats/minifies the JS

mattstratton:
I just did a cursory check on the latest deploy preview, and all javascript I am aware of seems to still work.

bridgetkromhout:
> Fixes #8205

Doesn't this only fix a small part of that issue? Might want to avoid an accidental auto-close.

mattstratton:
Nope. I added a commit that added the gulp script so it’s all fixed :)

Issue #6538, opened 2017-07-25
Talk and Speaker pages should support sharing images


mattstratton:
Right now, even if you set a sharing image, it won't be used.

PR #8201, opened 2019-09-08
[Work in progress] Speaker sharing images

adrianmoisey:
This is mostly so I can test with a public URL.

PR #8212, opened 2019-09-08
Social sharing images

adrianmoisey:
Allow a user to set `og:image` on any page

This PR replaces https://github.com/devopsdays/devopsdays-web/pull/8201
(for a cleaner git history)

Fixes https://github.com/devopsdays/devopsdays-web/issues/6538


themes/devopsdays-theme/layouts/partials/head/seo.html
```diff
@@ -23,6 +23,10 @@
   {{- if (where (readDir "static/events") "Name" $e.name) -}}
     {{- if (where (readDir ($.Scratch.Get "contentdir")) "Name" "sharing.jpg") -}}
       {{- $.Scratch.Set "sharing_url" ((printf "https://www.devopsdays.org/events/%s/sharing.jpg" $e.name ) ) -}}
+    {{- else if .Params.sharing_image -}}
+      {{- with .Params.sharing_image -}}
+        {{- $.Scratch.Set "sharing_url" (printf "/events/%s/speakers/%s" $e.name .) -}}
```
bridgetkromhout: I'm a little confused because it sounded like the scope was expanding to "any page can use this" but then we're suggesting storing it in the `speakers` dir. Maybe we want a directory `/events/%s/sharing/%s` instead?

mattstratton: oh, good catch. I agree with @bridgetkromhout that it should go in a `sharing` directory in the event's static. @adrianmoisey 	you'll need to update this code and also the docs you created.

adrianmoisey: 👍 Good catch! Will fix.

adrianmoisey: Fixed in https://github.com/devopsdays/devopsdays-web/pull/8212/commits/7dd8a98992cf06dbf83e5de494467a22af821aed

PR #8225, opened 2019-09-09
Replace non https link in static pages

mattstratton:
A bunch of the rendered static pages have the insecure link to the feed; this is creating a mix-content issue


adrianmoisey:
While you're there, does it not make sense to also strip the `www` from the URLs?

mattstratton:
> While you're there, does it not make sense to also strip the www from the URLs?

I don't think it makes a big difference; it's just a link (and the link in the standard footer is `https://www...` so this at least matches the rest of the site :)

Issue #8192, opened 2019-09-08
improve organizer guide


bridgetkromhout:
- I don't recall why I didn't build a table of contents and jump-to-section anchors for https://devopsdays.org/organizing but I think both are useful and we should add them.

- Take a pass through the guide and update as needed.

bridgetkromhout:
Thanks again, @adrianmoisey, for the table of contents! I'm closing this out as I've also taken another pass through and updated the docs a bit.

PR #8244, opened 2019-09-11
[Work in progress] First pass at adding a ToC to the organizing page

adrianmoisey:
Add a table of contents, as per https://github.com/devopsdays/devopsdays-web/issues/8192

I'm not sure if this is the correct solution, but I thought I'd take a try.
I'm also unsure if my bootstrap classes are right.

Please test on mobile and desktop and see what you think.


bridgetkromhout:
Super-cool! On desktop it's a bit farther over to the right that I was imagining (I'd be fine with it in the middle column) but this is a great start.

adrianmoisey:
Here are a few options I've been playing with. I'm not 100% happy with any of them, but I think one of them may be a good base to start with.

## Option 1: Moved to between the left panel and the contents:
image

## Option 2: Changed the look and feel of the ToC (using https://gist.github.com/percygrunwald/043e577beb90db72e09727a3ed3053c3):
image

## Option 3: Move the ToC to above the contents, and put it in a collapsable box:
Collapsed:
image

Uncollapsed:
image

bridgetkromhout:
Let's see what @mattstratton thinks; I don't particularly mind replacing the event-list sidebar for this page.

mattstratton:
Replacing the event list sidebar might be a little tricky; I like option 3 the best.

bridgetkromhout:
Sounds good. Let's get option 3 merged in!

adrianmoisey:
> Replacing the event list sidebar might be a little tricky; I like option 3 the best.

I can make it on the event sidebar (I may be doing nasty hacks which you don't approve of though... but let's cross that bridge when we get there)...

## Here's what it looks like with the standard Hugo ToC:
image

## Or with the custom one (option 2 from above):
image


Given that it is an option, what are your thoughts?
Still the original Option 3, or ToC on the left?
For what it's worth, I wasn't a massive fan of the original ideas I had come up with, and my vote is the second screenshot in this comment.

mattstratton:
I vote for the second screenshot in the above comment (replace event sidebar with the fancier ToC for the guide).

The "hack" isn't a bad hack and I approve it.

bridgetkromhout:
My only concern at this point is numbered sections. We're highly likely to rewrite/refactor. I don't want links to "item 16" to exist and then stop working. The only anchors I want are section headers.

mattstratton:
I concur. If the numbering is automatically generated by markdown that’s fine. Let’s just not hardcore it.

adrianmoisey:
I've pushed the latest version (ToC on the left, replacing upcoming events).

Please play around on the preview and see what you think.

adrianmoisey:
Whoops! Accidentally deleted the branch.

mattstratton:
This looks good to me. I'm going to merge it and we can continue to tweak.

@bridgetkromhout 	just fyi - the sections are numbered, but they are generated dynamically and not hard-coded, so I think it's fine.

adrianmoisey:
Indeed, it's just an HTML ordered list.
If we want it to be a bullet point list, it's easy enough to change the partial to an unordered list.

PR #8298, opened 2019-09-16
[Cha 2019] fixing filename

bridgetkromhout:
Hi, @asachs01 & @MikeRosTX - we don't use special characters in filenames to prevent a variety of tedious issues. I've fixed this from the recent edit in https://github.com/devopsdays/devopsdays-web/pull/8293 - please be sure to pull this change in before you add this speaker to the program. Thanks.


MikeRosTX:
Thanks @bridgetkromhout! Missed the image file naming.

PR #8313, opened 2019-09-17
[Columbus-2019] Adding images for two speakers.

kjenkins19:
Adding two images. If they are too large, let me know, and I can try and shrink them.


bridgetkromhout:
Can you please rename kristen-biddulph.jpeg to kristen-biddulph.jpg, since some of the processing expects that - and also please shrink that image if possible. Preview on a Mac will let you do so. 600px by 600px is plenty big.

bridgetkromhout:
Given the event starts tomorrow I’ll merge now, clear up issues later.

PR #8289, opened 2019-09-16
Use a single sharing directory

adrianmoisey:
Fixes https://github.com/devopsdays/devopsdays-web/issues/8223

PR #8316, opened 2019-09-17
Updating per new convention

bridgetkromhout:
Updating docs to match change from https://github.com/devopsdays/devopsdays-web/pull/8289.


adrianmoisey:
:+1:

Thanks for catching that

PR #8320, opened 2019-09-18
[REC-2019] Add C4P and new organizers 

yagonobre:


MikeRosTX:
Set to **_do not merge_** until an email is sent to the core team with a list of all the new Recife organizer's full name and email.

PR #8322, opened 2019-09-18
[OSL-2019] Updated logo

larstobi:
We have a new logo and want to use the new logo on the web page.


bridgetkromhout:
If you intend for this to be used on the front page you also need to put it here: https://github.com/devopsdays/devopsdays-web/blob/master/static/events/2019-oslo/logo-square.jpg

PR #8326, opened 2019-09-18
Testing moving MSP 2017 to static

bridgetkromhout:
Hey, @mattstratton - is this the right model for archiving a year?

- Leaving the data file alone
- generating the public directory: `hugo --theme=devopsdays-theme --buildDrafts=false --baseURL="https://www.devopsdays.org/"`
- copying the generated files to `static`
- removing the `content` .md files


mattstratton:
That sounds right to me :)

bridgetkromhout:
> 12:58:34 PM: Build exceeded maximum allowed runtime

Because it's trying to build _everything_, I think.

bridgetkromhout:
I'm going to try merging this. the blast radius is relatively limited, and I can revert if it doesn't work as expected.

mattstratton:
Agreed.

PR #8336, opened 2019-09-19
[EDI - 2019] -  new sponsor

MichalBryxi:


bridgetkromhout:
Please also add an image for a new sponsor - thanks.

PR #8354, opened 2019-09-20
[GHENT-2019] Launch logo & program

yvovandoorn:
* Reveal logo for 2019 Ghent
* Launch program
* Add missing talks
* Cleanup welcome page / remove mailing list
* Go!


bridgetkromhout:
Hey, @yvovandoorn - are you sure you want to put Ken on the schedule twice with the same talk? https://deploy-preview-8354--devopsdays-web.netlify.com/events/2019-ghent/program for reference.

Also, I try to livetweet everyone, but I usually have to decompress immediately after I speak - if that matters, you might want to swap my timeslot with Jabe's - just a thought!

bridgetkromhout:
Ah, the logo-square file needs to actually be square or it will get stretched to be square when it's on the front page after it has gone through the build process (this won't show up in the preview). I recommend pasting the image onto a blank or white square of the correct size.

yvovandoorn:
@bridgetkromhout thanks for catching the double Ken talk and non square logo. I've also swapped you with Jabe!

jedi4ever:
@yvovandoorn not sure there is a type ignite? or did they not get listed here

yvovandoorn:
@jedi4ever the type ignite doesn't allow for specific times (only something for program/talks). So I had to compromise on using the 'talk' type with a specific background color.

PR #8368, opened 2019-09-22
Organizing update

kmugrage:
Adding budgeting reminder discouraging extravagant organizer expenses.

### DIFF
--- a/content/page/organizing.md
+++ b/content/page/organizing.md
@@ -142,7 +142,7 @@ Examples of reasonable expenses for organizers:
 Examples of unreasonable expenses for organizers:
 
 - Gifts not given to attendees and / or speakers
-- Extravagant food and beverage for meetings. 
+- Extravagant food and beverage for meetings
 
 ---
 
--- a/content/page/organizing.md
+++ b/content/page/organizing.md
@@ -118,7 +118,7 @@ Expenses:
 
 - Venue for talks & open space discussions
 - Internet for attendees
-- Livestream/recording of talks
+- Livestream/recording/captioning of talks
 - Catering for breakfast/coffee breaks/lunches (some venues require the use of their caterer)
 - Evening event
 - T-shirts/other swag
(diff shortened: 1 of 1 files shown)


content/page/organizing.md
```diff
@@ -129,6 +129,22 @@ Expenses:
...
+- Economy travel / lodging to and from other DevOpsDays events for the purpose of making yours better
+- Local transportation to from the event or planning meetings
+- Clothing such as shirts to make the team stand out during the event
+- **Reasonable** food and beverage for meetings before and after the event
+
+Incomplete list of examples of unreasonable expenses for organizers:
+
+- Gifts not given to attendees and / or speakers
+- Extravagant food and beverage for meetings
+- Business / First class travel
```
iennae: I think this is problematic. We don't know what folks might need. Personally many economy flights are not great for larger folks so putting this in as an unreasonable expense is not awesome.

content/page/organizing.md
```diff
@@ -129,6 +129,22 @@ Expenses:
...
+- Economy travel / lodging to and from your event
+- Economy travel / lodging to and from other DevOpsDays events for the purpose of making yours better
+- Local transportation to from the event or planning meetings
+- Clothing such as shirts to make the team stand out during the event
+- **Reasonable** food and beverage for meetings before and after the event
+
+Incomplete list of examples of unreasonable expenses for organizers:
+
+- Gifts not given to attendees and / or speakers
+- Extravagant food and beverage for meetings
```
iennae: What would extravagant mean here?

content/page/organizing.md
```diff
@@ -129,6 +129,22 @@ Expenses:
...
+
+Incomplete list of examples of reasonable expenses for organizers:
+
+- Economy travel / lodging to and from your event
+- Economy travel / lodging to and from other DevOpsDays events for the purpose of making yours better
+- Local transportation to from the event or planning meetings
+- Clothing such as shirts to make the team stand out during the event
+- **Reasonable** food and beverage for meetings before and after the event
+
+Incomplete list of examples of unreasonable expenses for organizers:
```
MikeRosTX: Shouldn't there be an item under **_unreasonable expenses for organizer_** that reads something like this:

- Economy travel / lodging to and from other for-profit vendor events

Because I remember a few years ago when several organizers wanted to go to DOES **_for the purpose of making their better_** because that year they were planning on having a Leadership Summit, and several core members raised concerns objecting to paying for organizer's expenses to attend a for-profit event. For-profit events to name a few DOES, AWS re:Invent, Microsoft Ignite, ChefConf, Puppetize, etc.

bridgetkromhout: Deciding which non-devopsdays events are community-focused is out of scope.

content/page/organizing.md
bridgetkromhout: Reworded.

content/page/organizing.md
bridgetkromhout: Reworded to make it clear we're expecting expenses to be typical for a region.

content/page/organizing.md
bridgetkromhout: Reworded to make it clear we expect event travel to be devopsdays-specific.

content/page/organizing.md
```diff
@@ -129,6 +129,24 @@ Expenses:
 - Insurance/taxes
 - Payment for accounting services
 
+**Important Reminder:** As you've seen in the rules above, these events are not intended for individual or corporate profit. This includes unnecessary high personal expenditures for the benefit of organizers. We can't give precise numbers as to what qualifies as large amounts, since that would vary widely. Consider typical appropriate pricing for your region, and determine if you would find it to be unusually expensive.
+
+Reasonable expenses for organizers:
+
+- Travel & lodging to participate in your event
+- Travel & lodging to participate in other DevOpsDays events for the purpose of making yours better
```
phrawzty: Notice how it's `DevOpsDays` here and then `devopsdays` on the next line. As a core team, we should probably just decide how we want to spell it on our own damn website. :)

PR #8381, opened 2019-09-23
Clarifying ticket costs

bridgetkromhout:
As pointed out by @phrawzty, currencies and local prices vary, so we want to be less dollar/euro-centric about typical costs.

### DIFF
--- a/content/page/organizing.md
+++ b/content/page/organizing.md
@@ -344,7 +344,7 @@ We do not recommend making your event free. Experience has taught us that 'free'
 
 ### Minimal Cost
 
-Asking a minimal fee (covering about food cost for two days - about 200USD or 200 Euro) gives us the best of both worlds:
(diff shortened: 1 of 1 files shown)


content/page/organizing.md
```diff
@@ -344,7 +344,7 @@ We do not recommend making your event free. Experience has taught us that 'free'
 
 ### Minimal Cost
 
-Asking a minimal fee (covering about food cost for two days - about 200USD or 200 Euro) gives us the best of both worlds:
+We recommend charging a minimal fee. This will keep your event accessible by making it cost a fraction of a typical high-priced conference ticket (so, perhaps 10%-20% of what someone might pay to attend a commercial tech event in your region). This gives us the best of both worlds:
```
phrawzty:
```suggestion
We recommend charging a minimal fee. This will keep your event accessible by making it cost a fraction of a typical high-priced conference ticket (perhaps 10% to 20% of what someone might pay to attend a commercial tech event in your region). This gives us the best of both worlds:
```

PR #8395, opened 2019-09-24
Rename/rearrange emeritus core team listing

bridgetkromhout:
Two issues:
- The way we've previously listed inactive past members of the core team looks clunky, confusingly listing them on the contact pages of every event, when they aren't associated with the events or familiar to the newer city teams. 
- The name "Historic" also was a bit odd/confusing. 

Solution:
Going with the more traditional "Emeritus" and changing it such that they only appear on the "About" page.


bridgetkromhout:
I'm going to handle this in a more elegant way.

PR #8431, opened 2019-09-28
Organizer refactor

bridgetkromhout:
Problem:
- Some organizers aren't directly working with new events, but haven't entirely retired. Adding a new level "Advisory", to ensure organizers can be listed accurately.
- The way we've previously listed inactive past members of the core team looks clunky, confusingly listing them on the contact pages of every event, when they aren't associated with the events or familiar to the newer city teams.
- The name "Historic" also is a bit odd/confusing.

Solution:
- Changing "Historic" to the more clear "Emeritus".
- Separating out core team members who are just advising core, not actively working on website or onboarding events, and calling them "Advisory".
- Listing Advisory and Emeritus core members on the "About" page but not individual event pages.

Moving @geekle to be the first "Advisory" member per discussion with him. Let's re-evaluate to make sure people are listed at the level where they're participating.

### DIFF
--- a/content/page/about.md
+++ b/content/page/about.md
@@ -22,8 +22,7 @@ The devopsdays global core team guides local organizers in hosting their own dev
 
 **Core Organizers**
 
-<em>Active</em><br>
-{{< list_core >}}
+{{< list_core_active >}}
 
 {{< list_core_advisory >}}
 
--- a/content/page/about.md
+++ b/content/page/about.md
@@ -22,6 +22,7 @@ The devopsdays global core team guides local organizers in hosting their own dev
 
 **Core Organizers**
 
+<em>Active</em><br>
 {{< list_core >}}
 
 {{< list_core_advisory >}}
--- a/content/page/about.md
+++ b/content/page/about.md
@@ -18,7 +18,7 @@ The first devopsdays was held in Ghent, Belgium in 2009. Since then, devopsdays 
 
 
 ### About the organization
-The devopsdays global core team guides local organizers in hosting their own devopsdays events worldwide.
+The devopsdays global core team guides local organizers in hosting their own devopsdays events worldwide. Active core organizers onboard and guide events, answer questions, and maintain the website. Advisory core organizers are less involved day-to-day but weigh in on important matters and assist as needed. Emeritus core organizers are no longer involved in the core team; we thank them for their past efforts.
 
 **Core Organizers**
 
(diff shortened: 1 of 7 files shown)


bridgetkromhout:
You can see the updates on the About page - [new](https://deploy-preview-8431--devopsdays-web.netlify.com/about) vs [old](https://devopsdays.org/about) and on a city contact page - [new](https://deploy-preview-8431--devopsdays-web.netlify.com/events/2019-chicago/contact) vs [old](https://devopsdays.org/events/2019-chicago/contact).

mattstratton:
Hmm. Only thing is i understand we aren’t listing all advisory and emeritus folks on each city contact page, but having “active” at the top of that list seems a little awkward. Maybe we remove that from the short code and have “core organizing group” also be a link to About?

mattstratton:
I am 100% for the new levels btw.

bridgetkromhout:
> having “active” at the top of that list seems a little awkward. Maybe we remove that from the short code 

Yep. I went back and forth on that, and I'm happy to take it out of the shortcode and only list it in the one page where we want it; I've pushed a PR showing that. Do we want the label taken out of _all_ the shortcodes, then, is the question? (Note that currently, the Advisory and Emeritus groups only appear on the About page.)

> and have “core organizing group” also be a link to About?

This part is not trivial. "The core devopsdays organizer group" is unfortunately hardcoded into zillions of past event files, so I don't _really_ want to mess with it. Maybe an enhancement for the future?

mattstratton:
> "The core devopsdays organizer group" is unfortunately hardcoded into zillions of past event files

Well there you go. I rescind my request on that one then :)

bridgetkromhout:
> Well there you go. I rescind my request on that one then :)

I don't hate the idea of change - I just don't want to have to retcon the past too much. Basically I think it would be easiest if we can have this produce what we want, since it's what we have to work with:
```
**The core devopsdays organizer group**

{{< list_core >}}
```

So I would be fine with us changing `list_core` to a short explanation of the core team including a link to `/about`, instead of a bunch of names, if that's what people want.

mattstratton:
I’m not arguing to have the core names removed from the contact pages btw. I just thought adding a link would be nice.

bridgetkromhout:
> I just thought adding a link would be nice.

I agree. Plus the more I think about it, the more a giant list of names at the bottom of people's city contact pages is just weird and doesn't scale. I pushed a revision to correct this. If we remove `**The core devopsdays organizer group**` from future event templates it will still work well.

mattstratton:
👍

bridgetkromhout:
Okay, we can keep iterating but this seems solid.

PR #8451, opened 2019-10-01
[Work in progress] Add anchor links to headers

adrianmoisey:
Taken from https://discourse.gohugo.io/t/adding-anchor-next-to-headers/1726/9?u=kaushalmodi

I'm looking for some help from @mattstratton here...

I'm trying to allow people to direct link to a heading. For example if a sponsor has a question, we can direct them to a specific header. We already have the ID in the heading, but we need the heading to be a link.
I seem to have gotten half way there, but the way I'm doing it is swallowing the heading of the page.

### DIFF
--- a/themes/devopsdays-theme/layouts/_default/baseof.html
+++ b/themes/devopsdays-theme/layouts/_default/baseof.html
@@ -25,7 +25,7 @@
                </div>
 	    {{- else -}}
               <div class="col-md-8 order-md-12">
-                  {{- block "main" . }} {{- end -}}
+                  {{- partial "heading_link.html" .Content -}}
               </div>
               <div class="col-md-2 order-md-1">
                   <a href = "{{ "/events" }}" class="left-nav-navs">ALL EVENTS</a><br />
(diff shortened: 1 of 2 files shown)


MikeRosTX:
Well it's not the most cosmetically nice looking, but it does the job. 👍 
Would have prefer the invisible anchor to the left of the title that appears when hovered over it with the mouse.

adrianmoisey:
> Well it's not the most cosmetically nice looking, but it does the job. 👍

Yup! I figured we can start with the basics, can make it look pretty later on.

mattstratton:
I’ll dig into this a little later when I get some time on keyboard - I’m sure we can puzzle it out!

mattstratton:
Yeah, you need to understand how we use base templates and blocks...I would recommend reading up on them a little bit and you'll see why you can't remove the stuff you removed

https://gohugo.io/templates/base/

bridgetkromhout:
If this isn't going to merge in this weekend, can we get an issue documenting what we've found and then close this specific PR? Thanks!

adrianmoisey:
Done: https://github.com/devopsdays/devopsdays-web/issues/8494

adrianmoisey:
I'll try pick this up again when I have more time available.

PR #8471, opened 2019-10-03
[2019-TLV] Opened Registration and Added Speakers

shar1z:
I managed to get Hugo running locally - so it seems that everything is in order to merge.  Thanks!


data/events/2019-tel-aviv.yml
```diff
@@ -180,7 +183,7 @@ sponsors:
   - id: devopsloft
     level: community
 
-sponsors_accepted : "yes" # Whether you want "Become a XXX Sponsor!" link
+sponsors_accepted : "" # Whether you want "Become a XXX Sponsor!" link
```
mattstratton: Is the intent to close sponsorship? As it happens, i don’t think this part of the code works anyway (the “join as XXX sponsor” links still show up anyway)

mattstratton:
@shar1z - this should be okay to merge, but i had one question first (see above). If that’s okay, or even if you want us to merge and look into that later, just let us know in the PR here and we will merge right away!

data/events/2019-tel-aviv.yml
shar1z: Yes I technically want to close sponsorships - as we're sold out.  I don't mind leaving it "yes" if it doesn't work anyway.  I'd like to remove the sponsor links though.  I'll try and do that now. 

Thanks!

bridgetkromhout:
@shar1z you can remove all sponsorship links if desired - I wasn't sure if that's what you were trying to do?

PR #8480, opened 2019-10-03
[Kraków-2020] initial content

kgadek:


bridgetkromhout:
Hi! Excited to get Kraków listed on the website! I've set up your Slack and email, as you had most of the email addresses available. A few issues:

- We don't have an email address for Jeroen Fossaert so I can't add them - please email info@devopsdays.org when you have their email, and submit a PR to re-add them.

- Also, I removed the dates because this PR did not include an address. We require a specific confirmed location for the event to be listed if you're going to publish dates.

Welcome!

PR #8491, opened 2019-10-05
[Baku-2019] Changed pages

NasehBadalov:


bridgetkromhout:
This doesn't look ready to go. Also, we aren't going to link to a non-https site, as it causes build warnings.

Screen Shot 2019-10-05 at 9 51 15 AM

In general, we discourage new cities from running their own sites, as our SEO is much better and their time is better spent on their event, not on site problems.

PR #8510, opened 2019-10-09
[EDI - 2019] - priyanka rao

MichalBryxi:


MikeRosTX:
Picture for Riyanka Rao is 400x600px. Have a feeling it might not be showing up because it **_needs to be square_** according to the guidelines, but there's a possibility that is not the reason and it might be another reason.

MikeRosTX:
Disregard previous message, it all looks good. In the future, make sure they are square. Max 900x900px.

MichalBryxi:
Will do. Thanks for pointing that out.

PR #8332, opened 2019-09-18
Archiving 2017 events

bridgetkromhout:
Using this method:
- Leaving the data file alone
- generating the public directory: hugo --theme=devopsdays-theme --buildDrafts=false --baseURL="https://www.devopsdays.org/"
- copying the generated files to static
- removing the content .md files

PR #8526, opened 2019-10-11
Properly move old Skyscanner logo to 'before' standard

yvovandoorn:
In #8512, a new Skyscanner logo was added but the new one was added as `skyscanner-2019` instead of renaming the old one (and updating event files that pointed to the historic logo). 

This PR implements the best practice.


yvovandoorn:
With the archiving of 2017 & 2018 in #8332, I had to go back and rewrite history for the events using Skyscanner.

PR #8531, opened 2019-10-11
update README to reflect that static files may need to be updated

yvovandoorn:
Be explicit that static files may need to be updated.

### DIFF
--- a/utilities/README.md
+++ b/utilities/README.md
@@ -92,7 +92,7 @@ grep -r "sponsors\/pivotal.png" static/events/*
 [...]
 static/events/2016-chicago/contact/index.html:<img src="/img/sponsors/pivotal.png" alt="Pivotal" title="Pivotal" class="img-fluid">
 ```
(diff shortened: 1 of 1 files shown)


utilities/README.md
```diff
@@ -85,6 +85,15 @@ data/events/2016-atlanta.yml:  - id: pivotal-before-20190307
...
 ```
 
+It is possible that the previous city has already been archived and static items have already been created. You will have to update the image source for these pages
+
+```
+grep -r "sponsors\/pivotal.png" static/events/*
+[...]
+static/events/2016-chicago/contact/index.html:<img src="/img/sponsors/pivotal.png" alt="Pivotal" title="Pivotal" class="img-fluid">
+```
+:warning: The above will require you to be comfortable with search & replace utilities as you'd potentially be modifying hundreds of files.
```
kmugrage: Should we add something here about asking for help on Slack if they need it?

PR #8583, opened 2019-10-19
[EDI - 2019] postevent page

MichalBryxi:


bridgetkromhout:
We can certainly merge this; if you want the slides and/or videos to be accessible from the individual pages, I recommend looking at https://github.com/devopsdays/devopsdays-web/blob/master/themes/devopsdays-theme/REFERENCE.md#talk-page-fields - and you can see an example on a page like https://devopsdays.org/events/2019-minneapolis/program/liz-fong-jones.

PR #8627, opened 2019-10-24
[CC-2020] Changed contact email

curratore:


bridgetkromhout:
We aren't using that format anymore. The one generated by the template is correct and does not need to be changed.

PR #8644, opened 2019-10-29
[LDN-2020] Add page for 2020

chriscoffee:
We should be doing another year in 2020, given we've already booked the
event lets do this!

Add logo, masthead, welcome stuff

Signed-off-by: Chris M <millscj01@gmail.com>


data/events/2020-london.yml
```diff
@@ -0,0 +1,131 @@
...
+    level: community
+  - id: tech-notts
+    level: community
+  - id: london-devops
+    level: community
+  - id: cloud-native-london
+    level: community
+  - id: docker-london
+    level: community
+
```
iennae: I take it you want to leave all the community sponsors as is? Just checking about these.

chriscoffee: Yup! It might look like a copy pasta but they're intentional. Thanks for checking 👍They're projects run by organizers or that organizers are apart of and ones that we generally attend 😄

PR #8662, opened 2019-11-01
[GAL-2019] Remove speaker

electrofelix:
Remove talk from program of speaker unable to travel


mattstratton:
fyi, if you have a speaker image for them, you might want to delete it as well as sometimes the build acts weird locally if there is an image for a speaker that doesn't exist :)

electrofelix:
Done, thanks, had assumed it would just be ignored

mattstratton:
> Done, thanks, had assumed it would just be ignored

A reasonable assumption, but @crayzeigh discovered that weird edge case. It won't affect the site being published, or the netlify builds, but it can make odd things happen when you build locally :)

PR #8669, opened 2019-11-02
[VIX-2020] Add CFP link in welcome page

sprt-balarini:
Add CFP link in welcome page


bridgetkromhout:
I suggest also adding the link https://www.papercall.io/devopsdaysvix20/ to https://github.com/devopsdays/devopsdays-web/blob/master/data/events/2020-vitoria.yml#L21 (which can be the link for a CFP button and the propose menu item) and setting your CFP's open/close/announce dates in https://github.com/devopsdays/devopsdays-web/blob/master/data/events/2020-vitoria.yml#L17-L19 (which makes your CFP appear on https://devopsdays.org/speaking).

PR #8680, opened 2019-11-05
[GAL-2019] Added Microsoft as sponsor

scanepa:


electrofelix:
Please don't merge yet, just spotted that potentially it is `Microsoft Azure` instead of `Microsoft`

scanepa:
Closing as of error in resource selection should be Microsoft Azure

PR #8682, opened 2019-11-05
[BER-2019] Cleanup program, add speakers pics

range:


range:
Resized all speakers' pictures we have right now.

PR #8695, opened 2019-11-07
[BER-2019] Add links to social event. Remove Rafal from program

range:


bridgetkromhout:
This only removes the listing from the schedule, but does not remove him from the speaker page. If he is not speaking, please also remove the picture, program file, and speaker file - thanks.

range:
Check. Done.

PR #8702, opened 2019-11-09
[VIX-2020] add cfp, add link telegram, remove sponsor

sprt-balarini:
add cfp in navbar.
    add link telegram group.
    remove sponsor.


bridgetkromhout:
Thanks for the updates! I've removed that organizer from Slack and the email list since you're taking them off the organizer list.

PR #8710, opened 2019-11-09
Removing redundant header

bridgetkromhout:
In https://github.com/devopsdays/devopsdays-web/pull/8431 I refactored how contact pages for individual cities were displayed. I removed the list of core team members in favor of a link to the `/about` page. Given that, the header of `**The core devopsdays organizer group**` is redundant since the following line is just a link, not a list of people.

### DIFF
--- a/content/events/2019-berlin/contact.md
+++ b/content/events/2019-berlin/contact.md
@@ -10,6 +10,5 @@ If you'd like to contact us by email: {{< email_organizers >}}
 
 {{< list_organizers >}}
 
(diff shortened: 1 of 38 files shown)


bridgetkromhout:
For the record, I didn't want to retcon the past, but this is what I used for 2020: `$ sed -i '' "/\*\*The core devopsdays organizer group\*\*/d" content/events/2020-*/contact.md` and I individually selected the still-to-come 2019 events for the same treatment.

PR #8723, opened 2019-11-12
Only match ^/cityname, and nothing else

adrianmoisey:


adrianmoisey:
Sed needs to be updated to only match the first occurrence of this too.

adrianmoisey:
Closing this until I can figure it out.

adrianmoisey:
I think I found a way to do it.

PR #8744, opened 2019-11-15
[PT-2020] update team members details; call for sponsors deadline

eduardopiairo:
* update team members details; 
* added call for sponsors deadline


bridgetkromhout:
@eduardopiairo the problem is that pictures need to be square or they will not look right.

eduardopiairo:
@bridgetkromhout @MikeRosTX Uploaded the resized pic. Thank you.

PR #8747, opened 2019-11-15
fix program HTML titles

mattstratton:
The logic for setting the HTML titles on pages was missing the "program" page type.

Test it at https://deploy-preview-8747--devopsdays-web.netlify.com/events/2019-cape-town/program


bridgetkromhout:
I checked this against the case where it _was_ set (Minneapolis) and this is fine there too: https://deploy-preview-8747--devopsdays-web.netlify.com/events/2019-minneapolis/program - LGTM!

PR #8776, opened 2019-11-21
Revert "[VIX-2020] Add Platinum sponsor VipRede"

MikeRosTX:
Reverts devopsdays/devopsdays-web#8774


MikeRosTX:
Was a mistake reverting back the PR. Did it because I thought **_Zaragoza_** was using **_VipRede_** as a sponsor that had a different url.  We're safe to merge.

somatorio:
if the revert is a mistake shouldn't we close this instead of merging it?

PR #8777, opened 2019-11-22
[VIX-2020] add sponsor Platinum VipRede

sprt-balarini:
add sponsor Platinum VipRede


data/events/2020-vitoria.yml
```diff
@@ -151,6 +151,9 @@ sponsors:
   - id: guiafoca
     level: community
     url: https://guiafoca.org/
+  - id: viprede
+    level: platinum
+    url: https://guiafoca.org/
```
somatorio: are you sure this is the right URL?

PR #8782, opened 2019-11-23
[VIX-2020] add Silver sponsor AWS

sprt-balarini:
- add Silver sponsor AWS
- reintegration of Vinicius Camata in the organizers team


sprt-balarini:
I already sent the email to info@devopsdays.org

PR #8810, opened 2019-11-28
[PT-2020] welcome page update with cfp; propose page addition

eduardopiairo:
* welcome page update with cfp; 
* propose page addition.


somatorio:
Wasn't this added at #8809?

PR #8813, opened 2019-11-29
[rec-2019] - Added organizers pictures and event program

ronaldmiranda:
Updates organizers page with their pictures and adding an event program


ronaldmiranda:
DO NOT MERGE! - Few corrections incoming.

PR #8819, opened 2019-11-30
[Saint Louis 2020] Updating STL organizers

bridgetkromhout:
Removing and adjusting organizers per email.


bridgetkromhout:
The appveyor errors don't appear to be due to the content of this PR.

PR #8822, opened 2019-12-01
[Zürich 2020] Adding event group for Zürich

bridgetkromhout:
Hi, @RomanoRoth - as you've altered the city's display name for 2020, we need to use an event group to ensure that past events are linked properly.


RomanoRoth:
Hi @bridgetkromhout 
Thanks for changing.

PR #8834, opened 2019-12-03
[Kraków 2020] Update team_members

kgadek:
* remove Piotr Szwed
* sort all `team_members` by surnames (as it should be)

> If you are adding or removing organisers, please email info@devopsdays.org with the details of who is being added and removed, along with the full names and email addresses of the new team, so we can update Slack and the mailing list.

Mail will be sent briefly.


bridgetkromhout:
Removed departing team member from Slack and email as well.

PR #8849, opened 2019-12-05
[HOU-2020] Updating Sagar K's photo

coolblknerd:


bridgetkromhout:
@coolblknerd are you able to resize and crop this image so it's smaller and square? Thanks.

PR #8862, opened 2019-12-09
[TLV 2019] Fixes to issues with program

bridgetkromhout:
Hi, @shar1z! These fixes are mostly for issues around the way your program file is constructed. You have a good example of a multi-track event, which we could support better - the down side to manually building the program, of course, is less in the way of automated cross-checks.


bridgetkromhout:
Remaining issues @shar1z probably wants to look at:
- Is Andrei Burd still teaching Zen and the art of multi-cloud maintenance (Terraform, Consul, Vault and Nomad) with Lev Andelman?  (no speaker file)

- Why you should start public speaking and why ignites are the right place to start | Gil Zellner (no speaker or program file)

- Exactly Once Delivery is a Harsh Mistress Natan Silnitsy (no speaker or program file)

Also, you can look at how I fixed the program file for some multi-speaker talks; that's usually a better idea (aliases) than making two different pages.

PR #8878, opened 2019-12-12
[GDL-2020] adds community sponsors

estebancrw:
- adds itrevolution as community sponsor
- adds Docker CDMX meetup as community sponsor


data/events/2020-guadalajara.yml
```diff
@@ -99,8 +99,14 @@ proposal_email: "guadalajara@devopsdays.org" # Put your proposal email address h
...
 sponsors:
-  - id: meetup-docker-guadalajara
+  - id: docker
     level: community
+    url: https://www.meetup.com/Docker-Guadalajara
+  - id: itrevolution
+    level: community
+  - id: docker
+    level: community
+    url: https://www.meetup.com/Mexico-City-Docker-friends
```
bridgetkromhout: Hey, can you use an image that's specific to the meetup? I don't think we should use a company's corporate logo to represent a non-commercial meetup. It incorrectly implies that the corporate entity is getting a (free) community sponsorship, which will make other sponsors wonder why.

MikeRosTX: OK @bridgetkromhout, I'll work with @estebancrw on your suggestion. 👍

estebancrw: it should be fixed now 🙌

PR #8911, opened 2019-12-18
[Kraków-2020] update venue

kgadek:


kgadek:
Chocolatey fails on hugo installation. Not sure there's much I can do about it now.

PR #8923, opened 2019-12-21
[VIX-2020] adding meetups images

sprt-balarini:
adding meetups images


bridgetkromhout:
Do you want to look at https://deploy-preview-8923--devopsdays-web.netlify.com/events/2020-vitoria/meetups and see if this looks right? I think there might be a problem with the aspect ratio.

sprt-balarini:
Yes. I will review the sizes. Each picture was taken from different cell phones and so the difference so big.

PR #8930, opened 2019-12-26
[Vix 2020] Local images

bridgetkromhout:
Hi, @sprt-balarini - I noticed that we merged https://github.com/devopsdays/devopsdays-web/pull/8923 but I think we missed seeing the images were being linked from offsite - we try not to do that because then images become unavailable years later and are missing. /cc @MikeRosTX 

(Also, you don't need the organizer info from the contact page to be on other pages - looks like you'd just cloned that page.)


MikeRosTX:
Yeap, sorry I missed that @bridgetkromhout. I was looking at the resolution and size.

PR #8934, opened 2019-12-27
[NYC-2020] Add MS + 2Sigma to sponsors

kuritz:


data/sponsors/two-sigma.yml
```diff
@@ -1,2 +0,0 @@
-name: "SPONSORNAME"
-url: "URL"
```
kuritz: this looked like an abandoned config, lmk if it should not be deleted.

PR #8942, opened 2019-12-30
[ATL-2020] adding Franlkin Mosely as an organizer

cacorriere:
Adding Franklin as an organizer.


MikeRosTX:
No need to add to Slack, but dl email for Atlanta has been updated.

PR #8950, opened 2020-01-02
[AAR-2020] Add Devopsdays Aarhus 2020 event

RandomSort:
Venue scouting at the moment. Hope to close down venue and dates within the next two weeks.

I have used the non-existent alias: `organizers-aarhus-2020@devopsdays.org` with the assumption it would be created. I will be happy to change it, if I have done something wrong :)

Cheers!


bridgetkromhout:
Hi, @RandomSort - we recommend using the script (https://github.com/devopsdays/devopsdays-web/blob/master/utilities/add_new_event.sh) to generate new events because it is updated correctly, but if you don't, then the template you want to follow is https://github.com/devopsdays/devopsdays-web/blob/master/utilities/examples/data/events/yyyy-city.yml#L63 - so your email would be `aarhus@devopsdays.org`. Also, you manually copied from Copenhagen - please do make the necessary corrections to use your city name - thanks.

bridgetkromhout:
Also, event groups are used for events that move from city to city. Will this event be running in lieu of the Copenhagen event this year?

RandomSort:
Will recreate and repush using the script.
And yes @bridgetkromhout we will do it in Aarhus, for Denmark this year, with same organizers committee except for a few local additions :)

mattstratton:
@bridgetkromhout I had discussed with the organizers about creating a "Denmark" event group to include Aarhuas and previous Copenhagen events. I was planning on just doing the event group code for them after the Aarhaus event was created :)

bridgetkromhout:
@mattstratton sounds good - waiting for their corrections.

RandomSort:
I have used "Danish" rather than "Denmark" as the event group as the wording on the website is "more <event_group> events", which felt weird to me with Denmark. I'd be happy to conform to convention though. My creativity lacked sorely and we don't have a cool label like "Down Under" to use. I think "Up Over" would not be commonly recognized :D

RandomSort:
I have updated according to your requests. We still use the devopsdayscph. It is so far on purpose to keep the audience close. Do not know what our plan is going forward. Any comments @JasmineCrozier

mattstratton:
This looks good, but the only thing remaining is we need an email sent to info@devopsdays.org with the names and email addresses that should be added for the aarhus@devopsdays.org email address; please make sure that this list of names matches what you have on the PR. Thanks!

RandomSort:
Hi @mattstratton,
I sent one email yesterday from my work email. I have copy pasted the email and sent it from a different email address, to be sure. Thank you!

mattstratton:
Email alias has been created, and slack invites sent to the two new organizer additions. I'm going to go ahead and merge this now.

PR #8983, opened 2020-01-07
Archiving late 2018

bridgetkromhout:
Much like in https://github.com/devopsdays/devopsdays-web/pull/8332 I'm archiving past events here.

Reason:

Every event that needs to be built slows down the build. For past events no longer being built, we do not need to incur this overhead every time the repo builds.

Method:

* Leaving the data file alone
* generating the public directory: `hugo --theme=devopsdays-theme --buildDrafts=false --baseURL="https://www.devopsdays.org/"`
* copying the generated files to static
* removing the content .md files

And this is how I generated this PR:

```
cd public/events/
for foo in $(ls -d 2018-*);do cp -pr $foo/* ../../static/events/$foo/;done 
cd ../..
git rm -r content/events/2018-*
```

Before:

```
$ hugo server -w --baseUrl="http://localhost:1313"

                   |  EN    
+------------------+-------+
  Pages            |  5146  
  Paginator pages  |     2  
  Non-page files   |    41  
  Static files     | 12222  
  Processed images |     0  
  Aliases          |   315  
  Sitemaps         |     1  
  Cleaned          |     0  

Built in 73542 ms
```

After:

```
$ hugo server -w --baseUrl="http://localhost:1313"

                   |  EN    
+------------------+-------+
  Pages            |  3651  
  Paginator pages  |     2  
  Non-page files   |    39  
  Static files     | 13771  
  Processed images |     0  
  Aliases          |   258  
  Sitemaps         |     1  
  Cleaned          |     0  

Built in 34405 ms
```

Events archived:

2018-belem
2018-belo-horizonte
2018-bengaluru
2018-berlin
2018-boston
2018-cairo
2018-cape-town
2018-chattanooga
2018-columbus
2018-detroit
2018-edinburgh
2018-feira-de-santana
2018-florianopolis
2018-fortaleza
2018-galway
2018-hartford
2018-istanbul
2018-kansas-city
2018-kazan
2018-kiel
2018-london
2018-maringa
2018-montreal
2018-moscow
2018-newcastle
2018-oslo
2018-paris
2018-philadelphia
2018-phoenix
2018-raleigh
2018-riga
2018-rio-de-janeiro
2018-salvador
2018-santa-maria
2018-shanghai
2018-shenzhen
2018-silicon-valley
2018-singapore
2018-taipei
2018-tel-aviv
2018-warsaw
2018-wellington


mattstratton:
One thing I've started to think about with this model of archiving is that the "other CITY events" doesn't keep working? I mean, let's say you archive 2018 events. they will list the 2019/2020 versions as links (assuming a 2020 event has a date at this point in time) but they get frozen with that going forward.

I don't have much of a better solution for it right now, so I'm okay with doing this at this point, but something we may want to consider?

mattstratton:
As an example - if you look at 2017 chicago, which was archived prior to the 2020 chicago event existing, it doesn't show a link to 2020

image

mattstratton:
One way to solve this is to not archive the `welcome.md` (and therefore, the `/welcome/index.html` of the output) for archived events; this would be a bunch of extra work in the archiving step I think though.

bridgetkromhout:
Yes, and this has also been the case with all the "legacy" pages. I think having https://devopsdays.org/chicago as the authoritative "most recent" page helps with this a bit.

I agree that "update the archives from time to time" would be a nice model but I'm aware that everything is tradeoffs. We aren't incurring significantly worse SEO by doing this, and even without on-site search, https://devopsdays.org/events has a complete listing.

So, yes, you're pointing out a problem which definitely is real though I don't think it's critical. I can think of one fix - adding a link to the authoritative most recent page to the top of at least the front page for that event. I can open an issue to explore how we do this if you like. I don't think "don't archive welcome.md" is going to be workable.

PR #9005, opened 2020-01-09
[DET-2020] Add Detroit 2020 with dates TBD

aaronmaturen:
Same team as last year, just no location contract yet.


bridgetkromhout:
Thanks for getting this started! It looks like you cloned your 2019 event instead of using the script to generate the new event, so I have a couple suggestions:
- We're going to be encouraging everyone to move to the [new code of conduct template](https://github.com/devopsdays/devopsdays-web/blob/master/utilities/examples/content/events/yyyy-city/conduct.md), so you'll be ahead of the game if you include that in this PR.
- Change the mailto: to use detroit@devopsdays.org instead of the 2019 address

PR #9010, opened 2020-01-09
[VIX-2020] sponsor name correction

bridgetkromhout:
Hi, @sprt-balarini - we do not use special characters like `!` in filenames - it can have unexpected results. Also it appears that this sponsor already existed without the exclamation mark in the filename. Thanks.


bridgetkromhout:
This is a correction for https://github.com/devopsdays/devopsdays-web/pull/9001 which I merged without noticing this problem.

Issue #8195, opened 2019-09-08
automate the aliasing of a "current" url


bridgetkromhout:
A number of cities use a shortlink that just takes people to their current year, and then sometimes forget to update it for the next year. Let's look into adding an alias for the most recent year that's been published, so:

devopsdays.org/minneapolis goes to https://devopsdays.org/events/2019-minneapolis/welcome/
devopsdays.org/seattle goes to https://devopsdays.org/events/2020-seattle/welcome/
devopsdays.org/helsinki goes to http://devopsdays.org/events/2014-helsinki/

mattstratton:
This would be handled most likely in the redirects file; automating might be sketchy since we manually make changes to that file too. 

But `sed` might come to the rescue

mattstratton:
I'm thinking that what we could do is this:

Start by manually creating all the aliases in `static/_redirects`. An example would look like this:
```
/seattle		/events/2020-seattle
```
Update the `utilities/add_new_event.sh` script to check for a line that matches to `$DOD_CITY`. If it matches, then replace the line with:
```
/$DOD_CITY    /events/$DOD_YEAR/$DOD_CITY
```
If it doesn't find a line for `$DOD_CITY` then it should create it.

The only tricky thing is that right now `mention-bot` is configured to request a review from @mattstratton if `static/_redirects` is touched. We likely would want to edit the `.mention-bot` file to remove this, as I don't want to get tagged in a review on every single new event :)

mattstratton:
My bash-fu isn't as good as @bridgetkromhout 	so I am going to suggest this seems like a thing she could do :)

bridgetkromhout:
I'll take a look. It would be better if we could include a different file in `static/_redirects` but that doesn't look feasible. Maybe using the toml file instead is a better option.

mattstratton:
What we *could* do is move the existing redirects to the netlify config file, so that `_redirects` is only used for this purpose.

The main reason I wanted to be notified on it was to make sure nobody breaks any existing redirects. Frankly, as long as PR merge folks are watching the changes in that file it's Probably Fine.

mattstratton:
FWIW, making changes to `netlify.toml` is WAY scarier for things getting messed up that `_redirects`, as the latter only does one thing. So I really don't want a process or PR's to touch the `netlify.toml` file :)

mattstratton:
I also realized just now that `netlify.toml` isn't in `.mention-bot` so I'm going to fix that :)

[...]

PR #9102, opened 2020-01-18
[DEN-2020] Adding event passed banners to 2019 page.

mattk42:


mattk42:
It seems the 2019 shows up first in google searched for "devopsdays denver" and we have had some confused sponsors and attendees reach out over that. This is attempting to curb the confusion.

I did not see a standard way to handle this, but may have missed it. Just let me know if there is a better way to go about it,

bridgetkromhout:
So we've talked about this before. It's not super-easy to automate changes to non-current events because we eventually archive them. We did add https://github.com/devopsdays/devopsdays-web/issues/8195 to allow for (eg) https://devopsdays.org/denver to be a permanent "current" address, but that doesn't solve the SEO issue - past events by definition will rank higher in search, since they've got more links and more content and have been around longer.

I'll leave it up to @mattstratton to decide if this is something we want to open an issue and try again to tackle; I'm reluctant to because I don't think there is a simple, generalizable solution.

Meanwhile yes, what you're doing (putting a notice on a past event) is one option.

PR #9105, opened 2020-01-19
[Mombasa-2020] Kickstart site for Mombasa

rgevaert:


bridgetkromhout:
Hi, Rudy - I'm excited about this! As I mentioned in email, we can start this process of listing the event when we have three organizers. This is exciting, but it's too early. :)

PR #9107, opened 2020-01-20
[Austin 2020] Suggested update

bridgetkromhout:
Hi @scottsbaldwin! We have updated the suggested language for the code of conduct - see #9006. If you let us know that you approve this PR, we will merge it. Thanks.

Issue #8951, opened 2020-01-02
Update default Code of Conduct template


mattstratton:
When creating a new event via the script, there is a default Code of Conduct template that generates the CoC page. This template has been unchanged for many years, and it could likely stand with some modernization/updating. This issue is for tracking any feedback organizers (or other relevant parties) may have on this template, and working on updating it.

The template file is located at https://github.com/devopsdays/devopsdays-web/blob/master/utilities/examples/content/events/yyyy-city/conduct.md

lizthegrey:
Suggest hiring outside experts to evaluate it, but in short term would suggest that we remove the obligation by victims to report misconduct to event organizers.

somatorio:
I remember someone (@mattstratton?) sending this one at #general some time ago https://2019.jsconf.eu/code-of-conduct/ and i really like it :)

MikeRosTX:
> I remember someone (@mattstratton?) sending this one at #general some time ago https://2019.jsconf.eu/code-of-conduct/ and i really like it :)

I **_really like A LOT_** jsconf.eu CoC. The **_ONLY part_** that I personally do not like or agree with, is the **_Anonymous Reporting_** for several reasons (again this my sole personal opinion):

1. If things escalated out of proportion to the extreme that lawyers need to get involved (have not seen it happen yet, but it can potentially happen) according to the United States Constitution Sixth Amendment, it guarantees the rights of criminal defendant (in this case the CoC violator) the right to know who their accusers are, the nature of the charges, and evidence against them. Outside the United States, the **_Anonymous Reporting_** could very well be accepted just fine.

2. If a person is filing an **_Anonymous Report_**, how will the filer (offended person) know if we, as the organizers fully investigated the report, and took whatever action was necessary to prevent a recurrence if we can’t guarantee following up with them directly on the anonymous report? How will they even know if actions were taken or the anonymous report simple dismiss and filed away? Talking to the CoC violator without corroborating evidence could potentially be interpreted as defamation of character if we present them with the anonymous report and not the offended person who filed the report.

jjasghar:
If we are focusing on
> IV. If I am offended or harassed, I will inform people around me who make me feel safe and the event organizers. If I feel safe, at my discretion, I will inform those giving offense of the specific actions with the hope that the other party is well-intentioned and ignorant, but I am under no obligation to do so.

[...]

PR #9149, opened 2020-01-20
[austin-2020] Updating conduct page

scottsbaldwin:
Bringing the conduct page in line with the core organizing code of
conduct, but keeping the participate pledge we have used in Austin in
the past.

### DIFF
--- a/content/events/2020-austin/conduct.md
+++ b/content/events/2020-austin/conduct.md
@@ -1,16 +1,21 @@
 +++
 Title = "Conduct"
 Type = "event"
-Description = "Code of conduct for devopsdays Austin 2019"
+Description = "Code of conduct for devopsdays Austin 2020"
 +++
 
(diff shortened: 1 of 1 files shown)


scottsbaldwin:
@bridgetkromhout, I have taken the changes you put in your PR for the Austin conduct page and retained our participant pledge. Can we use this instead of the PR (https://github.com/devopsdays/devopsdays-web/pull/9107) you submitted?

bridgetkromhout:
If you check out the discussion in https://github.com/devopsdays/devopsdays-web/issues/8951 you will find some participants pointing out the problematic nature of some of the language. That’s why we removed that from the template. Make sure your org understand the implications discussed there before you decide to keep that language.

scottsbaldwin:
@bridgetkromhout, thanks for the link to the concerns. We'd like to keep the participant pledge portion for Austin.

PR #9147, opened 2020-01-20
[VIX-2020] Update speakers and add program

somatorio:
By request of @sprt-michelle
Please wait for her approval here in the comments before merging

PR #9177, opened 2020-01-22
[Vitoria 2020] The logo on the front page must be square.

bridgetkromhout:
Hi, @sprt-michelle & @somatorio - the logo on the front page must be square, and I didn't notice the problem in https://github.com/devopsdays/devopsdays-web/pull/9147. I am fixing it here.


bridgetkromhout:
The problem I'm correcting is the image not being laid out correctly on the front page, which you can see here.
Screen Shot 2020-01-22 at 3 43 25 PM

PR #9224, opened 2020-01-28
added sample accessibility text

kmugrage:
Added sample text from MN to the location page to encourage events to include information about accessibility.


bridgetkromhout:
Great notion! Note that you might want to point to this on #general, so people can incorporate a version for their events if desired; about 75% of events have already set up their pages for 2020, so they won't get this by default.

PR #9228, opened 2020-01-29
[PDX-2020] Breaking down silos & keeping the PNW weird!

jyee:
We're back yo!


nvwls:
LGTM

data/events/2020-portland-or.yml
```diff
@@ -0,0 +1,103 @@
...
+    twitter: "miiiiiche"
+  - name: "Rich Burroughs"
+    twitter: "richburroughs"
+  - name: "Scott Nasello"
+    twitter: "scottnasello"
+  - name: "Tiffany Longworth"
+  - name: "Will Ellett"
+    employer: "Cozy"
+
+organizer_email: "organizers-portland-2019@devopsdays.org" # Put your organizer email address here
```
mattstratton: this should be `portland-or@devopsdays.org` - we don't use the year-based email aliases anymore.

bridgetkromhout:
The logo says "Portland" still (without the disambiguation) but as that's not going to affect navigation to the correct page, it's fine.

bridgetkromhout:
Note that we have removed Simone Smith from Slack and email due to her not being listed here, and we've added Scott Nasello to email and re-activated his Slack with the requested email.

PR #9229, opened 2020-01-29
[Porto Alegre-2020] Adding new community sponsor and speaker photo

csdiedrich:
Adding new community sponsor and speaker photo


data/sponsors/afropython.yml
```diff
@@ -1,2 +1,2 @@
 name: "AfroPython"
-url: "http://afropython.org/"
+url: "http://afropythonconf.org"
```
somatorio: afropython.org still does work and afropythonconf.org seems to be more related to porto alegre (as it is for an event from afropython there) i think it's better to set this url at your event yml so it doesn't affect other cities/years having this sponsor

PR #9247, opened 2020-01-30
Add swag vendor recommendations to organizing.md

DivineOps:
Added swag vendor recommendations based on past experience in Chicago, Seattle


bridgetkromhout:
Good notion to list the cities that can give more info about a specific vendor option.

PR #9248, opened 2020-01-30
[ATL-2020] Instana logo update

cacorriere:
opening a PR with Atlanta 2018 and Sao Paulo 2019 to see if it works in the preview.


bridgetkromhout:
As discussed, pre-2019 is archived. If you'd like to us to merge this change, at the minimum update all the yml files. Optionally, also update the static files.

cacorriere:
instana-before-20170810.png was on Nashville 2017 but not Boston. Both events were after August of 2017, so I updated them to instana-before-20200127 to keep it consistent.

I updated all the yml files in an editor with find/replace, did the static sites with find and sed just on html files. 

e.g.
`find static/events/2018-atlanta \( -type d -name .git -prune \) -o -type f -name *.html -print0 -exec  sed -i '' -e  "s/instana\.png/instana-before-20200127.png/g" {} \;`

I can update the docs if/when this PR lands cleanly. Let me know if I missed/need to fix anything.

bridgetkromhout:
Looks good.

PR #9272, opened 2020-02-03
[PDX-2020] Remove boilerplate submission info

jyee:
Our CFP is not yet open. We need to need to remove the boilerplate info to prevent email submissions.


nvwls:
LGTM

PR #9282, opened 2020-02-04
[Boise 2020] Updating CFP dates

tresni:
Also add Chef as Silver sponsor

### DIFF
--- a/data/events/2020-boise.yml
+++ b/data/events/2020-boise.yml
@@ -15,7 +15,7 @@ enddate: 2020-05-15T19:00:00-06:00 # The end date of your event. Leave blank if 
 
 # Leave CFP dates blank if you don't know yet, or set all three at once.
 cfp_date_start: 2019-11-01 # start accepting talk proposals.
-cfp_date_end: 2020-02-15 # close your call for proposals.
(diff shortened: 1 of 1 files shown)


data/events/2020-boise.yml
```diff
@@ -15,7 +15,7 @@ enddate: 2020-05-15T19:00:00-06:00 # The end date of your event. Leave blank if
 
 # Leave CFP dates blank if you don't know yet, or set all three at once.
 cfp_date_start: 2019-11-01 # start accepting talk proposals.
-cfp_date_end: 2020-02-03 # close your call for proposals.
+cfp_date_end: 2020-02-15 # close your call for proposals.
```
kmugrage: As noted in the comments above, using this format for your dates can lead to issues. For example, it will appear to be open for several hours longer or shorter than you intended. I'd recommend using the full string.

# All dates are in unquoted 2020-MM-DDTHH:MM:SS+TZ:TZ, like this:
#   variable: 2019-01-04T00:00:00+02:00
#   variable: 2019-01-05T23:59:59+02:00
# Note: we allow 2020-MM-DD for backward compatibility, but it can lead to unexpected behaviors (like your event disappearing from the front page during your last day)

PR #9294, opened 2020-02-05
Fix issue with sharing images

### LINKED ISSUE #8659: Sharing images are not working
Looking at https://devopsdays.org/events/2019-chicago/welcome/, you can see that even though a sharing image exists in the directory, it is setting the default devopsdays sharing image.

I also see that Facebook sees this

```
og:image:alt | Devopsdays Des Moines
```

mattstratton:
This modifies the method for setting sharing images. Now they must be set in the datafile. This fixes the issue where they were not working before. Also retroactively fixes 2019 and 2020 events that had images set.

Fixes #8659 

Please do not merge this without review/approval from @bridgetkromhout and/or @adrianmoisey 


mattstratton:
I ran a few samples (main page, city page with custom sharing, and talk page with custom sharing image) through the facebook debugger as well as the twitter card validator, and they all look fine.

I did notice that the way we were setting `og:type` was off of the spec, so I popped in another commit to fix that too.

bridgetkromhout:
Checked this over and it looks good.

PR #9296, opened 2020-02-05
Fix speaker image stuff

### LINKED ISSUE #8198: Resize speaker and organizer images
Commit daeeb1e updated the site to use css to resize images.

This may have had some unintended consequences for images that are larger than the space allowed.

Some images become very tall but not very wide:

Baron_Schwartz_-_devopsdays_Chicago_2019

Was this the intent?

mattstratton:
Fixes #8198 

I *think* this fixes it. We need to do a bunch of checking on various events to make sure it looks right before merging.

For testing, please use this URL https://deploy-preview-9296--devopsdays-web.netlify.com/

### DIFF
--- a/themes/devopsdays-theme/layouts/speaker/single.html
+++ b/themes/devopsdays-theme/layouts/speaker/single.html
@@ -34,10 +34,10 @@
   <div class = "col-md-3 offset-md-1">
     {{- with .Params.image -}}
       {{- if ne . "" -}}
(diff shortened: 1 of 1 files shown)


mattstratton:
Note: with this change, if a speaker image is not square, it will be made square on the speakers page, but on the speaker/talk pages it will use its own aspect ratio.

Examples:

image
image
image

bridgetkromhout:
Looks like this works.

PR #9343, opened 2020-02-10
Update Hugo to v0.64.1

mattstratton:
This seems to be okay on local testing, but we need to test the deploy preview prior to merging.


mattstratton:
I checked several events and a few specific pages (speaking, etc) that are special...it all seems fine to me!

PR #9352, opened 2020-02-10
[GDL-2020] Redis added as gold sponsor

blacked7:
Redis added as gold sponsor


MikeRosTX:
Already spoke to @blacked7 about updating sponsor logos to follow guidelines using the format of:
**_sponsor-before-yyyymmdd_**

image

PR #9356, opened 2020-02-11
SEA-2020 Adding speakers and program

harbinger55:
Add all accepted speakers for DoD Seattle 2020 and create the program


harbinger55:
premature PR

PR #9380, opened 2020-02-12
[EHV-2020] Add team member information

rgevaert:
Add Jos to organizers lists. Also moves me to the non local list. Add Joep his twitter.


MikeRosTX:
Every time new organizers are add, an email should be sent to the Core team with that person's email address so we can add the to the team's email dl and Slack.

We'll proceed to merge this PR, because I found Jos's Gmail address

PR #9393, opened 2020-02-13
[HOU-2020] Adding 4 speakers to the page

skryukova:
Note that the program.md, speakers.md and changes to other pages are intentionally omitted to avoid "premature" publishing of the program.


skryukova:
@bridgetkromhout Thank you for catching the accidental title commit!

PR #9395, opened 2020-02-13
[BUF-2020] Updating the Stark and Wayne logo

aspleenic:
Updating sponsor log as they have a new one.


MikeRosTX:
Even though this is not the way we update sponsor logos, we will proceed to merge this PR because this sponsor has only sponsored BUF.

PR #9398, opened 2020-02-13
Add CODEOWNERS file

mattstratton:
give that [mention-bot](https://github.com/facebookarchive/mention-bot) is not longer maintained, it seemed logical to move to [CODEOWNERS](https://help.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners).

I believe this config mostly matches how we were set up before. Note: we can use a GH team name instead of specific people, but I didn't want to tag all of core for any PRs that need these reviews; eventually we should probably set up a GitHub team called something like `@devopsdays/web-reviewers` or something and assign them in the `CODEOWNERS` file instead. But this should be right for now.


mattstratton:
Please comment if you see any other sensitive/non-content files that should kick off a review.

We used to want a review if `static/_redirects` was modified, but that happens with every event now, so maybe not :)

mattstratton:
Should probably add `/static/robots.txt` also :)

glasnt:
I would strongly recommend using teams, and having role separation (mayhaps 'lead core'  @mattstratton @bridgetkromhout, 'maintainers' that includes me (specifically what I signed up for))

MikeRosTX:
> I would strongly recommend using teams, and having role separation (mayhaps 'lead core' @mattstratton @bridgetkromhout, 'maintainers' that includes me (specifically what I signed up for))

Agree with @glasnt, it might best if possible by teams.

mattstratton:
I definitely would rather use teams; let’s use this issue to bikeshed on what areas go with which grouping, maybe?

adrianmoisey:
:+1: 🎉 
Good change!
I agree on using teams instead of individuals, but that can happen at a later stage.

bridgetkromhout:
Are we moving this to use teams? Thanks.

mattstratton:
I’ll set up some initial teams and configuration later today.

mattstratton:
OK, so I set up some teams and updated to use them (please note we can add more folks to the teams as we go but this was my first pass)

- @devopsdays/core-infra : this is me and Bridget; this is for stuff directly related to netlify, circleCI, etc
- @devopsdays/theme-maintainers : folks who work on the theme itself and know it well. This is currently just Bridget and me
- @devopsdays/code-reviewers : other folks who might be able to put a set of eyes on PRs that touch things like content, utilities, etc. Right now I have just have @phrawzty and @MikeRosTX in there with Bridget and me, but we can add @adrianmoisey and @glasnt as well (that happens outside of the codeowners file; it's in the GitHub config)

bridgetkromhout:
> @devopsdays/core-infra :

Maybe call it build-infra so it doesn't get confused with "core organizers"?

> we can add @adrianmoisey and @glasnt as well 

Yes; let's.

mattstratton:
I renamed the group to `build-infra` and updated the CODEOWNERS file. I haven't made changes to the group membership yet :)

mattstratton:
I think the teams are mostly okay right now; they shouldn't be a blocker to merging.

build-infra is @bridgetkromhout @mattstratton @glasnt @adrianmoisey 
theme-maintainers is @bridgetkromhout @mattstratton @adrianmoisey 
code-reviewers is @bridgetkromhout @mattstratton @adrianmoisey @glasnt @MikeRosTX @phrawzty

mattstratton:
Given that I think y'all have all weighed in, and I've made the changes that were requested, I'm going to merge this so we can test it :)

PR #9401, opened 2020-02-14
[ANK-2020] add Devopsdays Ankara

serhatcan:
New event: Devopsdays Ankara.
I've sent the organizers list to core.


data/events/2020-ankara.yml
```diff
@@ -0,0 +1,98 @@
...
+event_twitter: "devopsdaysank" # Change this to the twitter handle for your event such as devopsdayschi or devopsdaysmsp
+description: "Devopsdays is coming to Ankara!" # Edit this to suit your preferences
+ga_tracking_id: "" # If you have your own Google Analytics tracking ID, enter it here. Example: "UA-74738648-1"
+
+# All dates are in unquoted 2020-MM-DDTHH:MM:SS+TZ:TZ, like this:
+#   variable: 2019-01-04T00:00:00+02:00
+#   variable: 2019-01-05T23:59:59+02:00
+# Note: we allow 2020-MM-DD for backward compatibility, but it can lead to unexpected behaviors (like your event disappearing from the front page during your last day)
+
+startdate:  2020-10-01T00:00:00+03:00
```
MikeRosTX: **startdate** & **enddate** needs to be remove if contract for venue has not been signed/confirmed.

serhatcan: it is certain, we are just talking about final arrangements with two venues and will decide today. so, we can continue.

PR #9405, opened 2020-02-14
Update image docs

mattstratton:
The docs needed updating with the current state of speaker images.


bridgetkromhout:
Checks out.

PR #9409, opened 2020-02-14
[ANK-2020] Removing dates as requested.

MikeRosTX:


MikeRosTX:
@serhatcan this PR is being submitted, because I requested **startdate** & **enddate** be removed in #9401 until venue location is confirmed. But change was not made. 

We do not allow events date to be published if venue is not confirmed. Once venue is confirmed, you can add it along with the date in a new PR.

PR #9428, opened 2020-02-18
[MIAMI-2020] Miami devopsdays - create TBD event.

iberkner:
This is the Miami team's initial PR to create the TBD 2020 Miami event page on the deveopsdays website.  What has been done:

- Miami 2020 event page created
- Organizer information updated
- Logos uploaded
- Twitter account added


bridgetkromhout:
The logo-square needs to be square, and it's a good idea to have as little whitespace around the edges as possible so it looks as big as it can, on the front page. Something like this:
logo-square

PR #9454, opened 2020-02-20
Fix for welcome page when event spans months

### LINKED ISSUE #9453: Date on welcome page doesn't work properly if event spans months
see buffalo 2020 for example

mattstratton:
Fixes #9453


mattstratton:
I tested this with a month-spanning event, a non-month spanning event, and a single day event - all look good.

mattstratton:
I discovered the issue exists in the slack unfurl too; problem is I can't test the fix since the info is cached I think...so we'll have to check it when it gets merged.

PR #9455, opened 2020-02-20
[la 2020] update effx logo

irabinovitch:
[la 2020] update effx logo


bridgetkromhout:
Fine to change without preserving, as this is the first use.

PR #9501, opened 2020-02-25
[MTL-2020] Add Montreal 2020 event with dates

idcrosby:
Initial add of 2020 event for Montreal, with event dates.

PR #9503, opened 2020-02-25
[Montreal 2020] Correcting time zone

bridgetkromhout:
Hi, @idcrosby! You probably want the right TZ; correction for https://github.com/devopsdays/devopsdays-web/pull/9501. (This fix will prevent your event from being removed from the front page mid-event, if a site build happens during your second day.)


idcrosby:
Thanks a lot for catching this @bridgetkromhout!

PR #9530, opened 2020-02-28
[2020-bmore] Tenable requested we use their wide logo.  Created an al…

bushong1:
…ternate tenable sponsor to accomodate without impacting other events


bridgetkromhout:
They've only sponsored one other event before, so this update only needs to happen to preserve the old logo for Baltimore 2019.

bridgetkromhout:
I'm closing this PR and doing the necessary updates in https://github.com/devopsdays/devopsdays-web/pull/9538 instead. Thanks!

PR #9565, opened 2020-03-04
[HOU-2020] Fixed speaker info for Paul Meharg and Anjuan Simmons

coolblknerd:


bridgetkromhout:
This will require a bit more time to figure out what's triggering the error.

MikeRosTX:
Merging PR, because failing AppVeyor could potentially be tied to Hugo versioning.

PR #9586, opened 2020-03-06
[LIMA-2020] Add Lima-2020 event and organizers

snahider:
We also sent a email to info@devopsdays.org with the organizers.


MikeRosTX:
Let's chat on Slack, because I don't have permissions to do any changes.

MikeRosTX:
The event creation script drops in the additional-language files. If they aren't even linked, it's not crucial to remove them.

PR #9589, opened 2020-03-07
[NYC-2021] Init

kuritz:
- init event page for 2021
- rm contacts for 2020 coc page so we don't get spammed

NOTE: I intentionally left organizer pictures out at this point in case anyone doesn't come back next year in consideration of not-growing-the-repo-until-needed


MikeRosTX:
We don't recommend disabling organizer images, because with so many organizers worldwide we need checkpoints to keep web, Slack, and email all in sync.  If organizers decide not to come back, you can remove them once it's confirmed.

MikeRosTX:
> We don't recommend disabling organizer images, because with so many organizers worldwide we need checkpoints to keep web, Slack, and email all in sync. If organizers decide not to come back, you can remove them once it's confirmed.

Disregard previous message, just realized that only images were disabled which is perfectly fine.

kuritz:
Thanks! Sorry I wasn't more clear the organizer entries were still there!

PR #9638, opened 2020-03-13
Use newer Hugo

ghoneycutt:

PR #9648, opened 2020-03-18
Standardize on Hugo usage

ghoneycutt:
This standardized on using cibuilds/hugo instead of the normal
jojomi/hugo docker image as well as using 0.67.1 everywhere.

See https://github.com/devopsdays/devopsdays-web/pull/9638 for more history.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2
 jobs:
   production:
     docker:
-      - image: cibuilds/hugo:0.65.2
+      - image: cibuilds/hugo:0.67.1
     steps:
       - checkout
       - run:
@@ -11,7 +11,7 @@ jobs:
 
   build:
     docker:
-      - image: cibuilds/hugo:0.65.2
+      - image: cibuilds/hugo:0.67.1
     working_directory: ~/devopsdays
     steps:
       - checkout
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -2,7 +2,7 @@ version: 1.0.{build}
 install:
 #  Install Hugo from chocolatey and display the version number
 - cmd: >-
-    choco install hugo --version=0.67.0
+    choco install hugo --version=0.67.1
 
     hugo version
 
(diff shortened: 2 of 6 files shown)


ghoneycutt:
@somatorio @mattstratton continuation from https://github.com/devopsdays/devopsdays-web/pull/9638 by standardizing on a docker image and a version of hugo.

ghoneycutt:
`./hugoserver.sh` fails on me.. Is it important to use `cibuilds/hugo` instead of `jojomi/hugo` ? I don't get why we don't use the official release from jojomi.

mattstratton:
We don’t use it because it didn’t work on circleCI.

ghoneycutt:
Trying jojomi everywhere and circleci dies before I can even write this :>

ghoneycutt:
Not sure what's different about cibuilds/hugo as I don't really know anything about hugo. Perhaps someone else can take a look and see if running hugo-server from this docker image is possible.

somatorio:
you need to make some changes in the docker run for cibuilds image... 
[Here](https://github.com/devopsdays/devopsdays-web/pull/9638#issuecomment-598904507) is how i made it work...

jojomi: `docker run -tip 1313:1313 -v $(pwd):/src:cached -e HUGO_THEME=devopsdays-theme -e HUGO_WATCH=1 -e HUGO_BASEURL="http://localhost:1313" --name hugo-server jojomi/hugo:0.67.0`

cibuilds: `docker run -tip 1313:1313 -v $(pwd):/home/circleci/project:cached -e HUGO_THEME=devopsdays-theme -e HUGO_BASEURL="http://localhost:1313" --name hugo-server --entrypoint "" cibuilds/hugo:0.67.0 hugo server --watch --bind ""`

What changed and why:
$(pwd):/src:cached -> $(pwd):/home/circleci/project:cached - that's where cibuilds expect our hugo dir to be

HUGO_WATCH=1 -> removed - that was a env var that jojomi's script used, hugo doesn't know this one as far as i could search

added --entrypoint "" - for some reason the cibuilds/hugo entrypoint wasn't hugo, so we set it as blank and use hugo at the CMD (we could set it as hugo, but that wouldn't be so readable, at least to me)

added hugo server --watch --bind "" - here we start hugo in the server mode (HUGO_WATCH was used for that at jojomi image) without bindind to a address (it was bindind to 127.0.0.1 without it)

mattstratton:
Oh yeah, I was reviewing on my phone and didn't see that the `docker run` command hadn't been updated; I assumed that @ghoneycutt had seen that in the comments of the last PR...but yeah that's the trick and thanks for the deep dive, @somatorio !!

ghoneycutt:
Thanks! Got the hugoserver scripts updated. I ran the .sh though am unable to test the ps1.

PR #9651, opened 2020-03-19
[Columbus-2020] Adding cancelled notice.

kjenkins19:
Adding a cancellation notice for 2020. If you have any additional suggestions for what to add, remove, etc. Let me know. We will working with the venue to assign new dates for next year.

Also removed Warner and Renee. E-mail as requested has been sent.

Thanks!

### DIFF
--- a/content/events/2020-columbus/welcome.md
+++ b/content/events/2020-columbus/welcome.md
@@ -81,6 +81,13 @@ Description = "devopsdays columbus 2020"
   </div>
 </div>
 
+<br />
+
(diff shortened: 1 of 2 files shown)


mattstratton:
I would suggest removing the dates from the data file as well - see how it is done in https://github.com/devopsdays/devopsdays-web/blob/master/data/events/2021-new-york-city.yml#L16-L17 for example (basically leave the fields there but delete the dates)

mattstratton:
You may want to delete the data file and the content folder(s) altogether at some point, but I see the reason to keep it on the site for now (so that people can see the notice you posted). I would definitely remove the dates though.

serhatcan:
Slack and email updated for removed organizers. Merging now but please consider removing dates as Matt suggested @kjenkins19 . Thanks!

PR #9652, opened 2020-03-19
[ATX-2020] Adding alert to POSTPONE

scottsbaldwin:


mattstratton:
You might want to delete all the dates from those fields as well?

scottsbaldwin:
@mattstratton, I'm going to remove the dates once we have a clearer picture for reschedule vs cancellation.

mattstratton:
Makes sense. We are working on coming up with some consistent guidance on what to do with the website for all this, so it's all in flux of course :)

mattstratton:
The main reason I suggest removing dates is that this affects how the homepage looks - and right now with the dates intact it shows up as coming up very soon. Removing the dates would push the event into the "TBD" part of the sidebar instead of showing up under April.

I appreciate we are all just trying to figure out the best way to handle all of this!

scottsbaldwin:
We definitely know we aren't holding the event on those dates. We do have a committee working on a potential reschedule dates. Our hope was to just update the dates one time.

scottsbaldwin:
@mattstratton I have just removed the dates.

PR #9661, opened 2020-03-22
[DET-2020] Add Logos

aaronmaturen:


MikeRosTX:
Where are you planning to use **_ccs_map.png_**?

aaronmaturen:
It  was used last year on our location page and just got copied over... just removing it to be simpler

PR #9674, opened 2020-03-31
Aju2020 - Postponing DOD

igorleoem:
2020-aracaju.yml event date changed


MikeRosTX:
Although the reason is **OBVIOUS** why the event is being postponed, please put a message on the **Welcome** page explaining the reason why it is being postponed

PR #9691, opened 2020-04-07
Always ensure a trailing newline before appending to this file

adrianmoisey:
Got the solution from https://unix.stackexchange.com/a/31955


bridgetkromhout:
Looks like this works! Thanks.

PR #9703, opened 2020-04-15
Add instructions for canceled events

mattstratton:
This is my first pass at some instructions on what folks should do in a PR if an event needs to be canceled.


bridgetkromhout:
Most of the non-US English-speaking world says `cancelled` not `canceled` - maybe go with the international spelling if we want to try to be as clear as possible?

mattstratton:
>maybe go with the international spelling if we want to try to be as clear as possible?

Heh. I have been spelling it "cancelled" myself a lot lately and everyone makes fun of me, so I thought it was wrong. I guess I am just international!

mattstratton:
question - should we link to this on the README?

bridgetkromhout:
> question - should we link to this on the README?

Yes, I think that's wise.

bridgetkromhout:
And maybe point to a PR like Minneapolis where it's been done?

mattstratton:
>And maybe point to a PR like Minneapolis where it's been done?

the only thing I don't love about using the MSP PR is that the same PR included the 2021 event addition; it's a little hard to follow.

PR #9707, opened 2020-04-16
Add field to event data file to keep cancelled events from showing alongside upcoming events

### LINKED ISSUE #9681: Consider adding option "canceled" field for events
Currently, with events being canceled, we are suggesting that they keep the current event in the site, in case people have links to it, but remove the dates.

This will make the event still show up on the sidebar under "TBD".

This also causes some confusion if a new event for the following year shows up.

My proposal is to add an optional field to the `YYYY-CITY.yml` file to reflect that the event was canceled. This would then be used to redact the event from display.

Thoughts?

mattstratton:
This adds support for the `cancel:` field in data files, which prevents events from showing up in the sidebar, on future lists, home page, etc.

Tested with 2020 MSP in this PR.

Fixes #9681

### DIFF
--- a/data/events/2020-minneapolis.yml
+++ b/data/events/2020-minneapolis.yml
@@ -13,6 +13,8 @@ ga_tracking_id: "" # If you have your own Google Analytics tracking ID, enter it
 startdate:
 enddate:
 
+cancel: "true"
+
 # Leave CFP dates blank if you don't know yet, or set all three at once.
 cfp_date_start:
 cfp_date_end:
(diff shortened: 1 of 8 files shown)


mattstratton:
The main point of this was to keep them from being listed. Folks still have to update their pages, etc, as noted in https://github.com/devopsdays/devopsdays-web/blob/master/utilities/docs/cancel-event.md

This is not a "one stop shop" feature to cancel your event. It's a setting to keep events that have been cancelled, but the pages must remain, from not displaying.

mattstratton:
@MikeRosTX I recommend you look at the issue #9681 which is what this resolves, to understand the scope.

We did discuss in that issue potentially adding a bash script that might automate some of the other tasks that are needed, but this is not comprehensive.

I'll update the title of this PR to be more clear. I can see how the title was misleading.

mattstratton:
>Because it was added to Minneapolis's site, but theirs wasn't a true cancel because it was postponed

yes, the instance is a cancellation - the 2020 instance of MSP is not occurring.

bridgetkromhout:
This is scoped correctly and solves the issue of removing cancelled events from the TBD sidebar.

PR #9718, opened 2020-04-20
[Sponsor-GCP] Add utm tracking codes

nathenharvey:
This does not change the URL that is linked but does add in utm tracking codes for, well, tracking and click counting.

Signed-off-by: Nathen Harvey <nathenharvey@google.com>

### DIFF
--- a/data/sponsors/googlecloud.yml
+++ b/data/sponsors/googlecloud.yml
@@ -1,3 +1,3 @@
 name: "Google Cloud"
-url: "https://cloud.google.com"
+url: "https://cloud.google.com/?utm_source=web&utm_campaign=devopsdays"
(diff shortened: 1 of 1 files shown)


bridgetkromhout:
While this does make a change for all events, it seems reasonable and useful. It doesn't pass along any individually-identifying infomation, merely that the click came from devopsdays (which could be equally found from the HTTP referer). @mattstratton for another opinion but this LGTM.

mattstratton:
I think it’s fine 👍

nathenharvey:
Thanks.  I considered creating a new data file for use moving forward but did not feel that this change warranted adding another to the repository and updating all of the previous events.

PR #9722, opened 2020-04-22
[MSN-2020] Removing madison 2020 since we never got it scheduled and aren't goin…

jazimmerman:
…g to run it this year.

@bridgetkromhout asked me to get this PR up based on what the plan was for this year given COVID-19.


bridgetkromhout:
Sadness, but also makes perfect sense. Thanks!

PR #9724, opened 2020-04-22
[ATL-2020] canceling 2020

cacorriere:
made changes per docs.


bridgetkromhout:
Thanks. When you have next year's date confirmed, it's a good idea to put it "on the map" by PR-ing in the minimal 2021 event generated by the new event script, as then it will be listed with dates when sponsors start looking.

PR #9728, opened 2020-04-27
[Porto Alegre-2020] Canceling the DOD Porto Alegre

csdiedrich:
Canceling the DOD Porto Alegre 2020


bridgetkromhout:
Hi, @csdiedrich, I am glad that you are making the right choice for your event. A few things to fix here; please let me know if you'd like help with these:

1) We recommend removing pages like your program and redirecting them to the welcome page, like this: https://github.com/devopsdays/devopsdays-web/blob/master/utilities/docs/cancel-event.md - we also recommend removing the dates.

2) You're adding new sponsors at the same time as cancellation of the event. Are you sure you want to do this? Did you instead want to list your event for 2021 with the sponsors who are staying?

3) If you are going to update the logo for sponsors, this is the correct way: https://github.com/devopsdays/devopsdays-web/blob/master/utilities/README.md#updating-a-sponsor-image-or-data

bridgetkromhout:
Thank you for the update!

PR #9751, opened 2020-05-13
[Msk-2020-2021] cancel event. add 2021 TBD

Frodox:
We are sorry, but we decided to cancel Msk 2020.
But we have scheduled 2021! :)


bridgetkromhout:
Glad to hear you'll return in 2021!

PR #9756, opened 2020-05-20
[NASH-2021] Updated 2020->2021

jwyckoff:


mattstratton:
Hey @jwyckoff - is the list of organizers changing from 2020 to 2021? If so, we need an email to info@devopsdays.org so we can update email lists, etc. 

also confirming that the intended behavior is the existing 2020 URL will completely disappear (if you haven't promoted the 2020 event, etc, and don't want to have any kind of verbiage about the cancellation on it, this should be fine.

if you want to preserve the 2020 website url (in case it's been shared) and have a notice about the cancellation, here are some instructions to help: https://github.com/devopsdays/devopsdays-web/blob/master/utilities/docs/cancel-event.md

jwyckoff:
thanks @mattstratton.  We are officially removing any reference of 2020 since we didn't do anything for this year's planning. 

I'm waiting for one last check on who's in for organizing this year.  Once complete, I'll send an email to info@, and resubmit this PR (I have a couple other small changes and will include them in the next one).

Issue #8663, opened 2019-11-01
Add search feature


mattstratton:
I thought we had an issue for this already, but we don't.

It's not terribly difficult, and there are several cases where it might be helpful.

Here's a helpful howto! https://blog.jeremylikness.com/blog/dynamic-search-in-a-static-hugo-website/

mattstratton:
I just did this for the ADO website, and it wasn't too hard.

Here's the PR for comparison https://github.com/arresteddevops/ado-hugo/pull/474

mattstratton:
Huh. This won't actually work out - when we move stuff to `static` it will be removed from the index, so the search will only work for relatively current content. Hmm.

mattstratton:
For reference, the key is creating this `index.json` page, which gets done with code like this:

```go
{{- $.Scratch.Add "index" slice -}}

{{- range (where $.Site.Pages "Type" "episode") -}}
    {{- $.Scratch.Add "index" (dict "title" .Title "subtitle" .Params.subtitle "description" .Params.description "image" ( printf "/%s" .Params.episode_image ) "content" .Plain "permalink" .Permalink) -}}
{{- end -}}

{{- range (where $.Site.Pages "Type" "guest") -}}
    {{- $.Scratch.Add "index" (dict "title" .Title "subtitle" .Params.subtitle "description" .Content "image" ( printf "/%s" .Params.Thumbnail ) "content" .Plain "permalink" .Permalink) -}}
{{- end -}}
{{- $.Scratch.Get "index" | jsonify -}}
```

so as you can see it needs to range over Pages. Which get lost when we migrate the generated pages of older events.

mattstratton:
We might be able to do it with google - this post is how Jeremy did his search prior to moving it into native Hugo, and it *might* address some of the concerns I have with using google search:

https://blog.jeremylikness.com/blog/more-hugo-migration-tips/

mattstratton:
actually that doesn't help; he uses AdSense with the search, which isn't what we want, of course!

mattstratton:
For some thoughts on this - I had set up a Google CSE for devopsdays a while ago (apparently!) and you can see what the search experience is like here (just look at the search box and then see the results; don't worry about page formatting):

https://cse.google.com/cse?cx=008769174023796316711:pj3syggtkz0

It embeds google ads at the top :(

mattstratton:
[...]

PR #9768, opened 2020-05-30
Test serverless search function

mattstratton:
Connected to #8663 

This creates a serverless function that we can query for search


mattstratton:
Search index is being built, however, the serverless function is returning the entire index. See https://github.com/sw-yx/netlify-plugin-search-index/issues/19

We need to wait to see what's up with that before moving forward. 

I'll close this PR but keep the branch up and add this update to the #8663

PR #9778, opened 2020-06-07
[CAI-20] Opening CFP 

aameen79:
And updating welcome page


data/events/2020-cairo.yml
```diff
@@ -37,9 +37,9 @@ location: "" # Defaults to city, but you can make it the venue name.
 location_address: "Online Event! Stay Safe!" #Optional - use the street address of your venue. This will show up on the welcome page if set.
 
 nav_elements: # List of pages you want to show up in the navigation of your page.
-#  - name: propose
+  - name: propose
 #  - name: location
-#  - name: local
+  - name: local
```
MikeRosTX, line 42: Since this is an online event, consider removing Local from the navbar or explain on the **_location.md_** page that it will be online,

Issue #9798, opened 2020-06-13
DOD_CITY environment variable breaks "add_new_event.sh"


skryukova:
I had DOD_CITY environment variable set, and I was not able to generate new event, without either deleting the variable or fixing the add_new_event.sh script on line 36
https://github.com/devopsdays/devopsdays-web/blob/55a9b876d93b12a3f122c685ea1ecc8a20984715/utilities/add_new_event.sh#L36

PR #9799, opened 2020-06-14
Fix variable name DOD_CITY

adrianmoisey:
Looks like this is a typo

Fixes https://github.com/devopsdays/devopsdays-web/issues/9798


mattstratton:
I’ve never used the environment variables, but this seems correct?

PR #9801, opened 2020-06-15
[AMS-2020] We don't do bulk tickets by invoice 

PaulusTM:
We don't do bulk tickets by invoice this year as tickets are free.


bridgetkromhout:
The email is just `amsterdam@devopsdays.org`.

PR #9818, opened 2020-06-24
[BUF-2020] Correcting Sponsor Statement for Virtual Event

aspleenic:
Making sponsorship notice clear for virtual event. Also resolving some merge errors on my fork.


aspleenic:
@MikeRosTX Yup - forgot to save them before checking them in

PR #9836, opened 2020-06-30
[AMS-2020] Add image to Karen

PaulusTM:
Missed the image for Karen.


MikeRosTX:
Was the image uploaded previously?

PR #9852, opened 2020-07-02
[Paris-2020] cancel event :(

phrawzty:
Paris 2020 is cancelled. I followed the [documentation](https://github.com/devopsdays/devopsdays-web/blob/master/utilities/docs/cancel-event.md) and believe this PR to be complete.

cc @jooooooon


bridgetkromhout:
Sorry to see you have to cancel, but of course safety is paramount. You do have a few more things that need changing - this PR only has file deletions. Thanks.

phrawzty:
> You do have a few more things that need changing - this PR only has file deletions. Thanks.

Right you are! This was me trying to use vscode to manage my commit for the first time. Now to _actually_ push all the changes. :P

mattstratton:
I just confirmed on the deploy preview. Only 2021 Paris shows up on all the pages, etc, so I think you are good.

PR #9861, opened 2020-07-06
Update deploy previews to disallow robots

### LINKED ISSUE #9860: deploy previews might be being indexed by google
This isn't happening a lot, but a link could get leaked. We want to make sure branch deploys and deploy previews are set to not be indexed.

This article explains how to do it:

https://www.tempertemper.net/blog/stop-search-indexing-for-netlify-deploy-previews-and-branch-deploys

if we don't want to do the npm stuff, we can just modify `bin/netlify.sh` to have a command at the end like this:

`echo 'User-agent: *\nDisallow: /' > public/robots.txt`

That will make the robots file be disallow for everything. This *should* work okay.

(note: need to also add this to `bin/netlify-development.sh`)

mattstratton:
Fixes #9860

### DIFF
--- a/bin/netlify.sh
+++ b/bin/netlify.sh
@@ -1,2 +1,3 @@
 hugo version
 hugo --theme=devopsdays-theme --buildDrafts=false --baseURL="$DEPLOY_PRIME_URL"
+echo 'User-agent: *\nDisallow: /' > public/robots.txt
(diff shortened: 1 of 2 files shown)


mattstratton:
This didn't work, so I am trying a different method to get the updated file over there :)

mattstratton:
OK, so I tested it and it works. Yay! I'll merge it.

PR #9893, opened 2020-07-20
Fix error on Chicago schedule

mattstratton:


data/events/2020-chicago.yml
```diff
@@ -297,7 +297,7 @@ program:
     type: open-space
     date: 2020-09-01
     start_time: "15:55"
-    end_time: "16:05"
+    end_time: "16:15"
```
adrianmoisey, line 300: Next on the schedule is at 16:20. Intentional 5 minute break?

mattstratton: this is still a rough schedule and likely to change a little...I just found a glaring error that I wanted to fix :)

PR #9904, opened 2020-07-26
Automate the process of changing a sponsor logo.

don-code:
This script implements the process defined in the [instructions for updating a sponsor image](https://github.com/devopsdays/devopsdays-web/blob/master/utilities/README.md#updating-a-sponsor-image-or-data). I also library-ified `sedcmd`, as I wanted it in my script, but didn't want to copy-paste it.

The output of this script can be seen at #9903.

**Note that I only tested this on a Linux box.** I don't know that I haven't broken any functionality on macOS.

### DIFF
--- a/utilities/README.md
+++ b/utilities/README.md
@@ -60,42 +60,21 @@ If you want to update a sponsor's info, keep in mind that we don't want to retro
 
 Determine which sponsor is being discussed. For this example, we'll use Pivotal.
 
-First, we'll do filename updates. Existing sponsors have an image and a data file:
-
+Run the script `change_sponsor_logo.sh` in the `utilities` subdirectory, giving it the name of the sponsor (`pivotal`) and the path on disk to your new logo:
 ```
-static/img/sponsors/pivotal.png
-data/sponsors/pivotal.yml
-```
-
-- When a sponsor wants an update to the image, data file, or both, first you copy the current files to a new filename containing the current date:
-
-```
-cp static/img/sponsors/pivotal.png static/img/sponsors/pivotal-before-20190307.png
-cp data/sponsors/pivotal.yml data/sponsors/pivotal-before-20190307.yml
+[don@zeus ~/devopsdays-web/utilities:don/bos-lightstep]  ./change_sponsor_logo.sh pivotal ~/Downloads/pivotal.png
+Modifying 2018-chicago.yml
+Modifying 2019-detroit.yml
+Modifying 2018-chicago/conduct/index.html
+Modifying 2018-chicago/contact/index.html
+...
 ```
 
-Do this even if you're only changing one of them.
-
-Then, make sure all events in the past get the update. Edit the "- id: pivotal" line for any past events:
-
-```
-$ grep "id: pivotal" data/events/*
-[...]
-data/events/2015-minneapolis.yml:  - id: pivotal-before-20190307
-data/events/2016-atlanta.yml:  - id: pivotal-before-20190307
-data/events/2019-minneapolis.yml:  - id: pivotal
-```
-
-It is possible that the previous city has already been archived and static items have already been created. You will have to update the image source for these pages
-
-```
-grep -r "sponsors\/pivotal.png" static/events/*
-[...]
-static/events/2016-chicago/contact/index.html:<img src="/img/sponsors/pivotal.png" alt="Pivotal" title="Pivotal" class="img-fluid">
-```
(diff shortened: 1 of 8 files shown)


bridgetkromhout:
Nifty! I don't have time to review today, but I'm excited about this. Thanks!

adrianmoisey:
At a glance, this looks good! 
I'll try do a more through review tomorrow.

utilities/add_new_event.sh
```diff
@@ -1,22 +1,10 @@
 #!/bin/bash
 
-
 set -e
 
-cd `dirname ${0}`
```
adrianmoisey, line 6: Could you re-add this ``` cd `dirname ${0}` ``` ?
Without it, the script can't be called from another location

adrianmoisey: Oh, I was reviewing the wrong file here.
I think this needs to move above `source common_code`

utilities/change_sponsor_logo.sh
```diff
@@ -0,0 +1,70 @@
+#!/bin/bash
+set -euo pipefail
+
+##
+# This script takes in a sponsor name, and path to a new logo. It replaces
+# all occurrences of the "old" logo in "old" events with an "old" sponsor
+# definition, and adds the new logo in its place for this year's events.
+##
+
+source common_code
```
adrianmoisey, line 10: I think this needs a ``` cd `dirname ${0}` ``` here in order to be run from any directory.

The problem is that `$IMAGE_TO_COPY` is relative to where it's run from.

I don't know of a better solution, but may be get the output of `pwd` before the `cd`, and concatenate that with `$IMAGE_TO_COPY`.
I tested it locally, and it seems to work, but is a little nasty.

utilities/README.md
```diff
@@ -60,42 +60,21 @@ If you want to update a sponsor's info, keep in mind that we don't want to retro
...
-```
-
-It is possible that the previous city has already been archived and static items have already been created. You will have to update the image source for these pages
-
+Run the script `change_sponsor_logo.sh` in the `utilities` subdirectory, giving it the name of the sponsor (`pivotal`) and the path on disk to your new logo:
 ```
-grep -r "sponsors\/pivotal.png" static/events/*
-[...]
-static/events/2016-chicago/contact/index.html:<img src="/img/sponsors/pivotal.png" alt="Pivotal" title="Pivotal" class="img-fluid">
+[don@zeus ~/devopsdays-web/utilities:don/bos-lightstep]  ./change_sponsor_logo.sh pivotal ~/Downloads/pivotal.png
```
adrianmoisey, line 65: Can this be replaced with a more generic prompt?
may be `$`?

utilities/add_new_event.sh
don-code: There are just a bunch of function declarations in `common_code`, so I think it shouldn't matter where. I could change it anyway if you want; I'd just want to update all of the others in turn.

utilities/change_sponsor_logo.sh
don-code: I was using a full path elsewhere on the disk (`~/Downloads/pivotal.png` in the example case) - it seems like it'd be easy to make a mistake if we require the "new" file to live inside the repository first, unless I move it instead of copy it.

utilities/README.md
don-code: Good call, I missed this.

utilities/add_new_event.sh
adrianmoisey: I just tested this locally, having ``` cd `dirname ${0}` ``` below the `source common_code` means that you have to run your script (in this case, `add_new_event.sh`) from the `utilities` direcgtory.

Swapping the order will fix that (restoring the previous behaviour).

utilities/change_sponsor_logo.sh
adrianmoisey: I think we should do two things here:
- add a ``` cd `dirname ${0}` ``` so that it can be called from anywhere
- In the readme call out that the path needs to be a full path
  - May be even mention it on line 19 of this file?

don-code:
Hi @adrianmoisey - sorry for taking so long to pick this back up.

Somehow it didn't occur to me that the scripts could get called as `utilities/add...` - so this is a classic case of not understanding what the code did before I changed it, that's my bad. I've put the order back, and tested `change_sponsor_logo` inside and outside of `utilities`.

mattstratton:
I just tested this myself and it's slick! Good job! I'll merge it.

don-code:
Awesome - thanks Matt and Adrian!

PR #9937, opened 2020-08-11
[FOR-2020] alter date event and remove member

renat473:
Update date events and add url c4p


MikeRosTX:
@renat473 please ping me on Slack.

PR #9950, opened 2020-08-19
[CHI-2020] Updated Launchdarkly logo for 2020 version and updated old logo file

margaretvaltie:
Cindy from LaunchDarkly sent a new logo updated in early 2020. I've renamed the existing logo to before 2020 and added the new logo with same .png file name


mattstratton:
@margaretvaltie there are more steps. You need to update all the older events that had launchdarkly to use the "old" filename. 

See https://github.com/devopsdays/devopsdays-web/blob/master/utilities/README.md#updating-a-sponsor-image-or-data

PR #9995, opened 2020-09-12
[Detroit 2020] canceled

bridgetkromhout:
Hi @rexroof - in https://github.com/devopsdays/devopsdays-web/pull/9945 you canceled the Detroit 2020 event, but you did not properly mark it so it would stop showing up as an upcoming event. This PR corrects that so it no longer displays as upcoming.

### DIFF
--- a/data/events/2020-detroit.yml
+++ b/data/events/2020-detroit.yml
@@ -10,7 +10,7 @@ ga_tracking_id: "UA-94092408-1" # If you have your own Google Analytics tracking
 #   variable: 2019-01-05T23:59:59+02:00
 # Note: we allow 2020-MM-DD for backward compatibility, but it can lead to unexpected behaviors (like your event disappearing from the front page during your last day)
 
-# cancel: "true"
+cancel: "true"
 startdate: 2020-11-11T00:00:00+02:00 # The start date of your event. Leave blank if you don't have a venue reserved yet.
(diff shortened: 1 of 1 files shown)


rexroof:
Per earlier suggestions I saw on the global slack, I did that on purpose so that it would still show up in the sidebar.

bridgetkromhout:
@rexroof If you would like your event to show up as an upcoming event without a specific date, then remove the "cancel" line and also remove the start and end dates. Thanks!

mattstratton:
@rexroof thats by design - canceled events should *not* show up in the sidebar/main page; those are for active events

mattstratton:
To be more clear, there are a couple different situations and what you should do will differ:

### Detroit 2020 is not happening and we know we won't have one at all
Remove the dates, flag the cancel flag, etc 

### Detroit 2020 is not happening as planned, but we are going virtual and we know the dates
Just change the dates in the YAML and you're golden

### Detroit 2020 is not happening as planned, but we might still reschedule it for 2020 maybe virtual who knows 
In this case don't flag "cancel", but clear out the dates. It will show in the sidebar under TBD. We don't want to put any dates for events that are not confirmed. 

Make sense?

rexroof:
gotcha.  I misunderstood the part about the dates, I had reviewed posts in the last from back in march and more recently and somehow missed that bit.    I had intentionally meant to cancel the event but leave it in the sidebar.  

Thanks for pointing it out, sorry if I caused either of you more work.      

I have another PR in the works, we have decided to cancel the event entirely.

PR #10027, opened 2020-10-03
[BOS-2020] Link to event videos.

don-code:


MikeRosTX (changes requested):
It's suggested not to mix **_html_** code with **_yaml_**. Can you please move the **_html_** code to **_welcome.md_** file?

don-code:
I initially thought the same, but changed the description instead because it's a more visibile callout to the videos.

Screenshot from 2020-10-03 13-57-15

If that's a matter of policy, I'm happy to push a change down into the body.

Screenshot from 2020-10-03 13-58-50

MikeRosTX:
I'm just suggesting what I've been told on the past, but @mattstratton has the final say and is the tie breaker. 😉

mattstratton:
I will review shortly!

mattstratton:
Okay yeah, this can cause some unintended side effects, which is why we prefer you not put html in those fields.

mattstratton:
So I would rather you put it in the body/welcome page. Sorry!

don-code:
No worries - it's now in the welcome body.

MikeRosTX (approved):

PR #10033, opened 2020-10-08
[BOS-2021] I heard we're doing another one.

don-code:
I've also sent an e-mail to the core organizing team to confirm 2021's organizers on the `boston@devopsdays.org` distribution list.


MikeRosTX (changes requested):
Please review the dates of the event is going to be in person at a venue onsite if a venue has not been selected. If it will continue to be virtual like 2020, then please make it clear it will be virtual and leave the dates AS IS.

MikeRosTX (approved):

PR #10072, opened 2020-11-03
removing 2020-poznan event

luprzybyl:
Unfortunately we did not gather enough topics to fulfill our agenda.
Therefore we needed to make a decision to cancel upcoming 2020 event.


MikeRosTX:
Resolved via #10074

PR #10083, opened 2020-11-10
[various years/cities] name update

bridgetkromhout:
Name update upon request.


mattstratton:
this looks good to me - all archive pages updated!

PR #10110, opened 2020-12-12
[TX-2021] Added Sagar as organizer, adjusted dates, added blank image…

skryukova:
… for Karl, added Twitter/LinkedIn links


MikeRosTX:
Just curious before merging this PR... Why was a blank image added for Sagar?

skryukova:
@MikeRosTX You mean for Karl? Because Karl is the only one without an image for now, and without an image the page formatting is very odd (company name overlaps person's name) and it is just really odd traversing the page, since he is 1 out of 12 organizers without an image. It would be nice if a blank image were provided by default, if no image is given.

PR #10120, opened 2021-01-05
[TX-2021] Added Bridgecrew.io as gold sponsor

skryukova:
[TX-2021] Added Bridgecrew.io as gold sponsor


data/sponsors/bridgecrew.yml
```diff
@@ -0,0 +1,3 @@
+name: "Bridgecrew"
+url: "https://bridgecrew.io/?utm_medium=sponsoredeventsURLutm_source=devopsdaysURLutm_campaign=tx-210302"
```
glasnt, line 2: The string URL appears in the middle of this URLs' UTM. Did you mean "&"?

I'm not sure what the general vibe for UTMs is in general, but this one will at one point expire, and if bridgecrew sponsor again it will require some sort of updating. 

Ping @mattstratton for comment

mattstratton: The better move is to override the url in the city/event data file so that the UTM version is only associated with this particular event.

mattstratton: So I would make the URL in the sponsor file the main website and then in the sponsor field for them in the Texas event data file override the URL. 

This is documented somewhere :) I'm on my phone so I can't find it handy right this second.

skryukova: Thank you. Will figure it out, and update the PR

mattstratton: here is the relevant section in the docs - https://github.com/devopsdays/devopsdays-web/blob/master/themes/devopsdays-theme/REFERENCE.md#sponsor-fields

mattstratton:
I would double-check @glasnt 's question here:

>The string URL appears in the middle of this URLs' UTM. Did you mean "&"?

mattstratton:
It's probably right, but just double-check? Otherwise, this looks good.

skryukova:
@mattstratton @glasnt not sure what happened to the URL, I am pretty sure I just did copy/paste from the vendor's email (and tested it as well). Anyhow, thanks for catching this. I fixed it.

PR #10127, opened 2021-01-13
[TX-2021] Updated chef information with new logo, added Chef as gold

skryukova:
Chef has asked to update logo, as they were acquired by Progress
Followed instructions here by renaming old files and updating various event yml and static files https://github.com/skryukova/devopsdays-web/blob/master/utilities/README.md#updating-a-sponsor-image-or-data
I also had to update some 2016/2015 files, because apparently there already was a rename chef-before-20160714, but it wasn't used where it was supposed to. Hope I did it correctly!


phrawzty:
This is a massive changeset. I did a spot check of a few changes and it looks fine, but I'd like one more set of 👀 before we merge. Thanks for understanding!

skryukova:
@phrawzty Understood. I would love for this to get through soon though, as this is needed to add the sponsor to texas 2021 and is basically an obstacle to other changes to.
I also think that given that changes like this require massive commits, I wonder if we should rethink the process for changing logos. Rather than use the "change the old references to sponsor-before-date-xxx" add the new sponsor data as "sponsor-after-date-yyy". I realize that this could be error prone, but won't result in commits like this.

somatorio:
Looks fine to me too

skryukova:
@somatorio @phrawzty any chance to get this merged today?

phrawzty:
> I also think that given that changes like this require massive commits, I wonder if we should rethink the process for changing logos. Rather than use the "change the old references to sponsor-before-date-xxx" add the new sponsor data as "sponsor-after-date-yyy". I realize that this could be error prone, but won't result in commits like this.

Agreed in principle. There's certainly a reason why it's _not_ this way, but it's absolutely worth bringing up with the folks that manage the website—which I'll do on the Slack right now. :)

PR #10136, opened 2021-01-18
enable rendering of emojis

yvovandoorn:
Hugo supports emojis via https://www.webpagefx.com/tools/emoji-cheat-sheet/. Just needed to turn it on.

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -10,6 +10,7 @@ archetypedir = "archetypes"
 PaginatePath = "blog"
 buildDrafts = false
 disableKinds = ["taxonomy", "taxonomyTerm"]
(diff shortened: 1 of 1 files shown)


mattstratton:
I think it's possible that I turned this off on purpose? I don't remember :)

mattstratton:
I mean that it might have broken something somewhere, but I'll check out the deploy preview and we can just turn it off again if we find an issue.

mattstratton:
Cursory check of the preview makes it look like it's fine; I'll approve and merge this.

mattstratton (approved):

PR #10151, opened 2021-02-03
[Kraków-2021] update event info

kgadek:


MikeRosTX (changes requested):
@kgadek see DM in Slack.

MikeRosTX (approved):

PR #10162, opened 2021-02-17
Cancel Paris 2021 :(

phrawzty:
I didn't bother with all the aliases because we didn't open any additional sections up in the first place.

Hopefully see y'all in 2022…

cc @jooooooon


somatorio:
static/events/2021-paris/logo.png

seems to be missing, so the logo doesn't load :/

phrawzty:
> static/events/2021-paris/logo.png seems to be missing, so the logo doesn't load :/

Correct. Since I don't really care about the logo I have just commented it out. Good catch!

PR #10180, opened 2021-02-24
[CAI-21] Updating sponsor page

aameen79:
Updating sponsor page


phrawzty:
@aameen79 To be clear, you're removing the sponsorship table. I just want to make sure this is the intended outcome as it's a big change.

aameen79:
@phrawzty 
Yes, we are changing the pricing, so updated table will be introduced later.

PR #10187, opened 2021-02-25
Update to Hugo 0.81.0

mattstratton:
Signed-off-by: Matt Stratton <matt.stratton@hey.com>

This updates the version of Hugo; it could provide better build performance as well.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2
 jobs:
   production:
     docker:
-      - image: cibuilds/hugo:0.76.5
+      - image: cibuilds/hugo:0.81.0
         auth:
           username: devopsdaysbuild
           password: $DOCKERHUB_PASSWORD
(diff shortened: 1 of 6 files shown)


mattstratton:
hugo build time with this version:
`Total in 104658 ms`

previous build:
`Total in 91822 ms`

so, uh, it's slower?

mattstratton:
It's about ten seconds slower with this version, but that is slightly anecdata maybe since only one build?

MikeRosTX:
> It's about ten seconds slower with this version, but that is slightly anecdata maybe since only one build?

I know 10 secs is not much difference, but I would not go backwards unless the new build provides some features that would benefit us. Just my .02¢.

mattstratton:
The main reason to update is that this is the version people are likely pulling down. We're pretty behind on versions so as long as it's not causing bugs, we should get ourselves updated. 

TBH one of my concerns is that I am on this version of hugo locally, so developing against it when we have an older version in prod could cause issues.

bridgetkromhout:
I would suggest merging this since those of us who use hugo for multiple projects are already on this version locally.

PR #10210, opened 2021-03-17
[SHA-2021] Add Shanghai event

overloadzy:
Add new Shanghai event, organizers are same as the last event.


mattstratton:
The error is happening on the sponsors partial; I did a quick look and the indent was incorrect on the navigation elements; let's see if this works.

If this fix doesn't do it, I'll pull it down locally and try to troubleshoot

mattstratton:
The build is passing (the netlify deploy preview isn't completed yet, but the hugo build step passed which is what was failing before).

I'll check it after the netlify deploy preview is completed, but I think this is fixed.

PR #10214, opened 2021-03-18
[TPE-2021]  Add new event

thebrecht:


mattstratton:
You should also update the `static/_redirects` file so that `/taipei` forwards to this event and not the 2019 version (it's not very easy for me to add new files to the PR or I would do it for you).

if you can't add this to this PR, let me know and I'll do another PR after we merge this one to update the redirects file.

thebrecht:
Hi Matt, thanks for the instruction, I think I'll add redirect file and then do another PR, is that ok?

PR #10220, opened 2021-03-25
[PRG-2021] event postponement

EtssMM:


glasnt:
Hi @EtssMM 

If you take a look at the preview, you'll see that with this change there are two Prague events that will be "TBD". This is going to confuse visitors. 

Since this event is newer, I suggest you take the time to mark it as deleted. Details on how to do that can be found here: https://github.com/devopsdays/devopsdays-web/blob/main/utilities/docs/cancel-event.md

I understand the 2020 event redirects to 2021, please make sure you address those. 

Please let us know if you need assistance making these edits.

PR #10247, opened 2021-04-10
[VAN-2021] add chime as gold sponsor and confirm another speaker

Velutas:
and remove community and streaming sponsor options


Velutas:
forgot to add time zone

PR #10250, opened 2021-04-10
[2021-PDX] CFP is open!

BillWeiss:
Woo! I'm not sure what the `cfp_date_announce` field does, but I don't seem to need it to get the propose link on the site.


MikeRosTX (changes requested):
Please set the announce date as per the remarks. 👇🏽👇🏽👇🏽
# Leave CFP dates blank if you don't know yet, or **set all three at once**.

Usually the announce date is 2 to 3 weeks after the end date.

BillWeiss:
@MikeRosTX there you go

MikeRosTX (approved):

PR #10278, opened 2021-04-30
[HOU-2021] Removed two organizers

skryukova:


mattstratton:
Were the two organizers removed from the email alias? I haven't had a chance to check yet.

PR #10303, opened 2021-05-13
[BUF-2021] Add TBD preliminary event info for DODBUF 2021 and add new organizers.

onelioubov:
New organizers were already added to Slack and distro.


mattstratton:
@onelioubov you might want to also add a change to this line to this PR so that devopsdays.org/buffalo redirects to this year event:

https://github.com/devopsdays/devopsdays-web/blob/main/static/_redirects#L19

PR #10332, opened 2021-05-25
[HOU-2021] Replaced AlertOps logo per sponsor's request

skryukova:
[HOU-2021] Replaced AlertOps logo per sponsor's request


glasnt:
new logo matches their website favicon, and this is the first instance of AlertOps as a sponsor, so I'm comfortable approving this change.

PR #10345, opened 2021-05-27
[HOU-2021] Changing AlertOps image per sponsor's request

skryukova:
[HOU-2021] Changing AlertOps image per sponsor's request


phrawzty:
n.b. Currently, this logo is only used in this event, so no old-logo dance necessary.

PR #10362, opened 2021-06-04
[AMS-2021] change sponsor logo

sheela-n:
This change is per sponsor request


mattstratton:
If this sponsor exists on events prior to AMS-2021, the sponsor logo change process should be followed so we do not "rewrite history":

mattstratton:
as discussed in the slack, #10357 should not have been merged and the retroactive sponsor logo changes should have been done in that PR.

We need to get the previous version of `vmware-tanzu.png` and use that to create `vmware-tanzu-before-06042021.png` and `vmware-tanzu-before-06042021.yml` and then update all the past events that use `vmware-tanzu` to use `vmware-tanzu-before-06042021` as the sponsor.

mattstratton:
Upon further review, it looks like the only change to the logo was the size, which means we should not have to worry about the historical changes, so this should be fine.

PR #10367, opened 2021-06-07
[AMS-2021] Add Splunk as silver sponsor

sheela-n:


data/events/2021-amsterdam.yml
```diff
@@ -116,6 +116,8 @@ sponsors:
     level: silver
   - id: contrast_security
     level: silver
+  - id: 2021-splunk
```
glasnt, line 119:
```suggestion
  - id: splunk
```

Splunk already exists as a sponsor, and while their image is a little low res, this PR only adds the new image, not the data required to make it a valid sponsor (it's not currently loading on the preview in it's current state)

PR #10375, opened 2021-06-08
[GE-2022] cancel 2021 edition + prepare 2022 edition

deniger:
- Cancel GE-2021
- Prepare GE-2022 and start CFP


mattstratton:
You will also want to update this line https://github.com/devopsdays/devopsdays-web/blob/05611a5c80f2231722cecf76a42267301fe0d186/static/_redirects#L40 to point to the 2022 version; you can either add to this PR or if it's easier, do a second PR after this one is merged (it'll be smoother if you add to this PR, but it would be okay to do it the other way too!)

deniger:
Thanks @mattstratton  for your feedback. I've just updated the redirects file ( and fixed some typos).

PR #10387, opened 2021-06-14
[POA-2021] Add new sponsor and program

csdiedrich:
[POA-2021] Add new sponsor and program


mattstratton:
There's something odd happening with the ignites as well, so let me take a look

image

mattstratton:
@csdiedrich it looks like you don't have files in the `program/` directory for the ignites/panel which is why they aren't showing up (see the screenshot in my comment above)

csdiedrich:
The ignites will be filled by the end of the month, we are still looking for the names.

csdiedrich:
We don't have ignites yet, so there's no content in this section, we'll close the ignites by the end of the month

data/events/2021-porto-alegre.yml
```diff
@@ -176,3 +179,103 @@ sponsor_levels:
...
+    start_time: "17:00"
+    end_time: "18:00"
+    background_color: "#b7da5b"
+  - title: "Encerramento e Sorteios"
+    type: custom
+    date: 2021-07-10
+    start_time: "18:00"
+    end_time: "18:30"
+
+ignites:
```
mattstratton, line 271: I'm confused about what these entries are for if it's not yet selected?

mattstratton:
well, I guess when I was trying to do the resolve of the conflicts, it ended up merging it, which is...fine?

csdiedrich:
Great, I will adjust ignites soon. Thanks.

PR #10414, opened 2021-06-23
[NASH-2021] Sponsor updates

GeoffCraig:
Adding sponsors and fixing a sponsor level


data/sponsors/2021-gitlab-nash.yml
```diff
@@ -0,0 +1,3 @@
+name: "GitLab"
+url: "https://gitlab.com/"
+twitter: gitlab
```
phrawzty, line 3: 👋  This file is identical to `data/sponsors/gitlab.yml`, so I assume that the reason you're duplicating it is because you want to use a new sponsor logo? If so, the process for changing sponsor logos is [documented here](https://github.com/devopsdays/devopsdays-web/blob/main/utilities/README.md#updating-a-sponsor-image-or-data) (it's just running a script basically 😉 ).

phrawzty (changes requested):

data/sponsors/2021-gitlab-nash.yml
bridgetkromhout: Same for the Teleport one - if you want to change the logo for `teleport`, do that, but don't make a new year-delimited sponsor. Thanks.

PR #10419, opened 2021-06-25
[AMS 2021] add cloud pirates as bronze sponsor

PaulusTM:


glasnt:
The domain for this sponsor does not appear to be valid. Is this the correct domain?

PaulusTM:
@glasnt Yes, looks a bit strange. They are working on a single page as its a startup that has no online presence yet.

glasnt:
My apologies for not getting back to this one in a timely manner.

PR #10435, opened 2021-07-06
[POA-2021] Adjusting program

csdiedrich:
Adding new speaker and adjusting bio.


bridgetkromhout:
Hi, @csdiedrich - you probably want to pull in changes from upstream so you can fix these conflicts. See https://github.com/devopsdays/devopsdays-web/blob/main/CONTRIBUTING.md#editing-your-site for details.

PR #10455, opened 2021-07-13
[AMS 2021] Post event updates

PaulusTM:
With this PR I'm updating the sessions to include a Youtube link and cleaning and updating the team page to include everyone that was involved and uploading pictures.


bridgetkromhout:
Hi @PaulusTM I suggest marking this one ready for review so we can get your videos up. It's fine to add another post-event PR later. Thanks!

PR #10471, opened 2021-07-22
[DEN-2022] Create Event

mattk42:


bridgetkromhout:
The French and Brazilian Portuguese files aren’t used by default for English-language events, even though the script adds files without regard for localization. So you can remove them if preferred with no ill effect; your event page won’t use them.

mattk42:
Thanks Bridget! I caught a couple of other updates to make this morning as well and think this should be ready to go. Let me know if you see anything else I missed.

PR #10476, opened 2021-07-25
[TLV-2021] New Sponsors Added

shar1z:
Hey there team - git really is HARD.

I deleted, rebased, reforked, refetched, created two different branches and I'm still not sure I'm 100% in sync with origin/master.  I also wasn't able to push with the suggested 'origin' & branch - had some kind of authentication issue (despite being on SSH not HTTPS and having a token).

I created the sponsors at least 5 times and then lost them every time I checked out main to verify I was up to date (even though I was sure I stashed).

In any case... TL;DR - I hope I don't have any major conflicts.


phrawzty:
👋  I'm not sure why you have such rotten luck with git, but this PR looks totally fine, so kudos. 😄

PR #10526, opened 2021-08-18
NASH-2021 Adding sponsors and changing sponsor logo

GeoffCraig:
This adds a few sponsors and updates the logo for a sponsor per request.


yvovandoorn:
Adding @mattstratton as a reviewer.

Reason: One of the files is from 2016, aren't those static sites now -- meaning this change needs to be made elsewhere as well right?

PR #10529, opened 2021-08-19
[VIX-2022] New logo and organizer

giovanibalarini:
- New logo for event
- New organizer Jandson Mendes


giovanibalarini:
I sent the email with data about new organizer

PR #10586, opened 2021-09-13
[PDX-2021] speaker photo updates

BillWeiss:
A couple of folks liked the drawn avatars we made for them :) Also, a bio update for Aaron.


phrawzty:
The drawn avatars are super great!

PR #10588, opened 2021-09-14
[HOU-2021] Adding JJs Abstract

jjasghar:
Can you double-check this @skryukova?


skryukova:
Looks good to me. Let's merge this, and I will add a sharing.png afterwards.

PR #10593, opened 2021-09-17
[PDX-2021] Add an image and update Aaron's bio

BillWeiss:


glasnt:
Hi @BillWeiss, it looks like you dropped a speaker image instead of replacing it in this PR. Is that being re-added elsewhere?

BillWeiss:
@glasnt you're right! I'll get the right image added to this PR shortly. Thanks for looking out!

BillWeiss:
Oh! No, this is correct. The file is still there, it just got a lot smaller. The diff view on the picture shows it's right.

PR #10597, opened 2021-09-17
[PDX-2021] Adding a Watch Live page

BillWeiss:
One of the embeds will have to be updated once the event starts, but this gets it into place. We would appreciate if this would be merged even in the slightly ugly state before the event.

"watchlive" as a top-level link is kind of gross but works. Eh. Suggestions appreciated :)


glasnt:
Seems fine to me! Happy to take any updates if you get a chance to make them before your event!

PR #10599, opened 2021-09-18
[PRG-2022] Moving Prague from 2021 to 2022

EtssMM:
- Event rescheduled
- CFP opened
- Inactive organizer removed
- Sponzor level updated


phrawzty:
I'm glad to see that your sponsors stuck with you!

PR #10601, opened 2021-09-19
[HOU-2021] Matt's talk is now a workshop

carlosonunez:
@skryukova PTAL

### DIFF
--- a/data/events/2021-houston.yml
+++ b/data/events/2021-houston.yml
@@ -402,11 +402,10 @@ program:
     start_time: "16:00"
     end_time: "16:30"
   - title: "matthew-ostlind"
-    type: "workshop"
+    type: "talk"
     date: 2021-09-22
-    start_time: "16:00"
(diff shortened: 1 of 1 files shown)


skryukova:
@carlosonunez There are a couple of issues with this commit (although it is mostly what we are trying to do).
- The title is in fact wrong, it is the reverse of what is intended and what the change is.
- The talk should be 30 minutes long, at the moment it is 40 minutes long.
- I believe the change in times is not enough to properly render on the website, the lines also need to be moved up in the file, so that the talk shows up in the correct spot (and not just with the correct times)

skryukova:
@glasnt @carlosonunez this PR can be cancelled. I've created a new one that fixes the issues above: https://github.com/devopsdays/devopsdays-web/pull/10604

carlosonunez:
What @skryukova said. Cancelling this. Refer to this PR instead: #10604

PR #10602, opened 2021-09-20
[CAI-21] Updating sponsors

aameen79:


phrawzty:
Note to other organisers: normally we don't accept changes to sponsor profiles that have already been used; however in this case it's obvious what the change is and does (fixes an outdated URL), so this time I'll give it the 👍 . :)

PR #10604, opened 2021-09-20
[HOU-2021] Moving Matthew Ostlind's workshop up in the schedule and c…

skryukova:
[HOU-2021] Moving Matthew Ostlind's workshop up in the schedule and changing to a talk


carlosonunez:
Looks better than my attempt, @skryukova ! LGTM.

PR #10606, opened 2021-09-21
[BOS-2021] Minor spelling fixes

don-code:
I am changing the title text of two sponsors to be human-readable, and one speaker's name to be spelled correctly.


glasnt:
Auto-accepting the speaker correction. 

Sponsor updates: looking at the netlify preview these `name` values are used for the alt/title text and do not need to match the yaml name, and the render looks fine. This change will affect other events, but in a positive way.

PR #10608, opened 2021-09-21
[PDX-2021] Swapping out Ceora's talk and QA

BillWeiss:
Last minute swap!


BillWeiss:
I don't think those Netlify failures are in something I changed...

PR #10625, opened 2021-09-26
[CAI-21] Adding Agenda, Speakers

aameen79:
And opening registration


aameen79:
@MikeRosTX @phrawzty @mattstratton 
Sorry for mentioning, but we need the update for the website urgently

PR #10626, opened 2021-09-26
[DC-2021] fix: removed CFP content

davezen1:
- Removed CFP content asking for proposals since proposals are now closed.


glasnt:
unrelated build failure

PR #10632, opened 2021-09-29
[BOS-2021] [HOU-2021] [DFW-2022] Update LaunchDarkly logo

don-code:
I'm updating the logo on behalf of Boston, but am also updating it retroactively for Houston on behalf of @skryukova.

I also found that the Dallas event is using an even older logo, and have fixed it.

Additionally, I fixed a bug in the `change_sponsor_logo` script, which prevented it from ignoring logos that have already been changed previously.


phrawzty:
LGTM! Thanks for the bugfix, too.

PR #10657, opened 2021-10-13
[BIRMINGHAM,AL-2022] Add Birmingham AL Event

mkoperator:
Adding Birmingham event and sponsors. In anticipation of mailing list and slack additions.


thecrazyrussian:
Hi @bridgetkromhout, we did the abbreviation to de-conflict with UK's Birmingham event that will also likely happen around the same time based on their previous guidance. What is the recommended way of dealing with that?

PR #10662, opened 2021-10-16
[HOU-2021] Adding SlidesLive embedded videos to all talks

skryukova:


skryukova:
I will redo this using different way.

PR #10698, opened 2021-11-01
[CAI-22] Creating Cairo 2022 event

aameen79:


phrawzty:
👋  This PR is fine and I will merge it. You will probably want to change the [locale redirect](https://github.com/devopsdays/devopsdays-web/blob/main/static/_redirects#L23) from 2021 to 2022 at some point, however. :)

PR #10709, opened 2021-11-10
[LDN-2022] Add 2022 TBD event

chriscoffee:
We've decided to host a 2022 event aiming at something smaller, the venue and time is currently undecided.


mattstratton:
I confirmed the organizer list; this lgtm!

PR #10732, opened 2021-12-02
[Aar-2022] Add Atea as sponsor and update copy on welcome page

RandomSort:
The title says it all :)

Thanks!


data/events/2022-aarhus.yml
```diff
@@ -98,6 +98,8 @@ organizer_email: "aarhus@devopsdays.org" # Put your organizer email address here
 sponsors:
   - id: eficode
     level: gold
+  - id: eficode
```
yvovandoorn, line 101: @RandomSort it looks like you're adding eficode again instead of atea. Please confirm.

RandomSort: Thank you for catching that. I updated to do the thing I was claiming to do. Cheers.

PR #10743, opened 2021-12-09
[NASH-2022] Adding event

GeoffCraig:
This adds the 2022 event in Nashville.


yvovandoorn:
LGTM @GeoffCraig. Looking at the organizers list, I'm not seeing any changes to what your PR has vs what we have listed.

PR #10769, opened 2022-01-05
[LA 22] Add new sponsors 

irabinovitch:
[LA 22] Add new sponsors


data/events/2022-los-angeles.yml
```diff
@@ -100,12 +100,16 @@ sponsors:
...
     level: gold
+  - id: komodor
+    level: gold
   - id: solo
     level: gold
   - id: twingate
     level: gold
   - id: blameless
     level: silver
+  - id: slimai
```
glasnt, line 111: Hi! You've added `slimai` here and as the image name, but `slim.yml` as the file name, so this won't work!

Can you please rename `slim.yml` to `slimai.yml` so they all match? Thanks!

PR #10773, opened 2022-01-06
[GE-2022] update logos by removing white borders

deniger:
[GE-2022] update logos by removing white borders


phrawzty:
Verified that this only affects this event (checkmarx in particular has a long history).

deniger:
Thanks. For Checkmarx, I did a recent change ( and renamed the log) so it should only affect Geneva last event

PR #10783, opened 2022-01-09
[ATL-2022] Changed Date

ravilach:
Changed Date for ATL 2022 to April 2022
Added "progresschef" as sponsor since name change was different from "chef" sponsor. Was at the request of Progress Chef.  Ran modify sponsor and was inconsistent with other events so safer to have a new sponsor vs update "chef". 

Cheers!

-Ravi


athenot:
Looks good to me. Reflecting new dates as we all agreed.

PR #10787, opened 2022-01-11
[EHV-2022] Enable the CFP link & add logo

PaulusTM:
The CFP is open. With this change it should be easier to know where.


static/events/2022-eindhoven/2020-eindhoven.yml
```diff
@@ -0,0 +1,120 @@
+name: "2020-eindhoven" # The name of the event. Four digit year with the city name in lower-case, with no spaces.
```
phrawzty, line 1: This file addition doesn't seem correct. Why put an un-used `2020` welcome page in your `2022` event?

phrawzty (changes requested):

PR #10830, opened 2022-01-31
[MEL-2022] Cancelling Melbourne 2022 event

ArjenSchwarz:
Due to ongoing restrictions and high case numbers we've made the call to
cancel this year's event and reconvene for next year.


phrawzty:
Sorry to see that. Hopefully next year!

PR #10841, opened 2022-02-07
2022-boise updates to orgs / dates / sponsors

jessucka:
Hey organizers! :wave:
We're updating our site with a few changes and would like to get this merged in.
This PR adds some of our new organizers, imagery for organizers and sets missing dates. Also fixes some sponsor information and the Eventbrite link.


ksatirli:
Hey folks!

I asked @jessucka to open the PR so it would originate from an organizer that is already listed for the event.

(I'm the original author of these changes and helping out with Boise in 2022)

PR #10854, opened 2022-02-13
[rdu-2022] adding speakers and program

normalfaults:
[rdu-2022] adding speakers and program


yvovandoorn:
Awesome line up @normalfaults!

PR #10864, opened 2022-02-16
[rdu-2022] adding jessica fields as speaker

normalfaults:
[rdu-2022] adding jessica fields as speaker


glasnt:
FYI your ignite speakers aren't appearing the schedule as yet (applies to all ignites, not just this one)

PR #10893, opened 2022-02-25
[ATX-2022] add sixnines as happy hour sponsor

scottsbaldwin:


glasnt:
I see this also updates the Tonic logo, but given Tonic is a first time sponsor, there's no historical changes needed

PR #10927, opened 2022-03-05
AMS 2022 bulk update

PaulusTM:
This PR updates the AMS 2022 event with:
- Start of the ticket sale ( Early Bird now live )
- Add all sponsor tears with amount of slots
- Add first sponsors


PaulusTM:
Using draft PR to validate the build. As soon build looks good, I'll update to regular PR.

PR #10930, opened 2022-03-07
[MDE-2022] Adding Call for papers button in Welcome Page

mpinedag12:
[MDE-2022] Adding Call for papers button in Welcome Page


phrawzty:
👋 It's not a requirement, I _heavily suggest_ also enabling the proposal link in the menu bar:
image

I will merge this PR now since there's nothing wrong with it, however. Feel free to open another PR with the proposal link noted above, if you like.

PR #10947, opened 2022-03-16
[AARHUS-2022] Add new organizers 

jonasvinther:
Adding new organizers to Aarhus-2022


phrawzty:
Hello,

Please send an email to `info@devopsdays.org` with the names and email addresses of the people you're adding and removing, so that we can keep your mailing list synchronised. Once we have that email, we can merge this PR at the same time.

PR #10957, opened 2022-03-17
Fix Speaking page

mccricardo:
The speaking page was broken, not showing events.

Screenshot 2022-03-17 at 15 18 30


mattstratton (approved):
Whoa. I don't know how this happened (I'm pretty sure I did it when trying to fix some perf stuff) good catch!!!

PR #10967, opened 2022-03-21
[TLV-2022] Hello World

shar1z:
Just opened the base files - added the organizers and basic info.
Hoping there are no major conflicts.

Thanks!


data/events/2022-tel-aviv.yml
```diff
@@ -0,0 +1,216 @@
...
+    website: "https://logz.io"
+    image: "dotan-horovits.jpg"
+
+  - name: "Avishai Ish-Shalom"
+    twitter: "nukemberg"
+    employer: "Null"
+    github: "nukemberg"
+    linkedin: "https://www.linkedin.com/in/sallyfields"
+    website: "https://statscraft.org.il"
+    image: "avishai-ishshalom.jpg"
```
phrawzty, line 120: There are two organiser images that didn't render properly on the preview. I'm not sure why since the images are present in the PR and render properly in the GitHub interface. If you build locally, do they render properly?

image

shar1z: Hey - yes - I did see them in my local preview.
It might have something to do with the jpg vs. jpeg format - I believe I resolved everything before committing.

shar1z: I also just noticed that I missed two "sally fields" in there :-) I'll fix it after the commit - just want to have the hello world up by our meetup on March 30th.

Thanks!

PR #10970, opened 2022-03-22
[ATL-2022] Atlanta initial schedule in plain text

aclairefication:
Minimum viable schedule
- Added color-coding key for schedule
- 2-day plain text schedule (i.e. no references to talk abstracts or speaker bios)

### DIFF
--- a/data/events/2022-atlanta.yml
+++ b/data/events/2022-atlanta.yml
@@ -157,30 +157,30 @@ program:
     end_time: "10:25"
     background_color: "#d5c5ef"
   
-  - title: "DevOps 1"
+  - title: "Build Custom Cloud Images with Packer & Ansible with Mason Egger"
     type: custom
     date: 2022-04-19
     start_time: "10:25"
     end_time: "10:55"
-  - title: "Security 1"
+  - title: "What Would Deming Do - Cyberwarfare with John Willis"
     type: custom
     date: 2022-04-19
     start_time: "10:25"
     end_time: "10:55"
     background_color: "#fffa99"
-  - title: "SRE & Complexity 1"
+  - title: "Trauma-informed Site Reliability Engineering with Lesley Cordero"
     type: custom
     date: 2022-04-19
     start_time: "10:25"
     end_time: "10:55"
     background_color: "#ffb566"
   
-  - title: "DevOps 2"
+  - title: "Building and Leading Remote Teams with Amber Vanderburg"
     type: custom
     date: 2022-04-19
     start_time: "10:55"
     end_time: "11:25"
-  - title: "Security 2"
+  - title: "Argue Better with Active Listening with Waldo Grunenwald"
     type: custom
     date: 2022-04-19
     start_time: "10:55"
--- a/data/events/2022-atlanta.yml
+++ b/data/events/2022-atlanta.yml
@@ -144,7 +144,7 @@ program:
     start_time: "09:15"
     end_time: "09:25"
     background_color: "#d5c5ef"
-  - title: "Keynote 1"
+  - title: "The Org that DDD & BAPO Built with Nivia S. Henry"
     type: custom
     date: 2022-04-19
     start_time: "09:25"
@@ -277,7 +277,7 @@ program:
     start_time: "09:15"
     end_time: "09:25"
     background_color: "#d5c5ef"
-  - title: "Keynote 1"
+  - title: "Trust Nothing, Not Even This Talk! with André Henry"
     type: custom
     date: 2022-04-20
     start_time: "09:25"
--- a/data/events/2022-atlanta.yml
+++ b/data/events/2022-atlanta.yml
@@ -131,19 +131,16 @@ program:
     date: 2022-04-19
     start_time: "08:00"
     end_time: "09:00"
-    background_color: "#d5c5ef"
   - title: "Opening Welcome"
     type: custom
     date: 2022-04-19
     start_time: "09:00"
     end_time: "09:15"
-    background_color: "#d5c5ef"
   - title: "Sponsors"
     type: custom
     date: 2022-04-19
     start_time: "09:15"
     end_time: "09:25"
-    background_color: "#d5c5ef"
   - title: "The Org that DDD & BAPO Built with Nivia S. Henry"
     type: custom
     date: 2022-04-19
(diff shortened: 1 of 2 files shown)


data/events/2022-atlanta.yml
```diff
@@ -124,3 +124,285 @@ sponsor_levels:
...
+    background_color: "#8ad0d3"
+  - title: "Open Space #3: Visit the Aquarium"
+    type: custom
+    date: 2022-04-20
+    start_time: "16:00"
+    end_time: "16:45"
+    background_color: "#8ad0d3"
+    custom_url: "https://www.georgiaaquarium.org/wp-content/uploads/2019/02/Email_Brand-Anthem_01.47.mp4"
+
+ignites:
```
phrawzty, line 388: These ignites render as such:
image

This is because the targets don't exist:
image

phrawzty: @mattstratton ooh right, that makes sense! So the solution—assuming they don't want to make talk pages—is to use a(nother) custom block for the ignites. Correct?

mattstratton:
Question: I assume that you will be adding markdown pages for the speakers and their talks? If this implementation of the program, that is without those pages, is what you’re going to have… You’ll do better to actually just use raw HTML on the program page instead of having to try to do it programmatically and fight the template :)

data/events/2022-atlanta.yml
mattstratton: To be honest, if the intent is to not create talk pages at all… It would be better to not use the programmatically generated program, and instead just use tables or something inside the `program.md` (I think the type for that page would have to be changed so it didn’t try to generate it)

I infer from “minimum viable” that this is temporary maybe?

mattstratton: I think trying to force something like ignites with that much information into a custom block is more than the programmatic program generation can support… It’s not intend for that

aclairefication:
I just needed a first draft to get a structure in place without having to understand everything since the directions were hard to follow. I intend to flesh it out afterward.
I used our Atlanta 2019 schedule as a starting point and will fill in details later this week, c.f. https://devopsdays.org/events/2019-atlanta/program

mattstratton:
Many cities will put in placeholders for the slots as they fill them (using custom slots with TBD for the ones they don’t have yet) but usually adding the pages for the talks and speakers happen at the same time as the initial program. 

There are some scripts to help make those!

aclairefication:
Hmm given that there was already a program.md with some placeholder info, would I have to delete it and then follow this set of steps in order to have the script generate the program for me? https://github.com/devopsdays/devopsdays-web/blob/main/utilities/README.md#program

aclairefication:
If you think it's better to just copy-paste-modify the HTML generated for our 2019 schedule, then I can redo the whole thing?

mattstratton:
The script is for starting from scratch; it can’t read in custom stuff already existing. 

The programmatic program doesn’t do very well with multiple tracks in general; I think that your best bet is to use the 2019 Atlanta one as a template as that’s the only devopsdays I personally recall with that many tracks.

mattstratton:
> If you think it's better to just copy-paste-modify the HTML generated for our 2019 schedule, then I can redo the whole thing?

Don’t copy paste that HTML; what I suggest is:

1) create the speaker and talk pages. I imagine that you have bios and abstracts for all of those talks… The speaker photos are optional
2) plug those talk IDs into the program YAML as you have it now, just change all the “custom” to use type of “talk” with the right ID.

aclairefication:
re: add_speaker.sh - should this be portable to bash on Windows 11?

There's no usage statement, so I guess the script takes no inputs until the prompts show? I can't tell because it doesn't run for me on my local, i.e. Windows 11 with bash shell GNU bash, version 5.0.17(1)-release (x86_64-pc-linux-gnu)

```
$ ./add_speakers.sh
./add_speakers.sh: line 2: $'\r': command not found
: invalid optionh: line 3: set: -
set: usage: set [-abefhkmnptuvxBCHP] [-o option-name] [--] [arg ...]
./add_speakers.sh: line 4: $'\r': command not found
./add_speakers.sh: line 5: cd: $'.\r': No such file or directory
: No such file or directorycommon_code
./add_speakers.sh: line 7: $'\r': command not found
./add_speakers.sh: line 8: cd: $'.\r': No such file or directory
./add_speakers.sh: line 9: $'\r': command not found
./add_speakers.sh: line 108: syntax error: unexpected end of file
```

I guess this means I'll be hand-crafting the folder structure and the markdown files for the talk abstracts & speaker pages 😭 

At this point, do I have another option that would make my pull request acceptable? e.g. converting the ignite talks to just custom plaintext in the schedule

mattstratton:
I don’t know that the scripts have been tested on windows. 

What i can do for you to help accelerate this for you is create all the files which will just be needing to them have the bios/abstracts pasted in. I’m pretty sure i can figure out the speaker names / talk titles from what you’ve got here. I’ll try to add that into this PR!

aclairefication:
And while I have you here: is there a way to add the background color I used in my 2022-atlanta.yml program custom styling to my program.md (i.e. for the DevOps track and the SRE & Complexity Track - I used workshop for the Security Track)

mattstratton:
We can’t really change overall page background colors on a per-event level. 

I also might be misunderstanding the ask :)

I think the current PR looks good for your MVP and I’ll work on trying to get you most of the way for the full support tomorrow perhaps - in a separate PR. Let me just see what the deploy preview of this latest commit looks like!

aclairefication:
Definitely want to do a separate pull request for the full schedule 👍

mattstratton:
I’ll merge this now - I think I know how to solve your color coding issue on the program.md, but I’m on mobile right now, so I’ll tweak it in the morning. But we can get this merged for tonight!

aclairefication:
🤔  I guess to get that styling in place I'd need to modify the site.css file to add another program element like this?
```
.program-devops-track {
  background-color: #d5c5ef;
  color: #000; }

.program-sre-complexity-track {
  background-color: #8ad0d3;
  color: #000; }
```

aclairefication:
Hooray!

mattstratton:
That would be one way, but we don’t usually add stuff to the css because it’s built from sass, etc. 

I think you should be able to just set style info inline but I’ll need to try it when I’m more awake :)

PR #10977, opened 2022-03-24
[ATL-2022] Schedule content

aclairefication:
Filled in content for the schedule
- Talk abstracts
- Speaker bios

### DIFF
--- a/content/events/2022-atlanta/program/ajuna-kyaruzi.md
+++ b/content/events/2022-atlanta/program/ajuna-kyaruzi.md
@@ -7,4 +7,4 @@ Type = "talk"
 Speakers = ["ajuna-kyaruzi"]
 +++
 
-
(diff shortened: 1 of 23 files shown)


aclairefication:
One thing that puzzles me is sometimes when the schedule page renders it shows the speaker name twice at the beginning of the schedule slot...

Another thing is the links to talks are not showing up on the speaker pages...

Both of these issues were already happening before I added the content to the existing files in the program and speaker folders. 😕

mattstratton:
The program page sometimes acts weird on local builds. 

Take a look at this preview of this PR; it should look correct:

https://deploy-preview-10977--devopsdays-web.netlify.app/events/2022-atlanta/program

The next step (which we can do in a separate PR later) is to add the ignites. Also, when you get speaker headshots we can add those.

PR #11007, opened 2022-03-30
[BHX-2022] add sponsor

Waterdrips:
Add new sponsor, update existing sponsor to correct version!


phrawzty:
A number of the checks choked on this, but the PR looks fine, so 🤷 . Merging.

PR #11011, opened 2022-03-30
[Boise 2022] redirect for 2022

bridgetkromhout:
/cc @ksatirli 

Looks like Boise may have been configured manually instead of with the new-event script. This redirect will point https://devopsdays.org/boise at the current year.


ksatirli:
Great catch @bridgetkromhout - thank you!

PR #11031, opened 2022-04-03
[BHM-AL-2022] Adding Program Schedule and New Sponsor

zonywhoop:
Full program schedule being added and adding Dynatrace as a silver sponsor.


yvovandoorn:
@zonywhoop I noticed JR's last name is fixed in markdown but the actual program file` content/events/2022-birmingham-al/program/jr-baeudoin.md` still contains the typo. Did you want to fix the data file & file name as well?

PR #11033, opened 2022-04-04
[CLT-2022] Update 2022 Charlotte sponsor content

jmaggio:
Updating our sponsor page with this year's content. 

Add Red Ventures as gold sponsor.

Also getting the website logo we have used in the past up. We have a new logo in the works but it might be destined for swag instead of the site.


jmaggio:
doh I just realized I left bronze in, fixing that

PR #11074, opened 2022-04-12
[la 22] add new sponsor logos

irabinovitch:
[la 22] add new sponsor logos

- rootly
- jackpocket
- era


phrawzty:
@irabinovitch A wild merge conflict appears!

image

Did you mean to replace, or was this meant to be an addition?

PR #11086, opened 2022-04-13
Fix cancel event list for events page

mattstratton:
this fixes the /events page to remove canceled events


bridgetkromhout (approved):
Yes, this fixed the events page. (Does seem to have the race condition that sometimes hits the front page; we'll see if that happens in the prod build.)

PR #11112, opened 2022-04-20
[Aar/2022] Update speakers

RandomSort:
Hello, fantastic folx!
This pull request will make two great speakers show up on our website, while removing the speaker that will unfortunately be unavailable. All in all, great stuff :)


RandomSort:
Unsure if the circle CI builds needs a restart 🤔

PR #11115, opened 2022-04-20
[AARHUS-2022] Replace sponsor logo and replace speaker image and bio

jonasvinther:
Replace sponsor logo and replace speaker image and bio


glasnt:
Sponsor is unique to this event, so image can be updated without retroactive changes

PR #11118, opened 2022-04-21
[2022- Ukraine] - Add links, etc, to main website

mattstratton:
This adds the current speakers and info to catch any URLs that index, but also makes more CTA links to the offsite webpage and adds more info!


DivineOps:
Looks good to me!

PR #11143, opened 2022-04-26
[PT-2022] - Changing Julia Lamenza's Bio

julialamenza:
Changing Julia Lamenza's Bio


somatorio:
It seems the error on circleci isn't related to the changes, @mattstratton may you rerun the check?

PR #11183, opened 2022-05-05
[SEA-2022]  speakers, talks and program

harbinger55:
Add the speakers, talks and program for 2022


mattstratton:
You’re missing Kat’s talk - was that an oversight or intentional? It makes the program read incorrectly. 
16C8E076-0985-4898-93F6-E7711E102995

PR #11199, opened 2022-05-09
[DC-2022] Added language and link for CFP

davezen1:
- Added link to the Sessionize CFP and language for the CFP on the welcome.md page


MikeRosTX (changes requested):
Please review suggestions.

PR #11203, opened 2022-05-09
[CHI-2022] updated rebraned sponsor logo

margaretvaltie:
updated seismic logo after rebrand plus new sponsor moogsoft


mattstratton:
Not quite sure what you’re doing with the seismic images?

mattstratton:
Yeah, just changing the images doesn’t do it - they need to be changed everywhere the old one was used!

mattstratton:
You have to run this script too :)

mattstratton:
it looks like seismic has only sponsored Chicago in the past? no need to make the "2021" version of the logo I guess if it only affects Chicago :)

mattstratton:
okay I did a little digging.

seismic is new this year and only touches Chicago 2022. 

@margaretvaltie you can just overwrite the current logo. please remove the "-2021" version from this PR as we don't need it in the repo. :)

mattstratton (changes requested):
Please remove the `2021-seismic.png` file from this PR :)

PR #11222, opened 2022-05-12
[GE-2022] add speaker photo

deniger:
[GE-2022] add speaker photo


deniger:
please do not merge for now as the image is not there :(

PR #11231, opened 2022-05-14
[CLT-2022] organizing team update

jmaggio:
Updating our team for this year to include some new folks and remove folks that are not participating this year.
 Also one last name change.


MikeRosTX (changes requested):
We can get the full names (first, last) from the PR, but not their emails.
Please send an email to the core team listing:

**REMOVE:**
full name - email

**ADD NEW:**
full name - email

MikeRosTX:
> Member

DISREGARD, I found the email. 😉

MikeRosTX (changes requested):
Please review email listing because it does not match this PR.

MikeRosTX (approved):

PR #11270, opened 2022-05-25
[SEA-2022] Add a sponsor-info page

harbinger55:
Add a sponsor-info page


harbinger55:
Unlinked for now. We plan to update it and link it in a few days.

PR #11300, opened 2022-06-02
[DC-2022] Add Spacelift as Platinum sponsor, also fix spacelift data

pburkholder:
Correct case, URL, add Twitter for spacelift.io

Add Spacelift as Platinum sponsor


pburkholder:
@trevorbryant or @gallimorej OK?

PR #11326, opened 2022-06-08
[la 22] add stormforge

irabinovitch:
[la 22] add stormforge

### DIFF
--- a/data/events/2022-los-angeles.yml
+++ b/data/events/2022-los-angeles.yml
@@ -142,6 +142,8 @@ sponsors:
     level: silver
   - id: sonatype
     level: gold
+  - id: stormforge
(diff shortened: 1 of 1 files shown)


data/events/2022-los-angeles.yml
```diff
@@ -142,6 +142,8 @@ sponsors:
     level: silver
   - id: sonatype
     level: gold
+  - id: stormforge
+    level: startup
```
glasnt, line 146: You'll either need to use "bronze" for this level, or change the ID of the startup level (you have id "bronze" labelled as "startup", so you'll need to match the IDs for this to show up. 
```suggestion
    level: bronze
```

PR #11347, opened 2022-06-14
[CAI-22] Synching versions & Updating CFP

aameen79:
Synching versions


glasnt:
Hi! It looks like this PR is from an out of date fork. You'll need to [sync your fork](https://help.github.com/articles/syncing-a-fork/) and update your PR.

aameen79:
Hi @glasnt !
Thank you for your prompt reply
I already did this step, before the PR, and gave me that this fork is 42 commits ahead of main still!
So I created a PR
The only new thing is the last commit, updating the event

PR #11395, opened 2022-06-22
[AMS 2022] Update the workshop from Matt to Grote Zaal

PaulusTM:
Workshop location update


PaulusTM:
Can't move after discussing with @Seth-Karlo

PR #11436, opened 2022-06-29
[CLT-2022] Organizer name update

jmaggio:
Updating an organizer's name. Just sent the email to core with subject Charlotte 2022 organizer update


phrawzty (approved):
I've also adjusted the mailing list (email request verified).

PR #11465, opened 2022-07-06
[RDU-2022] fixed bitdefender logo

normalfaults:
the bitdefender logo was too small. Updated with new logo that displays correctly. (cc @MikeRosTX )


bridgetkromhout:
This is fine to make as a global change. It's the same logo and just removes the useless extra whitespace.

PR #11513, opened 2022-07-17
[MDE-2022] Adding Sponsors and Updating Home

mpinedag12:
[MDE-2022] Adding Sponsors and Updating Home
Removing Call for papers butttong and adding tickets buying buttons


glasnt:
@mpinedag12 Can you please remove the `hugo.exe` from this PR? Thanks!

PR #11547, opened 2022-07-23
Do not show cancelled events in the event lists

don-code:
This change was spawned from a [Slack thread](https://devopsdays.slack.com/archives/C07T0BHHN/p1658539934998329) regarding why some cancelled events show up in the event lists, but some do not. (It was actually due to one event having a start date, and one not, but the bigger issue is that past events aren't respecting the cancel flag.)

Before (note presence of Philadelphia in 2020):
Screenshot from 2022-07-22 21-38-35

After (note absence of Philadelphia from 2020):
Screenshot from 2022-07-22 21-39-37


bridgetkromhout:
Thanks, @don-code!

PR #11565, opened 2022-07-27
[DFW-2022] update particle 41 logo

baldwinjim:
update particle 41 logo


glasnt:
New sponsor, no past sponsorship updates required.

PR #11566, opened 2022-07-27
[BUF-2022] Add JMS as Sponsor and Shift CircleCI

onelioubov:
- Add JMS Tech Solutions as Badge sponsor
- Shift CircleCI from Silver to Captioning sponsor


glasnt:
Presuming you also meant to add mandtbank even though it was part of another commit 👍

onelioubov:
> Presuming you also meant to add mandtbank even though it was part of another commit 👍

I just cleaned up that section a bit and organized it when I added JMS; mandtbank just got moved in that shuffle 🙂 Thanks!

PR #11601, opened 2022-08-02
[KYIV-2022] add the second event in 2022

delgod:
We are organizing the second online event in 2022 🎉


phrawzty (changes requested):
👋 This PR is technically fine, but presents a small taxonomy problem, which is that you are effectively replacing an event that was _already named_ `2022-kyiv`. This does weird things with git history, SEO, and some other meta issues. If you look at the only other examples of this (London and New York from 2013), you'll see that they're named by season.

Basically, _this_ event needs to be named `2022-kyiv-fall` (or `winter`—up to you). Everything else in the PR (including renaming the previous event) is fine.

mattstratton:
I agree with @phrawzty. Please do not over write the previous event :)

delgod:
👍 @mattstratton @phrawzty it is updated

phrawzty (changes requested):
The only thing you need to address now is the locale redirect:
https://github.com/devopsdays/devopsdays-web/blob/f8958804ff2bf5441dd344d7beba815a0a93eab5/static/_redirects#L57-L58
Update that for your `autumn` event and then we'll be good to merge.

phrawzty (approved):

PR #11632, opened 2022-08-09
[CHI-2022] logo updates and lanyard sponsors

margaretvaltie:
updated newrelic and slimai logos after rebrands plus updated our lanyard sponsorship


data/events/2023-salt-lake-city.yml
```diff
@@ -118,7 +118,7 @@ organizer_email: "salt-lake-city@devopsdays.org" # Put your organizer email addr
     #   level: gold
     # - id: instana
     #   level: gold
-    # - id: newrelic
+    # - id: newrelic-before-20220808
```
phrawzty, line 121: Given that this is an event in 202**3**, maybe it should use the new logo as well?

mattstratton: Yeah, it should be going-forward. @margaretvaltie can you revert just this one file change?

margaretvaltie:
in SLC it's in there but commented out 

Screen Shot 2022-08-09 at 8 45 00 AM

data/events/2023-salt-lake-city.yml
margaretvaltie: SLC it's in there but commented out @phrawzty @mattstratton
Screen Shot 2022-08-09 at 8 45 00 AM

also Rundeck is living in the future?

PR #11637, opened 2022-08-10
[DFW-2022] update nobl9 logo

baldwinjim:
update nobl9 logo


phrawzty:
n.b. The other events using this logo are all in the _future_, so no `-before` modification necessary.

PR #11644, opened 2022-08-11
Fix redirects to point to 2022 events

yvovandoorn:
Lots of events must not be using the script to launch a new event because there were quite a few redirects pointing to old events :-).


phrawzty:
@yvovandoorn Somebody with more time and energy than I could envision a way to automate this, haha

PR #11729, opened 2022-08-26
[Montreal-2022]

sebagomez:
Program and speakers added


mattstratton:
Not a blocker on merging, but i notice your welcome page still says `The current CFP round is now closed. We’re reviewing entries` and you might want to update this as you are announcing your program?

sebagomez:
@mattstratton good catch, thank you! it is fixed now

PR #11733, opened 2022-08-28
[LPZ2022] add call for proposals and a sponsor

tebo:
Adding a call for proposals and the Univalle sponsor


MikeRosTX (changes requested):
Waiting to speak with @tebo about renaming the Propose.md page.

MikeRosTX (approved):

PR #11752, opened 2022-08-31
[HOU-2022] Added several sharing images, added 2 speaker headshots, a…

skryukova:
…dded Zoho as Gold sponsor


bridgetkromhout:
The mixed-content error isn't due to anything in this pull request, so I'm going ahead and merging it.

PR #11758, opened 2022-08-31
[Oslo-2022] Add event

larstobi:


yvovandoorn:
Hi @larstobi!

 i’m seeing some organizer changes in your new event. Mind communicating which organizers have left and been added in an email to [info@devopsdays.org](mailto:info@devopsdays.org)?

PR #11783, opened 2022-09-05
[CAI-22] Adding sponsors, with fixes

aameen79:
Adding sponsors, with fixes


data/events/2022-cairo.yml
```diff
@@ -89,6 +89,8 @@ sponsors:
    level: community
  - id: accflex
    level: glod
```
phrawzty, line 91:
```suggestion
   level: gold
```

phrawzty (changes requested):

PR #11800, opened 2022-09-07
Fundamental Change: Enforce lowercase on most PRs. 

yvovandoorn:
This PR forces most of the files stored in the repo to be lowercase. 

* Sponsors have been forced to lowercase, even in acronym situations
* Speakers have had their names changed to lower case, only `2022-prague` 
* All images and files, and files referencing those files, have been updated to use lowercase
* Generally if a file referenced a file stored in upper case or with an upper case, it was touched by `sed` or `tr`.


jmgress:
Thanks, this problem caused me much grief :)

PR #11806, opened 2022-09-07
Make hugo build conditional on linter

mattstratton:
This should make it so the hugo build gh action doesn't run unless the linter is a success


mattstratton:
This isn't ideal - the build step doesn't even show up as an action until the linter succeeds, so in theory you can merge without it? Let's find out.

PR #11811, opened 2022-09-08
Troubleshooting Appveyor PR

yvovandoorn:
Initial PR is to get appveyor to print out which hugo its using then go from there.


mattstratton:
Uh, appveyor doesn’t run anymore at all so ??

PR #11812, opened 2022-09-08
[Oslo-2022] Add CFP end date

larstobi:


mattstratton:
I am pretty sure this will fail since you haven't set the `cfp_date_announce` :)

larstobi:
@mattstratton I have added the announce date now :-)

PR #11826, opened 2022-09-10
[EHV 2022] multiple updates

PaulusTM:
Sponsors:
- update available slots
- Add Stream
- Add Philips
- Add Palo Alto

Misc:
- Updated pronounce


data/sponsors/philips.yml
```diff
@@ -0,0 +1,2 @@
+name: "Philips"
+url: "https://philips.nl/"
```
phrawzty, line 2: I realise that Philips hasn't sponsored a DOD event before, but might it make sense to set this to their generic `.com` site, then use a local override to point it to `.nl` for this event? Just trying to future-proof here (i.e. avoid having to rewrite history…).
```suggestion
url: "https://philips.com/"
```

PR #11860, opened 2022-09-16
[PT-2022] add sponsors, remove org member, and update event info

MiguelAlho:
Changes:
* tomé unfortunately could not continue with us on the team, so removing from the org list
* added solo io with new logo (changed old one to a "-before-<date>" name (please confirm if it the correct way to go
* changes basecone logo
* added venue details (entrance and parking)
* add missing keynote title and description


somatorio:
You have to change the yaml of the events listing spot to use the "before image"... The easiest way to do that is using `utilities/change_sponsor_logo.sh` :)
More info: https://github.com/devopsdays/devopsdays-web/blob/main/utilities/README.md#updating-a-sponsor-image-or-data

MiguelAlho:
thanks @somatorio . I undid my sponsor changes and applied the script. Boston 2019 seems to have been the only one affected.

PR #11866, opened 2022-09-18
Advertise the use of Docker for running Hugo

ghoneycutt:


ghoneycutt:
The docs still shared the old way of installing hugo locally, though the recommended way has long been to use `hugoserver.sh` and run through docker. Without this update, people are liable to waste a lot of time getting hugo going as the docker approach is not documented.

bridgetkromhout:
If I recall correctly, there were recent discussions about questions someone had about the docker approach. Since I haven't tested it with docker, I'll let someone who has tested it vet this update to the docs; thanks.

mattstratton:
I don't know that I agree with this statement:

> though the recommended way has long been to use hugoserver.sh and run through docker.

The mode we are moving towards as the "recommended" way will be gitpod, which isn't fully documented but works the best actually.

CONTRIBUTING.md
```diff
@@ -10,19 +10,37 @@ If you'd like to edit a specific devopsdays event site (and/or contribute code),
...
- - maxOS: `brew install hugo@0.67.1`
- - linux: `brew install hugo@0.67.1`
- - windows: `choco install hugo -confirm --version 0.67.1 --allow-downgrade`
-  
-2. [Fork](https://help.github.com/articles/fork-a-repo/) this repo and clone a copy locally.
+1. Have the ability to run Hugo. This can be done locally or through
+   Docker.
+1. [Fork](https://help.github.com/articles/fork-a-repo/) this repo and clone a copy locally.
+
+#### Run Hugo with Docker (Recommended)
```
mattstratton, line 17: I am not comfortable saying this is "recommended" as it's "unofficially" maintained by us. I do agree with including it like this, and I'm fine making it the "first" choice - I just don't want the word "recommended" on it :)

CONTRIBUTING.md
```diff
@@ -10,19 +10,37 @@ If you'd like to edit a specific devopsdays event site (and/or contribute code),
...
+   - linux: `brew install hugo@0.67.1`
+   - windows: `choco install hugo -confirm --version 0.67.1 --allow-downgrade`
+
 
 ### View site locally
 
 To watch for changes and rebuild on the fly, open a new terminal, change directories to your fork of the repo, and execute the following:
 
 ```
+# Run through Docker (recommended)
```
mattstratton, line 36: same as the previous comment - keep this, but remove "recommended"

mattstratton (changes requested):

CONTRIBUTING.md
ghoneycutt: You have recommended that I use it when I had issues with brew :>

PR #11869, opened 2022-09-18
[INDY-2023] Add sponsorship info

ghoneycutt:
Add sponsorship levels


bridgetkromhout:
(Went ahead and merged, but fyi I think you mean "tablecloths" not "tableclothes".)

PR #11870, opened 2022-09-18
[DSM-2022] Schedule complete!

timgifford:
add speakers and program files
added speaker image placeholders
Files created for schedule


bridgetkromhout:
Ah, you may not want to copy in multiple copies of the default image and list each under the name of the later-to-be-uploaded intended image. If you don't set a speaker image, the default image is used.

Also, the social sharing image can't just point to the headshot of the speaker. Please see https://github.com/devopsdays/devopsdays-web/blob/main/themes/devopsdays-theme/reference.md#social-sharing-image for more details of how that is stored and used.

timgifford:
Thanks for the feedback @bridgetkromhout! I think I got these all addressed. 

I manually created the schedule HTML page. I attempted to use the speaker and program top matter but couldn't figure it out. I think I might need a template.  I didn't know if there is the capability for an event to have its own template. 
Open to learning if you have any references or examples.

yvovandoorn:
@timgifford there are some scripts that help with this in the `utilities` directory. 

That said there are ways to construct enough `awk`, `touch` and `tr` to spit out the file names & contents from a CSV exported from sessionize, papercall or google sheets. Sadly, I don't have a ready made example on this computer.

timgifford:
I'd love to get this iteration merged in. I can look for ways to make it easier for the next iteration.

PR #11899, opened 2022-09-26
[Montreal-2022] New organizer added

sebagomez:


yvovandoorn:
@sebagomez I saw the email exchange between Joost and the montreal group, so an email to info@ is not necessary.

sebagomez:
Cool, thanks @yvovandoorn

PR #11919, opened 2022-09-30
[ALMATY-2022] update sponsor logos

tabu-a:


glasnt:
These sponsors have only been used in this event, so logo edits are A-OK and don't require backporting.

PR #11930, opened 2022-10-04
[Montreal-2022] Elastic requested a different logo

sebagomez:
I used the `change_sponsor_logo.sh` script so everything should be right... right? :)


phrawzty:
😨 
image

PR #11931, opened 2022-10-04
Cancel DevOpsDays Kyiv 2022 Autumn

delgod:
Unfortunately, we don't see good attraction from sponsors for the second event for now.


phrawzty (commented):
This PR is safe to merge however I suggest that you also modify the splat in `static/_redirects` to redirect `/kyiv` to the spring edition.

delgod:
> This PR is safe to merge however I suggest that you also modify the splat in `static/_redirects` to redirect `/kyiv` to the spring edition.

thank you, fixed!

PR #11940, opened 2022-10-07
[CHA-2022] Adding Patriot, Stratusgrid and other updates

asachs01:
Signed-off-by: Aaron Sachs <98627+asachs01@users.noreply.github.com>

Sponsor updates, mainly.


glasnt:
Sponsor logo update is only for resolution, no other visual changes, and appears as a noop in the 2018 sponsor list for the same event. No need to do the update logo dance here.

PR #11944, opened 2022-10-09
[EHV 2022] add ignite slots to schedule

PaulusTM:
Trying to add ignite details to schedule. Hope this works :)


glasnt:
Close! Ignites need to be setup as a separate category of talk listings. 

[Here's an example from another event](https://github.com/devopsdays/devopsdays-web/blob/main/data/events/2022-tampa.yml#L319).

PaulusTM:
@FloorD LGTM

PR #11954, opened 2022-10-11
[AMS-2023] Launching AMS 2023

yvovandoorn:
Initial launch. No new organizers yet (this will change, right @FloorD ;-) ), removed two and those have already been removed from the right places.


FloorD:
I knew I should have just started a Codespace. Will squash later.

PR #11923, opened 2022-10-01
prebuild slices to avoid sorting issue due to invalid date on event

MiguelAlho:
attempt to fix the sorting issue on event listing page

PR #11956, opened 2022-10-12
orders years in list of "other events" in event page header

### LINKED ISSUE #11933: Clean up event dates
On the chapter pages you can navigate to previous years for that city. When you do though the order of the years becomes all mangled and confusing. 

Screenshot_20221004-120920_Chrome (1)

There's ongoing work to fix this, but we would love to actually fix broken date formats and clean up the repo.

**Dates should be reflected like so:**
Unquoted YYYY-MM-DDTHH:MM:SS+TZ:TZ
For example: `2019-01-04T00:00:00+02:00`, or `2019-01-05T23:59:59+02:00`

Please comment with a chapter you found using the wrong date format, and note whether you'd like to pick up the task of changing it..

MiguelAlho:
Fixes #11933 

Similar to https://github.com/devopsdays/devopsdays-web/pull/11923, we need to take into account that `.startdate` might not be defined. This pre-calculates the list of other events to display in the event page welcome block, and correctly sorts the list.

I've used variables `$otherTitle` to determine if `city` or `event_group` is used to match in the filter, and `$other` to hold the list of events to print out the links to.

### DIFF
--- a/themes/devopsdays-theme/layouts/partials/welcome.html
+++ b/themes/devopsdays-theme/layouts/partials/welcome.html
@@ -89,37 +89,35 @@
 
         {{- if $e.city -}}
           {{- $.Scratch.Set "past-counter" 0 -}}
-          {{- range sort $.Site.Data.events "startdate" -}}
-          {{- if .startdate -}}
-          {{- if ne .cancel "true" -}}
-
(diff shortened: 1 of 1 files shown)


FloorD:
Omg yeah!!

PR #11998, opened 2022-10-23
[Montreal-2022] "Dynamic" program added

sebagomez:
We want to have a fast way to update the schedule in case we have some last-minute changes.
We'll roll back to the static schedule after the event is over.

(and sorry, some meaningless commits trying to fix local stuff)

### DIFF
--- a/content/events/2022-montreal/program.md
+++ b/content/events/2022-montreal/program.md
@@ -12,3 +12,6 @@ Icons = "false"
     <hr />
   </div>
 </div>
+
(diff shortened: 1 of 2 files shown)


adrianmoisey:
@sebagomez Do you also need to comment out the static schedule for the 26th October?

This is what I see in the preview, and it seems confusing:
image

Let me know what you want, happy to merge if you want it this way.

sebagomez:
That's weird, I meant to comment out the static part. Thanks for the heads up, I'll fix it

sebagomez:
hey @adrianmoisey I had commented Thrursday out but not Wednesday. Here's the updated version. Thanks

adrianmoisey:
LGTM!

PR #12015, opened 2022-10-27
[theme] Add gtm aka google analytics v4 support

yvovandoorn:
Adds support for Google Tag Manager which is Google Analytics v4.


adrianmoisey:
LGTM

PR #12019, opened 2022-10-28
[Montreal-2022] Post event cleanup

sebagomez:
Removed the registration link and rolled back to the static program.


phrawzty:
This PR is clean and I will merge it; however, if your intent is to remove registration mechanisms, you may wish to remove the registration link from the welcome page in a subsequent PR.

PR #12042, opened 2022-11-16
[CPH-2023] Add CPH 2023 event

jonasvinther:
This PR is adding the CPH 2023 event. More information(dates, price, etc.) will follow soon.


yvovandoorn:
Hi @jonasvinther, could you send the current organizers, including their emails, to info@devopsdays.org. The last  CPH event was 2019 and I see a lot more than 8 organizers in the distribution list. 

Thanks!

jonasvinther:
Hi @yvovandoorn 
I have just updated the organizer list in this PR. It should match the official list now. If it's still inconsistent then let's discuss on the #denmark devopsdays slack channel. We have a shared devopsdays organizer team for Denmark. So every year we switch between two city's: Aarhus and Copenhagen. In 2023 it will be in Copenhagen.
As we do have a shared organizer team in Denmark, it's the denmark list that is the official. And Aarhus + Copenhagen organizer lists should be synced with this.

PR #12057, opened 2022-11-28
[MEL-2023] Create Melbourne 2023 event

DXPetti:
- Performed initial creation of event via create event script
- Updated event basics such as date and team members
- Brought over customizations from 2020-22 pages

Let's finally make this event happen 🎉


yvovandoorn:
Hi @DXPetti! Exciting to see Melbourne organize another event for 2023. 

I see there have been quite some organizer changes. This happens all the time. However could you please email `info@devopsdays.org` with the list of current organizers & email addresses. A member of core can then do some clean up on our end (mailing list removal, Slack access revoked, etc). 

Once that bit of house cleaning is done, we will merge the PR. 

Thanks!

DXPetti:
Thanks for the reminder @yvovandoorn I should know to do this already 🤦‍♂️

Updated team membership sent via email as requested.

Cheers

PR #12092, opened 2022-12-18
[LA-2023] Updating sponsor's logo

omargohan:
**Summary of changes:**

- Updating ctrlstack's logo

**Tested:**

- Locally using Hugo


adrianmoisey:
Since it's only LA-2023 that has ctrlstack as a sponsor, there's no need to maintain the old logo.
I think it's fine to just change the logo, meaning that you can delete `static/img/sponsors/ctrlstack-before-20221218.png`

PR #12128, opened 2023-01-10
Fix Nashville redirect

mattstratton:
The redirect for devopsdays.org/nashville still points at the 2022 site. This fixes that.


glasnt:
TIL splatfile

PR #12137, opened 2023-01-13
[RDU-2023] added alex as an organizer

normalfaults:
[RDU-2023] added alex as an organizer


FloorD:
Since you're adding an organiser, did you email [info@devopsdays.org](mailto:info@devopsdays.org) with the details of who is being added, along with their full name and email address? That way we can update Slack and the mailing list.

PR #12142, opened 2023-01-14
[la 23] more sponsors 

irabinovitch:
[la 23] more sponsors


data/events/2023-los-angeles.yml
```diff
@@ -129,10 +129,22 @@ sponsors:
     level: gold
   - id: port
     level: gold
+  - id: prosperops
+    level: gold
+  - id: mapleops
```
glasnt, line 134:
```suggestion
  - id: maplelabs
```

The ID used needs to match the filename for the yaml and png files that are being added.

PR #12148, opened 2023-01-16
[AMS-2023] Add Lian as an organizer for Amsterdam 🎉

FloorD:
Lian decided to join us as an organizer for Amsterdam - yay! Lian's already been added to all relevant lists and apps.


yvovandoorn:
Added Lian to Slack & distro list

PR #12211, opened 2023-02-02
[MEL-2023] Added agenda for event

DXPetti:
- Added program page from template
- Added program navigation item
- Added basic agenda (using all custom types) minus speaker details

### DIFF
--- a/data/events/2023-melbourne.yml
+++ b/data/events/2023-melbourne.yml
@@ -35,8 +35,8 @@ location_address: "100 St Kilda Rd, Melbourne VIC 3004" #Optional - use the stre
 
 nav_elements: # List of pages you want to show up in the navigation of your page.
   - name: location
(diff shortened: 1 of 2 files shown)


data/events/2023-melbourne.yml
```diff
@@ -38,6 +38,7 @@ nav_elements: # List of pages you want to show up in the navigation of your page
   - name: propose
```
mattstratton, line 38: Maybe remove this line since your CFP is closed?

DXPetti: Done.

Side note, I thought the time/date data for the CFP did this for us based on the comments within the YML. Might be worth adjusting the comment in the template?

mattstratton: It takes off the *button* on the welcome page, but not the nav element

PR #12242, opened 2023-02-09
[CAC-2023] DevOpsDays Caceres 2023 - feb 2023 update

neovasili:
This PR adds more sponsors.

It also opens the registration period plus adding the links to the EventBrite page.

Closes [#7](https://github.com/neovasili/devopsdays-web/issues/7), [#8](https://github.com/neovasili/devopsdays-web/issues/8)


phrawzty:
👋 You're updating two sponsor logos that have been used in the past (by your 2020 event, as it happens). Updating sponsor logos is covered in the [documentation here](https://github.com/devopsdays/devopsdays-web/blob/main/utilities/README.md#updating-a-sponsor-image-or-data). Please follow that procedure the update the logos.

neovasili:
hi @phrawzty,

true, I totally missed it, thanks for noticing, now should be fixed 😄

data/sponsors/wordline.yml
```diff
@@ -1,2 +1,2 @@
-name: wordline
+name: worldline
```
glasnt, line 1: This file is `wordline` and you already have a correct `worldline` (Two "L"s). 
This file won't do anything, and for cleanliness should be removed.

neovasili: Hi @glasnt 

Thanks for the feedback

This is from the 2020 event and while doing the changes, I've noticed the name of the sponsor was wrong, while keeping the files as is I thought to fix the name, should I revert this change?

neovasili: I reverted this change, so shouldn't be an issue any more

PR #12244, opened 2023-02-10
[SEA-2023]  Location updates etc

harbinger55:
- Add organizer Mia Villasenor
- Add location
- Add Dates
- Add "hidden" sponsor-info page

### DIFF
--- a/content/events/2023-seattle/location.md
+++ b/content/events/2023-seattle/location.md
@@ -4,10 +4,15 @@ Type = "event"
 Description = "Location for devopsdays seattle 2023"
 +++
 
-Watch this space for information about the venue including address, map/direction, parking/transit, and any hotel details.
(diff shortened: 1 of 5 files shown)


jyee:
Small nit on this PR, but we should probably update Mia's last name from `Villasenor` to `Villaseñor`

davenash-lytx:
I just copied & pasted it from our slack channel, but I will check in with Mia and update it to the correct character set

villasenor:
Thanks @jyee - either one works! Depending on the system, I sometimes omit it because the first ASCII standard (X3.64-1963) did not have a tilde, and `ñ` was not added officially until Part 1 of ISO/IEC 8859 (extended ASCII) in 1987 so I often worry about support for the character in poorly designed systems. If it's easy to add, that's great, but if not no worries @harbinger55

jyee:
Haha, be like my friend Jérôme and be an agent of chaos!

(small note that he also will insert emojis into his name to try to break systems 😆 )

PR #12251, opened 2023-02-10
[NASH-2023] Powershell version of change sponsor logo

GeoffCraig:
This adds a Powershell version of the change sponsor logo utility.


glasnt:
Thanks for splitting this out, @GeoffCraig!

My Powershell is a bit rusty, but a few things I noticed trying this out on Windows 10: 

 * it should be named `utilities/change_sponsor_logo.ps1` to match the shell utility
 * there are a number of messages that return that don't end the process
 * There's commented out outputs inside inner loops, and trailing debugging

The biggest issue I have is that trying to run this, the output never progresses, and what I thought should be a near instant command seems to never finish. One time after leaving it for a few minutes and trying to cancel, it actually returned the long tail of output showing all the things it had done. I'm not sure if there's output buffering or something that's artificially impeding progress. 

Not sure if we have any more (actively) experienced Powershell devs on core to help out. @mattstratton ?

PR #12293, opened 2023-02-20
[DSM-2023] Add initial event details and organizers

espoelstra:
Venue is reserved, adding event details and updated organizers list.


yvovandoorn:
Hi @espoelstra. Could you confirm the removal of `j******y@gmail.com` (slightly redacted for privacy reasons).

yvovandoorn:
Confirmed via mail, merging.

PR #12197, opened 2023-01-29
[KRAKOW-2023] Add 2023 edition

MilyMilo:
Adding Kraków 2023

PR #12274, opened 2023-02-15
[KRK-2023] Add fixed event data

MilyMilo:
Hi again,

I had to iron out some details mentioned in the previous PR #12197 before submitting. We also had to change the location in the meanwhile, sorry for the delay!

Hopefully this is all right now!

PR #12333, opened 2023-02-28
[KRAKOW-2023] Add Kraków 2023 edition

ivellios:
Added a new edition information and updated the data properly as requested in #12197


ivellios:
It seems I missed one other PR: #12274 which solves the same issue. So, I'm closing this one. Sorry for the mess :)

PR #12355, opened 2023-03-06
[LOGO] Logo small change for Cisco AppDynamics

ghoneycutt:
This slightly changes the color and size as per conversation with AppDynamics. This is not a new logo and does not need the usual new logo process.


ghoneycutt:
They realized the logo looks small on our site and have just changed the color from gray to black to make it more bold.

PR #12372, opened 2023-03-08
list_organizers: Add and display an optional pronouns field

issyl0:
- We allow speakers to show their pronouns, but we don't do the same for organizers. This felt like a gap to me as a new organizer.
- I have confirmed locally that this doesn't break existing organiser listings without pronouns specified.
- Note to self, the `reference.md` tables will be a lot easier to edit in the future if I turn off word wrap in my editor! 😅

This looks like the following:

Screenshot 2023-03-08 at 20 48 43


mattstratton (approved):
Great!!

PR #12399, opened 2023-03-14
[PHX-2023] update dates and links for cfp

desktophero:
After an organizer meeting today, we're pivoting to a new date (+1 month) to give enough time for CFP, logistics, etc., to happen.

This updates links and dates.


mattstratton:
couple things:

the cfp opens in the future; because of how hugo builds, if we merge this now the cfp links won't "automatically" appear on the 16th unless the site is built in prod after the time of the cfp opening. So you might want to do a separate PR for the opening of the CFP and submit that on the 16th

contriwise, since you also should make the "Propose" navigation item un-commented when your CFP opens, you can have this one get merged and then put in another PR to be merged after that date/time with that change. hmu on slack if this doens't make sense :)

yvovandoorn:
unless @desktophero has already reached out over slack @mattstratton i'll just leave this open another 24 hours :-)

mattstratton:
> unless @desktophero has already reached out over slack @mattstratton i'll just leave this open another 24 hours :-)

we should also in that case update the PR to uncomment out the "propose" nav link item

desktophero:
CFP link is live now 🎉 

https://www.papercall.io/dod-phi-2023

desktophero:
@mattstratton … please let me know if there are other items to fix.

thank you!

PR #12408, opened 2023-03-14
[BUF-2023] Updating Bio in Buffalo

tpoland:


yvovandoorn:
Re-running build as Netlify had issues retrieving node packages.

PR #12410, opened 2023-03-15
[MEL-2023] Updating Melbourne program

peterkh:
- updating evening event to when its actually happening! :)
- changing name of pitch sessions


data/events/2023-melbourne.yml
```diff
@@ -211,6 +210,11 @@ program:
     date: 2023-03-16T00:00:00+10:00
     start_time: "16:45"
     end_time: "17:00"
+  - title: "Evening Event @ Ballers Clubhouse"
+    type: workshop
```
glasnt, line 214:
```suggestion
    type: custom
```

Technically not a workshop, also this should fix render errors

peterkh:
thanks @glasnt. the workshop tag was to make it yellow and stand out, but all good just happy it's on the right day now :)

PR #12437, opened 2023-03-20
[CPH-2023] Add program and speakers for Copenhagen 2023

jlandersen:
Add the initial program and speaker data for Copenhagen 2023 🎉


data/events/2023-copenhagen.yml
```diff
@@ -39,8 +39,8 @@ nav_elements: # List of pages you want to show up in the navigation of your page
     url: https://www.eventbrite.dk/e/devopsdays-copenhagen-2023-tickets-516510726487
   - name: sponsor
   - name: location
-  # - name: program
-  # - name: speakers
+  - name: program
+  - name: speakers
   - name: propose
```
mattstratton, line 44:
```suggestion
#  - name: propose
```

I would say remove the propose link from the navigation unless you're still accepting proposals?

mattstratton (changes requested):

data/events/2023-copenhagen.yml
jlandersen: Good point! Removed! Thanks :)

PR #12466, opened 2023-03-24
[NYC-2023] Initial PR to Setup Event

kaerus01:
This is the initial PR for NYC 2023 devopsdays! We have a couple organization membership changes.

PR #12469, opened 2023-03-25
[NYC-2023] Create event

kuritz:
This creates the 2023 event for New York City


kuritz:
This finishes implementing #12466 which can be closed

PR #12493, opened 2023-03-29
[DEN-2023] Remove Hailey from denver organizers

tavisto:
Had an organizer step down, so just need to remove them.


mattstratton:
please send an email to info@devopsdays.org with hailey's info so we can remove them from the mailing list!

PR #12563, opened 2023-04-10
[LDN-2023] Link the 2023 tickets from the homepage

issyl0:
- ~This is draft until someone from the London organizing team gives some form of approval. But i~ It seems logical to me that we do want people to buy tickets without having to trawl the website for the "register" link (which is in the header, but quite small, and much like the "propose" link, people might miss it).


issyl0:
@rjw1 said yes to this on Slack.

PR #12582, opened 2023-04-14
[CPH-2023] Add pagerduty as sponsor

jonasvinther:
Adding pagerduty as gold and community sponsor.

### DIFF
--- a/data/events/2023-copenhagen.yml
+++ b/data/events/2023-copenhagen.yml
@@ -101,6 +101,8 @@ sponsors:
     level: gold
   - id: vmware-tanzu
     level: gold
+  - id: pagerduty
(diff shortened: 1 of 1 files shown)


data/events/2023-copenhagen.yml
```diff
@@ -115,6 +117,8 @@ sponsors:
     level: community
   - id: copenhagen-devops
     level: community
+  - id: pagerduty
+    level: community
```
phrawzty, line 121: You've added this sponsor twice, once in `gold` and again in `community`. Is this on purpose?

jonasvinther: Yes it is.

PR #12608, opened 2023-04-20
[VLN-2023] Add event

turbut:
First event please let us know if we missed something for a start :)
More info coming


phrawzty:
core: contact list confirmed via email

PR #12649, opened 2023-04-29
[LDN-2023] Revise the COVID policy - not checking vaccinations

issyl0:
https://trello.com/c/6XaJcNQt/47-update-covid-policy-not-asking-for-proof-of-vaccination


mattstratton:
FWIW that trello link is not public so for the approvers of the PR it confuses us

issyl0:
Oh, sorry! I wanted the automatic linking on Trello's side, to prove that I did a thing, but now I remember that this isn't a professional setting with backlogs and managers so it doesn't matter.

mattstratton:
It’s fine! I didn’t meant that to sound as annoyed. I was more like “does this matter to me?” 😂😂

PR #12663, opened 2023-05-03
Oslo 2023 small changes site

nerdine:
Changed design of welcome sites and added missing location information.


MikeRosTX (changes requested):
Whatever this file is "/[.vs](https://github.com/devopsdays/devopsdays-web/tree/01fa0e817bfad9eb6db4b74155ee01278c4fa262/.vs)/slnx.sqlite" it needs to be removed.

nerdine:
> Whatever this file is "/[.vs](https://github.com/devopsdays/devopsdays-web/tree/01fa0e817bfad9eb6db4b74155ee01278c4fa262/.vs)/slnx.sqlite" it needs to be removed.

Hi! Thanks for looking over the PR :) The .vs folder is a folder used by visual studio to track changes. It should normally not be tracked by git, and added to the ignore file. It should not affect anything should it be checked in, but for order sake it should (as you say) not be checked in anyway. I though I had deleted all of them, and I hope now I have.

PR #12683, opened 2023-05-09
[LDN-2023] Close down London 2022

chriscoffee:
I'd forgotten to close down the previous year and provide links to the videos and photography.

This PR should suffice in correcting that mistake.

Further to this I also noticed a duplicate key in 2019: `masthead_background` which I've also corrected (though this'll make very little difference to anything other than suppressed warnings).


issyl0 (approved):
Thanks!

PR #12715, opened 2023-05-14
[INFRA] Hugoserver Updates

zonywhoop:
This PR updates `hugoserver.sh` such that you can:
1. Provide the remote url as argument 1, very useful when testing/running on a remote machine
2. Can change the hugo image used by setting `HUGO_IMAGE` in the environment, useful for using say an arm64 image on a macbook or testing a new image
3. Enables `verboseLog` with output to `hugo.log`, useful for troubleshooting missing or mis-rendering information in builds


mattstratton (approved):
This looks good!

PR #12721, opened 2023-05-15
[TPA-2023] - Update Name for James to include mid initial

jmgress:
[TPA-2023] - Update Name for James to include mid initial


jmgress:
Change not needed

PR #12745, opened 2023-05-19
Add warning for large files

kuritz:
### Context

Sometimes large files accidentally get checked into the repo. This inflates the size of the repo quickly, making it take longer to checkout. The guidance to avoid pdfs and other large files is in the contributing guide, but it may be missed or contributors may have read it a long time ago and forget that detail as they become comfortable making contributions. This actually happened to me recently, I'm sure I read the contributing guide before I made my first commit, but I clearly forgot about the restriction on checking in pdfs.

### Analysis

We can try to remind everyone that the contributing guide is always a work in progress and should be reread occasionally to stay up to date by linking it in the PR template.

Additionally, we can add a github action that checks for pdfs specifically, or any file over a certain size, and comments on the PR to alert the contributor and reviewer of the presence of a large file. A comment with a warning is less likely to cause issues when we might actually need to violate this rule, and should be adequate to prevent the accidental cases.

### Solution

- Link directly to the pull request guidelines in the PR template
- Add a new actions workflow that runs on pull_request events, checks for large files (>3MB or pdf) in the diff, and executes a script to comment the PR if one is found.
- Add script that comments a PR one time with octokit.

### Testing

Demo'd on my fork here:

https://github.com/kuritz/devopsdays-web/pull/3


mattstratton:
i assume that once we know this works (and it's merged into main) we should make this a Required check!

.github/workflows/large-files.yml
```diff
@@ -0,0 +1,39 @@
...
+    types:
+      - opened
+      - synchronize
+
+jobs:
+  check_large_files:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
```
adrianmoisey, line 14:
```suggestion
      - name: Checkout repository
        uses: actions/checkout@v3
        with:
            fetch-depth: 0
```

I think this is needed to solve the `fatal: origin/main...HEAD: no merge base` issue.
However, it may make the test take long to run.

I'm thinking that it would be nice to get it working, and we can optimise later.

kuritz: oops, missed that update broke it. it was a permission issue before. hmm if we use the default 2 dot diff I believe it could give errant warnings if the feature branch isn't up to date with main.

That might be preferable to `fetch-depth: 0` though if its rare enough?

kuritz:
merged the suggestion, lets see how long it takes

kuritz:
looks like that's actually faster than needing to run the subsequent `git fetch origin main`

mattstratton:
btw, currently this is throwing this error on the check (I think when it's trying to write the comment on the PR)

```
 message: 'Resource not accessible by integration',
      documentation_url: 'https://docs.github.com/rest/issues/comments#create-an-issue-comment'
```

phrawzty:
Hello,

This PR has been open for… a long time. Is there a reasonable expectation that this will be merged, or can we close it out? @mattstratton @kuritz Thanks!

mattstratton:
> Hello,
> 
> This PR has been open for… a long time. Is there a reasonable expectation that this will be merged, or can we close it out? @mattstratton @kuritz Thanks!

I just put in a change to fix the issue I raised. Not sure if there’s anything else to do.

mattstratton:
oof i think that my fixes should help but it doesn't seem that the action is firing on this PR anymore at all LOL

mattstratton:
ahh i think it runs on the fork; i am going to merge this (it's not a required action) and that should clean up some things and I can do some other testing

PR #12747, opened 2023-05-19
[PHX-2023] lunch sponsor updates

desktophero:
# Changes

- Add Lunch sponsor details
- Add Lunch sponsor image and config


data/sponsors/fastly-aws.yml
```diff
@@ -0,0 +1,3 @@
+name: "Fastly-aws"
+url: ""
+twitter: "fastly-aws"
```
adrianmoisey, line 3: There is no `fastly-aws` account on Twitter, should this line be removed?

mattstratton: to be fair, that field isn't really used by anything (it's just more like a comment) so I wouldn't make this block.

PR #12765, opened 2023-05-22
[INDY-2023] Add speakers and program

ghoneycutt:


data/events/2023-indianapolis.yml
```diff
@@ -120,3 +120,34 @@ sponsor_levels:
...
+  - title: "Preview of speakers - times will change"
+    type: custom
+    date: 2023-10-05
+    start_time: "09:00"
+    end_time: "09:15"
+  - title: "jj-asghar"
+    type: talk
+    date: 2023-10-05
+    start_time: "09:15"
+    end_time: "09:16"
```
glasnt, line 134: I think some of these times might need adjusting.

ghoneycutt: This is done purposely so we people can easily see what the selected talks are so far. Above that it reads "Preview of speakers - times will change"

ghoneycutt:
Thanks @glasnt

PR #12814, opened 2023-06-02
[BHX-2023] Schedule update

Waterdrips:
Adding our 2023 schedule


MikeRosTX (changes requested):
Spot checked and there were 3 or 4 speaker images too big. Make sure to comply with the guidelines below. Also don't know if the jpeg issues is fixed, but to be on the safe side just use jpg and png formats.

https://github.com/devopsdays/devopsdays-web/blob/main/themes/devopsdays-theme/reference.md#speaker-page-fields

The image for the speaker. This image is relative to the static/events/YYYY-CITY/speakers directory. It can be either .png or .jpg. It is recommended to be 600px square.

PR #12827, opened 2023-06-05
[BHX-2023] schedule

Waterdrips:
Finish schedule for BHX after downsizing images.


phrawzty (approved):
The output is clean. Well done! Solid line-up, too. :)

PR #12853, opened 2023-06-08
[AMS-2023] Update the talk from John

PaulusTM:
On request from John there is an other update.


FloorD:
Potential nitpick but: since the talk is already attributed to John, won't "facilitated by John Willis" be a little obvious?

PaulusTM:
Was in doubt about this. Can update if we feel it is.

PR #12870, opened 2023-06-14
[BHX-2023] Fix pre-conf social times

issyl0:
- This now matches the emails that have been going out.

### DIFF
--- a/data/events/2023-birmingham-uk.yml
+++ b/data/events/2023-birmingham-uk.yml
@@ -147,9 +147,9 @@ sponsor_levels:
 program:
   - title: "Pre-conference Social @ Tilt Birmingham"
     type: custom
-    date: 2023-06-14T19:30:00+00:00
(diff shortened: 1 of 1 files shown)


data/events/2023-birmingham-uk.yml
```diff
@@ -147,9 +147,9 @@ sponsor_levels:
 program:
   - title: "Pre-conference Social @ Tilt Birmingham"
     type: custom
-    date: 2023-06-14T19:30:00+00:00
-    start_time: "19:30"
-    end_time: "23:00"
+    date: 2023-06-14T18:00:00+00:00
```
issyl0, line 150: I noticed that all of these timestamps are in +00:00 (UTC), whereas the UK is now in +01:00 (UTC+1) as it's summer. Does this have any impact on the schedule? It seems like `start_time` and `end_time` are used instead, so maybe it's fine.

PR #12893, opened 2023-06-19
[INDY-2023] Allow Indianapolis organizers to edit their site

ghoneycutt:
This patch allows the organizers of DoD Indy to organize pages specific to them. This is needed so that we can do immediate changes of the site for open spaces while also giving autonomy to the local team.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -17,3 +17,5 @@ package.json    @devopsdays/build-infra
 /bin/   @devopsdays/build-infra
 /.github/   @devopsdays/build-infra
 /static/robots.txt @devopsdays/build-infra
+/data/events/2023-indianapolis.yml @ghoneycutt @apoland @napoleonjones @devopsdays/code-reviewers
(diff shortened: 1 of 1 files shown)


mattstratton:
This still would require these people to have commit bits on this repo as I understand it. 

Can you illustrate how this is an issue that needs addressing? I’m leery about changing the security model on this repo if there is another way to solve this. 

Most times we just have event organizers letting us know that they’ll need quick merges on certain days/times and we can be quick to merge them.

mattstratton:
Even if this was something that could be done technically, why only allow this for one city? And it wouldn’t scale well for multiple cities/events.

jyee:
To Matty's point about having folks ready to quickly commit changes: I'll be at DOD Indy and can be available to quickly merge changes for open spaces.

jyee:
Closing this PR as DOD Indy has passed.
There has been some discussion about moving the website to something more accessible than Hugo. I'll make a note of this request as another data point for non-PR gated editing.

PR #12936, opened 2023-06-28
[PHL-2023] Initial Event Creation

dgrizzanti:


MikeRosTX (changes requested):
Since it is your 1st year back after the pandemic and there are some discrepancies in the list of organizers, please send an email to: info@devopsdays.org

Include full names and email addresses for organizers:
- Returning organizers
- New organizers
- and organizers that should be removed

MikeRosTX:
Thanks for list Peter! But something doesn't match, because you say Hannah is Exiting but she is listed as an organizer in the PR.  Please confirm.

dgrizzanti:
@MikeRosTX I’ve removed Hannah from the list

MikeRosTX (approved):

PR #12947, opened 2023-06-30
[ORG-2023] Update schedule

FloorD:
Reflects the starting time of the social event.
Moves introduction of the Open Spaces process to the Introduction. I reviewed the time we have on the schedule to switch from session to session, which seems like enough buffer time. Might need to review this if @jjasghar has wild plans for virtual.

### DIFF
--- a/data/events/2023-organizer-summit.yml
+++ b/data/events/2023-organizer-summit.yml
@@ -201,7 +201,7 @@ program:
     date: 2023-08-08T00:00:00-05:00
     start_time: "08:00"
     end_time: "09:00"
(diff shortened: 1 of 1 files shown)


mattstratton:
@FloorD if the open space process introduction is happening during opening remarks when is the proposal process happening? I think it makes sense to do that after the talks bc people have ideas down open spaces generated from them?

jjasghar:
I'm good with this, my "tutorial"/idea isn't too far from the normal process. it's a lot of "yes and" other stuff. I think I can "prep" 95% of it so people can just _use_ it.

PR #12965, opened 2023-07-05
[BLR-2023] Add DevOpsDays Bengaluru 2023 event

ashwinimanoj:
Adding the event in October. 
We are yet to finalize the org team have included the same set of people who were part of last year's devopsdays bengaluru. 
Will raise another send an email on new organizers being added by the end of the month.


phrawzty:
This is currently failing on a weird warning. 😖 Raising in `#website`.

phrawzty:
Got the OK from @mattstratton; apparently this is a weird bug that pops up when the _fork_ is old, but it doesn't affect deployment.

PR #12971, opened 2023-07-05
[CHI-2023] Sponsor updates

mattstratton:
- Add Sym as live transcription sponsor
- Show transcription sponsorship as sold out
- Remove paypal button and replace with "contact us" for Gold


mattstratton:
something weird with how the transcription sponsor shows up...

PR #12991, opened 2023-07-10
[Montreal-2023] add organiser

sebagomez:
New organizer added


phrawzty:
n.b. email request also received

PR #12994, opened 2023-07-11
[AMS-2024] Updates

PaulusTM:
- Harm is leaving the team for 2024 :(
- Enable the location page


phrawzty:
Please send an email to `info@devopsdays.org` with the requested mailing list change as well. We'll do both at the same time.

FloorD:
> Please send an email to `info@devopsdays.org` with the requested mailing list change as well. We'll do both at the same time.

Done did this.

PR #13002, opened 2023-07-13
[CHI-2023] Update speaker-info.md

christianherro:
Add Corey & practice sessions, update some dates.


glasnt:
@christianherro Looks like there's only updates in this to speaker-info.md, you might have missed a `git add` if you're adding a new file? (going by the PR description you're looking to also add a speaker)

christianherro:
That's correct, only one file modified. The Corey comment was meant for an audience of @mattstratton , there is no additional speaker to add.

PR #13018, opened 2023-07-16
Move archived assets to CDN

don-code:
This change imports assets from archived events (2015-2018) into the CDN. Since they are no longer actively maintained or consumed by Hugo templating, they are not needed in the main site.

Many of these assets (e.g. all of the `@2x.png` files) do not appear to be used on any pages, but I have not made any attempt to remove unreferenced assets.

I've included the script which did this, at `utilities/archive_assets.sh`. Note that I only tested this script under Linux, not Windows or macOS.

There were only two name clashes between the CDN - `sponsorship-packages.png` and `sponsorship-exclusive-packages.png` in the Oslo 2017 event. These appear to be unreferenced, so I've removed them from devopsdays-web, and made no change to the existing ones in devopsdays-assets. The script detects name clashes and forcefully exits, requiring the user to resolve a name clash before trying again.

The script is slower than it could be (~1-2min per event year), because it's n^2 over the number of assets for a given year (it for-each'es over assets in the event, _then_ the event's HTML files for each one). I could fix this, but it didn't seem worth it. I'm happy to fix that if anybody disagrees.

I have only spot-checked some HTML files with `git diff`. I'd like to merge the paired PR in devopsdays-assets, run the site locally off this branch, and spot-check events again, before I give this PR a clean bill of health.


mattstratton:
@don-code do you think we should revisit this? If i'm reading correctly, the proper next step is to merge https://github.com/devopsdays/devopsdays-assets/pull/166 and then have you test via that?

I guess the risk is if we do that merge into devopsdays-assets and it goes sideways, we would have to revert that one, which is probably not awful? thoughts?

don-code:
First of all, thanks for bringing this back up - I've had it in the blocked column on my fridge, and just... not had the fortitude to say, "today's the day I follow up".

Merging the PR on `-assets` _should_ be fairly innocent, since it's just a bunch of adds - the files will sit there and do nothing, until we go point at them from `-web`. So I agree that our first step should be to merge that and let it deploy. Once that's done, I can go spot-check some of the events that were changed.

It looks like the deploy preview for this PR is still running, and while it's broken right now (the assets aren't on the CDN yet), I _think_ it should magically start working as soon as they're present. We've got at least 10 events represented in `#wg-website` in Slack - what do you think of getting them involved, and crowdsourcing the verification? We could draw a line in the sand and say, "At T+7 days, we'll take these changes live. Before then, take five minutes to check your 2018-and-earlier events on this deploy preview site." Separately from verification, that might also be a good way to engage folks who were interested in helping out.

<details>
<summary>[1] seriously, my fridge</summary>


</details>

don-code:
@mattstratton do you want to merge assets this week? It's a quiet week, so I've got some time to test this.

glasnt:
I'm not seeing the utilities script in this PR, but I assume based on what files I can load (my laptop is having a v hard time loading the diff on this one), we're removing old images from this repo and adding them to the other repo, which requires adding the assets URL to make relative links absolute. The asset archive also separates by year, not year-city, so there's a small `s/-/\//` required, but otherwise this makes sense to me. 

If the `utilities/archive_assets.sh` script can be included, that'd be appreciated (if it's already there but I'm not seeing it because of the chonker PR, apologies!)

glasnt:
Based on discussions: the archive util does exist in the PR, github is just truncating to 5,000 files, and `u` comes late in the alphabet 😭 

Tested the archive script on macOS, works fine for me. 

I think this is good to go.

PR #13020, opened 2023-07-17
Remove twitter from the footer

mattstratton:
Kind of messy and maybe could look a little nicer, but since it’s broken anyway…


phrawzty:
In that the Twitter block is no longer present, this PR accomplishes the stated goal.

That said, I wonder if it might look better (for values of) to put the Open CFPs block in the first position, such that it lines up with the All Events column?

phrawzty (approved):

PR #13030, opened 2023-07-19
[ORG] Open Spaces padding

FloorD:
Adds padding for Open Spaces opening - shortening community recs slightly, and the evening event is right across the street, correct? This should work, I'll timekeep like a HAWK.


data/events/2023-organizer-summit.yml
```diff
@@ -261,36 +261,41 @@ program:
...
-    start_time: "16:40"
-    end_time: "17:10"
+    start_time: "17:10"
+    end_time: "17:30"
   - title: "Closing Words"
     type: custom
     date: 2023-08-08T00:00:00-05:00
-    start_time: "17:15"
-    end_time: "17:30"
+    start_time: "17:4git 0"
```
mattstratton, line 297: git?

mattstratton:
i added a small change to the end time of the evening event since we had to update that anyway

data/events/2023-organizer-summit.yml
```diff
@@ -261,38 +261,43 @@ program:
...
```
glasnt, line 297:
```suggestion
    start_time: "17:40"
```

PR #13099, opened 2023-08-07
[INDY-2023] Add community sponsors, IndyAWS and IndyPy

ghoneycutt:


bridgetkromhout:
Hi, @ghoneycutt - if Indy AWS is the same user group you added in https://github.com/devopsdays/devopsdays-web/pull/4794 but with a new logo, please use the [sponsor update process](https://github.com/devopsdays/devopsdays-web/blob/main/utilities/README.md#updating-a-sponsor-image-or-data) instead of adding them as a new distinct sponsor.

ghoneycutt:
Updated to use sponsor update process

PR #13105, opened 2023-08-08
[KYIV-2023] add sponsors and local links

delgod:


data/events/2023-kyiv.yml
```diff
@@ -106,6 +110,10 @@ organizer_email: "kyiv@devopsdays.org" # Put your organizer email address here
 # List all of your sponsors here along with what level of sponsorship they have.
 # Check data/sponsors/ to use sponsors already added by others.
 sponsors:
+  - id: kubecon-2023
+    level: platinum
```
adrianmoisey, line 114: I think this should be KubeCon EU, so as to not conflict if anyone adds KubeCon US

shar1z: Looks fix - I'm going to resolve.

PR #13173, opened 2023-08-22
Include the LinkedIn profile url in the add_speaker script

sebagomez:
I've found less and less speakers provide a twitter handle (for obvious reasons) but everybody has a linkedin profile.

This modification in the add_speaker script (and template) allows to add the LinkedIn profile url while adding a speaker.


FloorD:
Nice! Have you tested that it works? Alternatively, @PaulusTM can we try for Eindhoven if this works?

sebagomez:
@FloorD yes I have, that's how I'm doing it with Montreal :)

PR #13199, opened 2023-08-26
[DC-2023] overwrite Kublr logo with Eastbanc

pburkholder:


MikeRosTX (commented):
If you are replacing the **_kublar_** logo for **_Eastbanc Technologies_** logo, should it be redirected to their website? Otherwise, it's confusing.

https://eastbanctech.com/

pburkholder:
You're right. Let's use #13200 instead

PR #13219, opened 2023-08-30
[DSM-2023] Add MongoDB as a gold sponsor

SonnevilleJ:


FloorD:
More Gold sponsors == more better.

PR #13224, opened 2023-09-01
testing some other netlify plugins

mattstratton:


mattstratton:
okay so this definitely does fingerprint the images and assets; the question is are they set with the proper cache...

mattstratton:
for the record, this broke pretty hard in production (very long build times and it timed out)

PR #13230, opened 2023-09-01
[BOI-2023] updates speaker pictures, schedule and adds sponsor

ksatirli:
Hello organizers 👋🏼 

This PR updates our schedule and adds speaker pictures


jyee:
Looks like there's a rendering error on the schedule page. I'm digging into this to see if we can fix it.

data/events/2023-boise.yml
```diff
@@ -156,21 +191,21 @@ program:
...
+  - title: "Sponsors"
     type: custom
     date: 2023-09-12T08:00:00-07:00
-    start_time: "04:00 PM"
-    end_time: "04:10 PM"
+    start_time: "04:10 PM"
+    end_time: "04:15 PM"
   - title: "tiauna-tia-dodd"
     type: talk
     date: 2023-09-12T19:00:00-07:00
```
jyee, line 206: @ksatirli Can you fix this line from `T19:00` to `T:08:00`? It's causing TiAuna's talk to appear as if it's a separate day.

jyee (changes requested):
One minor nit that's causing a display issue, but should be good to merge after that.

data/events/2023-boise.yml
ksatirli: Great spot @jyee - should be good now!

PR #13248, opened 2023-09-05
[INFRA] Swap out font awesome for local files

mattstratton:
this is a test to try to load font awesome locally files and not use the "kit"


mattstratton:
okay so the way the icons are referred to will need some renaming. for example, mastodon is working but not twitter

mattstratton:
not needed for now, as we are using the font awesome subscription at the moment

PR #13253, opened 2023-09-05
Test hashfiles plugin again

mattstratton:
Originally tested this in #13224 and it was fine until merged to `main` at which point the build in production timed out and took forever.


mattstratton:
I think we determined that we don't need this at all anymore, so I'm closing this PR.

PR #13261, opened 2023-09-06
updated utilities/add_speakers.sh.  Allow / in bio, title, abstract.   update speaker images logic. 

rexroof:
add_speakers:  updated sed delimiter to _, allowing / in bio, title, and abstract.

reworked logic for adding speaker images, letting you pre-load speaker images into static/events folder.

### DIFF
--- a/utilities/add_speakers.sh
+++ b/utilities/add_speakers.sh
@@ -33,12 +33,11 @@ mkdir -p ../static/events/$event_slug/speakers
 # Create empty speakers page file (will be auto-filled for display)
 speakerspage="../content/events/$event_slug/speakers.md"
 cp examples/templates/speakers.md $speakerspage
-sedcmd "s/CITY/$city/" $speakerspage
-sedcmd "s/YYYY/$year/" $speakerspage
+string_replace "CITY" "${city}" "${speakerspage}"
+string_replace "YYYY" "${year}" "${speakerspage}"
 
 # uncomment link to speakers page
-sedcmd "s/#  - name: speakers/  - name: speakers/" ../data/events/$event_slug.yml
-
+string_replace "#  - name: speakers" "  - name: speakers" "../data/events/${event_slug}.yml"
 
 # Prompt for inputting speakers
 while [ 1 ]
@@ -55,15 +54,15 @@ speaker_slug=$(echo $speakername | tr -dc '[:alpha:][:blank:]' | tr '[:upper:]' 
 speakerfile="../content/events/$event_slug/speakers/$speaker_slug.md"
 cp examples/templates/speakers-speaker-full-name.md $speakerfile
 
-sedcmd "s/SPEAKERNAME/$speakername/" $speakerfile
-sedcmd "s/SPEAKERSLUG/$speaker_slug/" $speakerfile
+string_replace "SPEAKERNAME" "${speakername}" "${speakerfile}"
+string_replace "SPEAKERSLUG" "${speaker_slug}" "${speakerfile}"
 
 # twitter handle
 read -p "Enter speaker twitter handle (return for none): " twitter
 [ -z "${twitter}" ] && twitter=''
 # remove @ if they added it
 twitter=$(echo $twitter | sed 's/@//')
-sedcmd "s/SPEAKERTWITTER/$twitter/" $speakerfile
+string_replace "SPEAKERTWITTER" "${twitter}" "${speakerfile}"
 
 #linkedin profile
 read -p "Enter speaker linkedin profile (return for none): " linkedin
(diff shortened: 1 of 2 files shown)


rexroof:
@yvovandoorn @don-code  you two have the most recent commits on this script, adding you for review.

don-code:
Devil's advocate, this just moves the problem to bios and abstracts with underscores, instead of slashes. The Boston event has one of those this year (it's a Wikipedia link, which feels common enough).

Is it reasonable to assume folks have gettext installed? It's a default on RHEL- and Debian-likes, but I can't speak to macOS. We could change the template to something like:
```diff
--- a/utilities/examples/templates/speakers-speaker-full-name.md
+++ b/utilities/examples/templates/speakers-speaker-full-name.md
@@ -1,10 +1,10 @@
 +++
-Title = "SPEAKERNAME"
-Twitter = "SPEAKERTWITTER"
+Title = "$speakername"
+Twitter = "$twitter"
 image = ""
 type = "speaker"
-linktitle = "SPEAKERSLUG"
+linktitle = "$speaker_slug"
 
 +++
 
-SPEAKERBIO
+$bio
```

...then run templating against the whole file, rather than line by line, like:
```
envsubst < utilities/examples/templates/speakers-speaker-full-name.md > $speakerfile
```

I'm happy to take a stab at that if it makes sense.

mattstratton:
> Is it reasonable to assume folks have gettext installed

I would say not; although if it’s something that is easy to install you can specify it as a requirement. But we also do have windows users.

MikeRosTX:
> > Is it reasonable to assume folks have gettext installed
> 
> I would say not; although if it’s something that is easy to install you can specify it as a requirement. But we also do have windows users.

But couldn't Windows users download and use these precompiled binaries 🤔❓

https://mlocati.github.io/articles/gettext-iconv-windows.html

don-code:
Fair. Could we escape the backslashes before feeding to sed ([elder backslashes](https://xkcd.com/1638/) anyone?):

```
sedcmd "s/SPEAKERBIO/$(echo "$bio" | sed 's/\//\\\//g')/" $speakerfile
```

mattstratton:
Overall I do not like having extra pre-requisites for our helper scripts :/

don-code:
Alternatively, just do it in two parts. Take out the `SPEAKERBIO` from the template, run `sed` like normal on the header fields, then:

```
echo $bio >> $speakerfile
```

rexroof:
reading your comments and thinking about it a bit more, I came across a general solution.    I understand what it's doing, though it does appear at first to be line-noise. 

```
# based on https://unix.stackexchange.com/a/129063
string_replace() {
  # requires 3 arguments
  if [[ $# -ne 3 ]] ; then
    echo "string_replace requires 3 arguments:  key to be replaced, string replacement, file"
    exit 1
  fi

  # escape our left hand side, and right hand side of sed args
  escaped_lhs=$(printf '%s\n' "$1" | sed 's:[][\\/.^$*]:\\&:g')
  escaped_rhs=$(printf '%s\n' "$2" | sed 's:[\\/&]:\\&:g;$!s/$/\\/')
  file=$3

  # sed -i'' works for macos and linux equally.
  sed -i'' "s/$escaped_lhs/$escaped_rhs/" "$file"
}

```

I'll push some new commits that throws this into `common_code` and uses it in add_speakers.   we could use it in other scripts, also.


For the `SPEAKERBIO` and `ABSTRACT` maybe a future iteration could accept multi-line inputs and just append the target template file.   They are both often multi-line and appear at the end of their yaml files.

rexroof:
okay, pushed an updated version.   Tested it on this talk, which if you see it in your CFPs, be sure to accept right away:

```
$ cat ../../../content/events/2023-detroit/program/rex-aaaron-roof--esq.md
+++
Talk_date = ""
Talk_start_time = ""
Talk_end_time = ""
Title = "how to give ^%$the worst talks//ever//"
Type = "talk"
Speakers = ["rex-aaaron-roof--esq"]
+++

Spare ribs ribeye ham tenderloin drumstick burgdoggen fatback chuck filet mignon kevin. Boudin frankfurter picanha ribeye. Kielbasa chicken short loin prosciutto porchetta. Kevin cow beef jerky. Pork drumstick tri-tip landjaeger salami.
```

```
$ cat ../../../content/events/2023-detroit/speakers/rex-aaaron-roof--esq.md
+++
Title = "Rex AAARON ROOF // ESQ"
Twitter = "rexcellent"
image = ""
type = "speaker"
linktitle = "rex-aaaron-roof--esq"

+++

this guy is great!!! he's so sweet and _AWESOME_.  "super duper" and also // not / that / great/
```

don-code:
I have no particular issues with this - just want to give my verbal +1 since I can't give an actual +1.

mattstratton:
probably want to see how #13173 affects this script now?

rexroof:
@mattstratton rebased, updated the script to use string_replace for the linkedin profile, ran some tests, it works as expected.

PR #13269, opened 2023-09-07
[Oslo 2023] Add workshop page

troubalex:
Adds a new page to promote the planned workshop ahead of the Oslo 2023 event


glasnt:
Hello! This PR has some from someone who isn't a listed organiser for this event, so I would want to confirm with an Oslo organiser before merging. @larstobi? Can you confirm? Thanks!

larstobi:
@glasnt I can confirm that this is legit. :-)

PR #13277, opened 2023-09-09
[Montreal-2023] DevOpsDays Montreal 2023 has been canceled :(

sebagomez:
We need to cancel DOD MTL 2023 for financial reasons. 
We're already working with the venue to move it to May 2024.


adrianmoisey:
Sorry to hear :(

PR #13283, opened 2023-09-10
Do some 🚮 in this repo

issyl0:
- I spotted some things that could do with 🚮.


issyl0:
```
ERROR 2023/09/10 11:31:46 You have the value 'taxonomy' in the disabledKinds list. In Hugo 0.73.0 we fixed these to be what most people expect (taxonomy and term).
But this also means that your site configuration may not do what you expect. If it is correct, you can suppress this message by following the instructions below.
If you feel that this should not be logged as an ERROR, you can ignore it by adding this to your site config:
ignoreErrors = ["error-disable-taxonomy"]
```

mattstratton:
why did you delete `contributing.json`?

config.toml
```diff
@@ -9,7 +9,7 @@ dataDir = "data"
 archetypedir = "archetypes"
 PaginatePath = "blog"
 buildDrafts = false
-disableKinds = ["taxonomy", "taxonomyTerm"]
+disableKinds = ["taxonomy"]
```
mattstratton, line 12: this change is throwing the error. while we can disable the error, what what the impetus of the change?

issyl0: The fact that when I built the site locally a deprecation warning came up in my terminal about this term.

issyl0: See also the commit message: https://github.com/devopsdays/devopsdays-web/commit/d849ca465e352c4add1ff6158a83d45903eb1e42. It seemed simple enough to fix. How I was wrong... 😆

yvovandoorn:
@issyl0 is still relevant?

issyl0:
If I back out the first commit about `taxonomyTerm` and keep the others, we can still get rid of some unnecessary things. I'll do this later today.

issyl0:
@yvovandoorn @mattstratton Ready for review! Sorry for the delay.

PR #13322, opened 2023-09-18
[CAI-23] Adding sponsors & agenda

aameen79:
Adding sponsors,  adding agenda


aameen79:
We need your support to speed up @MikeRosTX @phrawzty

PR #13352, opened 2023-09-25
[NSH-2024] Adds event and two new sponsors!

GeoffCraig:
We don't have a date, but we have a desire.  Adds 2024 date and adds a fake date to 2022 so that 2022 doesn't show up in the home page.


mattstratton:
you actually want to set 2022 to "canceled" which is what keeps it from showing up on the different places:

https://github.com/devopsdays/devopsdays-web/blob/main/utilities/docs/cancel-event.md

PR #13358, opened 2023-09-27
[NSH-2024] Sets TBD Event for 2024 and adds two new organizers 

GeoffCraig:
This does three things.

Sets up a TBD date, we have the drive, but we don't have a date
Correctly cancels the 2022 event
Adds two new organizers!


adrianmoisey:
When changing organisers, please email info@devopsdays.org with the details of those new organisers along with their email addresses so that we can add them to the mailing list and to slack

GeoffCraig:
> When changing organisers, please email [info@devopsdays.org](mailto:info@devopsdays.org) with the details of those new organisers along with their email addresses so that we can add them to the mailing list and to slack

I sent the email a few days back when I opened the original PR and just resent it.

PR #13379, opened 2023-10-05
[BLR-2023] Update schedule for blr 2023

ashwinimanoj:
- Embded schedule from Sessionize
- Update Venue
- Update Sponsors


content/events/2019-bengaluru/location.md
adrianmoisey, line 1: This is a 2019 file that's changing, is that intentional?

ashwinimanoj: it was accidental, removed it now

PR #13429, opened 2023-10-24
[CAI-24] Adding Cairo 2024 event

aameen79:
Adding Cairo 2024 event


data/events/2024-cairo.yml
```diff
@@ -0,0 +1,126 @@
+name: "2024-cairo" # The name of the event. Four digit year with the city name in lower-case, with no spaces.
```
jyee, line 1: Hi @aameen79 It looks like you probably copied the files from the 2023 event rather than [using the script](https://github.com/devopsdays/devopsdays-web/blob/main/utilities/README.md#events). This is ok, but it means that you also need to update the `_redirects` file manually. Specifically you should [update line 22](https://github.com/devopsdays/devopsdays-web/blob/main/static/_redirects#L22) so that devopsdays.org/cairo redirects to the 2024 event (it's currently pointing to the 2022 event).

jyee (changes requested):
Minor change request to keep the `_redirects` file up to date.

data/events/2024-cairo.yml
aameen79: Hi @jyee ,

Thanks for your feedback. Change done in other pull request

jyee: @aameen79 Thanks. I'll merge both.

For reference, we're merging this with #13434

PR #13443, opened 2023-11-02
[LA-2024] Adding a new startup sponsor

omargohan:
**Summary of changes:**

- Adding a new sponsor (VictoriaMetrics)
- Updating the CFP message

**Tested:**

- Locally using Hugo


jyee:
@omargohan When doing some manual QA, I also noticed that devopsdays.org/los-angeles still redirects to the 2022 event. To update that, you can make a PR that updates this file: https://github.com/devopsdays/devopsdays-web/blob/main/static/_redirects#L63

omargohan:
@jyee, thank you for the heads up. I will update that in a different PR.

PR #13485, opened 2023-11-30
[DEN-2024] Update kubeevents logo to transparent

stealthybox:
This asset is the 3x size from their official media kit.
It is slightly smaller than the current image and slightly larger than its corresponding kubecareers one that's already in our site.

I noticed this in dark mode on my phone.


stealthybox:
deploy preview looking better now.

PR #13519, opened 2023-12-15
[KC-2024] launch of cfp

jerdog:
Add details on our CFP, along with a 302 redirect to the CFP page


static/_redirects
```diff
@@ -52,6 +52,7 @@
 /istanbul/*		/events/2024-istanbul/:splat		302
 /jakarta/*		/events/2022-jakarta/:splat		302
 /juiz-de-fora/*		/events/2024-juiz-de-fora/:splat		302
+/kansas-city/cfp    https://talks.devopsdays.org/devopsdays-kc-2024/    302
```
somatorio, line 55: Just asking, why not add that url at the line 21 of `data/events/2024-kansas-city.yml`? 'Cause that redirect will stay in the next years if you forget to remove it in the future

jerdog: Great question @somatorio - I will need to change the redirect on the line below anyways for next year's event, which is why I decided to use it this way.

mattstratton:
@jerdog you have a merge conflict on the redirects file; please fix it and we can merge!

jerdog:
Yeah, `git pull upstream main --rebase` didn't seem to fix it, but pulling all the changes in manually looks to have @mattstratton?

PR #13531, opened 2023-12-20
[SP-2024] Add redirect to CFP page and rename event folder

strund3r:
Rename event file and folder from `2024-saopaulo` to `2024-sao-paulo` to maintain the same pattern from previous years and to fix the redirect from [devopsdays.org/sao-paulo](https://devopsdays.org/sao-paulo).


gomex (approved):
LGTM

static/_redirects
```diff
@@ -102,6 +102,7 @@
 /santa-maria/*		/events/2018-santa-maria/:splat		302
 /santiago/*		/events/2020-santiago/:splat		302
 /sao-paulo/*		/events/2024-sao-paulo/:splat		302
+/sao-paulo/cfp		https://talks.devopsdays.org/devopsdays-sao-paulo-2024/cfp 302
```
adrianmoisey, line 105: I think order matters here and `/sao-paulo/cfp` needs to be before `/sao-paulo/*`

mattstratton: Just to confirm for future reference, [yes it does](https://docs.netlify.com/routing/redirects/).

mattstratton (changes requested):
Please change the order of the redirect rules so that the /cfp one comes first or it won’t work 

https://github.com/devopsdays/devopsdays-web/pull/13531#discussion_r1433494879

static/_redirects
strund3r: Fixed

PR #13549, opened 2024-01-03
[RDU-2024] Removed CFP Link

jrn00:
Removed CFP link


glasnt:
You'll want to remove the value in `cfp_link` instead of removing the dates to remove the propose nav. You'll also want to close your linked Google form.

jrn00:
> You'll want to remove the value in `cfp_link` instead of removing the dates to remove the propose nav. You'll also want to close your linked Google form.

Thanks! I'll fix those things.

PR #13577, opened 2024-01-13
[LA-2024] Adding sponsor and updating logo

omargohan:
**Summary of changes:**

- Adding SigLens as `gold` sponsor
- Updating Doppler logo as requested by sponsor

**Tested:**

- Locally using Hugo server


adrianmoisey:
Since LA-2024 is the only event using the doppler sponsor, there's no need to retain the old logo, so the "doppler-before-2024011" files aren't needed

omargohan:
@adrianmoisey I'm just following the documentation. Please let me know. Thanks.

adrianmoisey:
> @adrianmoisey I'm just following the documentation. Please let me know. Thanks.
> 
> https://github.com/devopsdays/devopsdays-web/blob/main/utilities/README.md#updating-a-sponsor-image-or-data

The documentation states:
> keep in mind that we don't want to retroactively change history for past events
and also:
> We want to preserve past history before defining a changed default

Since no historical events use this sponsor, there is no history to preserve.

PR #13588, opened 2024-01-17
[minor] Sorting the contents of core.toml

phrawzty:
Personally I think the sort order should be randomised every time we push the site, but hey…

### DIFF
--- a/data/core.toml
+++ b/data/core.toml
@@ -8,7 +8,7 @@ active = [
 "Camilla Martins",
 "Adrian Moisey",
 "Michelle Ribeiro",
-"Sharone Zitzman"
+"Sharone Zitzman",
 ]
 
 advisory = [
@@ -32,7 +32,7 @@ emeritus = [
 "Matthew Jones",
 "Gildas Le Nadan",
 "Stephen Nelson-Smith",
-"Mike Rosado"
+"Mike Rosado",
 "Andrew Clay Shafer",
 "Julian Simpson",
 "Christian Trabold",
--- a/data/core.toml
+++ b/data/core.toml
@@ -16,27 +16,27 @@ advisory = [
 "Serhat Can",
 "Jennifer Davis",
 "Patrick Debois (founder, 2009–2014 lead)",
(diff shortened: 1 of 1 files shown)


data/core.toml
```diff
@@ -16,27 +16,27 @@ advisory = [
 "Serhat Can",
 "Jennifer Davis",
 "Patrick Debois (founder, 2009–2014 lead)",
+"Yvo van Doorn (2021–2023 co-chair)",
 "Nathen Harvey",
 "Bridget Kromhout (2015–2020 lead)",
 "Ken Mugrage",
 "Matty Stratton (2021–2022 co-chair)",
```
adrianmoisey, line 23: Heh, this trailing comma is an issue
```suggestion
"Matty Stratton (2021–2022 co-chair)"
```

data/core.toml
```diff
@@ -16,27 +16,27 @@ advisory = [
...
+"Gildas Le Nadan",
 "Stephen Nelson-Smith",
+"Mike Rosado"
+"Andrew Clay Shafer",
 "Julian Simpson",
 "Christian Trabold",
 "John Vincent",
 "James Wickett",
-"Andrew Clay Shafer",
 "John Willis",
```
adrianmoisey, line 41: This trailing comma is also an issue

```suggestion
"John Willis"
```

phrawzty: I don't think trailing commas matter in TOML?

data/core.toml
phrawzty: Last-line-of-a-list commas matter don't matter in TOML:

https://toml.io/en/v1.0.0#array

data/core.toml
phrawzty: Missing commas definitely matter though, yes ;)

data/core.toml
adrianmoisey: Huh, that's neat. 
I'm not sure I understand why it failed as we've had TOML without the trailing comma

Your solution of adding a trailing comma to all the arrays is a good one

PR #13606, opened 2024-01-21
[DFW-2024] updates to organizers, remove mike and megan

baldwinjim:
updates to organizers, remove mike and megan


phrawzty:
I have removed those two people from the Dallas mailing list as well (luckily their addresses were easy to identify).

PR #13607, opened 2024-01-23
[HOU-2024] Add Jim wang 

mabaums:
Adding Jim Wang to organizers.


glasnt:
Hello! If you can please email core with Jim's email, we can get them added to the mailing list and slack

glasnt:
Email received, organiser list updated!

PR #13616, opened 2024-01-28
[MTL-2024] update sponsership tarifs

polleydium:
I added 6 commits to make the changes a bit more readable for other organizers to review. Can this commit be squashed on merge? Or should I do something different at my end?

Changes:
- reformat sponsor html table
- remove lanyard sponsor
- add sustainable giveaway
- change prices for lunch, cocktail & coffee


glasnt:
We (or at I) don't mind about having separate commits. For complex changes like this you don't have to split things up by step for review if general description will do, but whatever works for your development flow.

PR #13643, opened 2024-02-06
Add description check to PRs

mattstratton:
In theory, this should provide a check to see if pull requests have removed the default text, and it should fail the check if they did not. 

I have now removed the offending text in this description! the PR should now show green.

adding the word Bluth to trigger to check the new message!


mattstratton:
for what it's worth, the check ran (and passed) on the first submission of this PR
image

I'll try to edit the PR to put the default text back (which should trigger the test again) and it should fail it.

mattstratton:
Cool! when the default text is in the PR, it fails a check:

image

image

mattstratton:
ookay I like this now because it also makes the comment and tags the submitter!

.github/workflows/pr.yml
```diff
@@ -0,0 +1,33 @@
...
+            echo "PR description does not have default text removed"
+            exit 1
+          fi
+      - name: Comment on PR
+        id: comment
+        if: failure() && steps.prcheck.outcome == 'failure'
+        uses: thollander/actions-comment-pull-request@v2
+        with:
+          message: |
+            Hi @${{ github.event.pull_request.user.login }}! Please remove the default text from the pull request description!
```
phrawzty, line 32: The functionality is great, though maybe this message doesn't drive the behaviour we want. Maybe something like?
```suggestion
            Hi @${{ github.event.pull_request.user.login }}! It looks like you've left the default text in the pull request description. Please replace that text with something meaningful. Thanks!

```

phrawzty, line 32: The functionality is great, though maybe this message doesn't drive the behaviour we want. Maybe something like?
```suggestion
            Hi @${{ github.event.pull_request.user.login }}! It looks like you've left the default text in the pull request description. Please replace that text with something meaningful. Thanks!

```

phrawzty (changes requested):

.github/workflows/pr.yml
mattstratton: yeah this is better. I'll accept it and test!

mattstratton:
> Hi @mattstratton! It looks like you've left the default text in the pull request description. Please replace that text with something meaningful. Thanks!

@phrawzty tada!

PR #13644, opened 2024-02-06
[LON-2024] redirect /london to the 2024 event

phrawzty:
Change the splat (redirect) for `/london` to point to the 2024 event.


chriscoffee:
Nice spot 👍

PR #13660, opened 2024-02-11
[MSP] Add new organizers

anfleene:
Adding 2 new organizers


phrawzty:
Note: email received, mailing list updated, slack invites sent out.

PR #13663, opened 2024-02-12
[core] Remove misplaced directory

phrawzty:
An update to the ATL-2022 event put this directory here by accident. It's the headshots of the organisers, which are also in the correct place, so these should be removed.


phrawzty:
Oh poop. I mixed up branches here. :(

phrawzty:
Closing this because frankly it's easier than fixing git shit

PR #13664, opened 2024-02-12
[core] Remove misplaced directory

phrawzty:
An update to the ATL-2022 event put this directory here by accident. It's the headshots of the organisers, which are also in the correct place, so these should be removed.

(Repeat of #13663)


phrawzty:
@mattstratton approved the previous iteration of this PR, of which this is a clone, so: merging.

PR #13674, opened 2024-02-14
[LA-2024] Adding everbridge as gold sponsor

omargohan:
**Summary of changes:**

- Adding everbridge as gold sponsor

**Tested:**

- Locally using hugo server


phrawzty:
@omargohan Your PRs are always clean and your descriptions are always appropriate. Just wanted to say thanks—it makes our jobs as maintainers easier. 😄

PR #13675, opened 2024-02-15
[DC-2024] Add two new organizers for DC

gallimorej:
We're adding two new organizers for the DC event this year. Modified DC YAML file. Added their images to the DC organizers directory.

I'll send an email to get them added to our email distro and Slack workspace/channel.


phrawzty:
note: email received; mailing list and Slack updated

PR #13684, opened 2024-02-16
[RDU-24] Update the things

shanedroid:
rdu24

- Added missing speaker profile photos and updated bios
- Updated program
- Updated sponsor logo for [portworx](https://portworx.com/)
- added more sponsors


data/events/2023-denver.yml
```diff
@@ -1,75 +1,75 @@
-name: "2023-denver" # The name of the event. Four digit year with the city name in lower-case, with no spaces.
```
phrawzty, line 1: I'm not entirely sure it seems like this entire file has been changed, when the "modifications" are exactly the same. Weird, but doesn't seem to break anything…

shanedroid: I had unstaged/restaged from IDE and saw this also.. the only intended change was line#91 `portworx-before-20240216` to account for new logo for this sponsor 😅

PR #13687, opened 2024-02-16
[AMS-2024] Add Eti as organizer

FloorD:
Adding Eti as an organizer for Amsterdam. 
- Added her to amsterdam@devopsdays.org
- She is already in Slack, invited her to the Amsterdam channel


phrawzty:
Normally an email with the name and email address of the person in question must be sent to info@devopsdays.org; however, in this case I was able to confirm that this person has been added to the AMS mailing list, so we can merge.

In the future, I encourage you—and everybody—to follow the guidelines set forth for these sorts of modifications. The reason is straightforward: having a paper trail helps to debug problematic situations later.

PR #13693, opened 2024-02-17
matty test PR

mattstratton:
OK, so this is maybe what will happen we existing PRs are updated if the refactor gets merged


mattstratton:
It's claiming a conflict on `data/events/2024-chicago.yml`; i suspect the conflict is that the refactor branch deleted that file so it can't figure it out. Which is what i expected.

PR #13686, opened 2024-02-16
Refactor data directory to use subfolders and update Hugo to 0.121.2

mattstratton:
## _ATTENTION: Do NOT merge this PR until coordinating with Matty, etc._

This adjusts the data directory to use subfolders per event, which will let the YAML files be broken up and worked on individually.

Fixes #13683 
Fixes #13658 
Fixes #13692

### Changes in this PR

- Move data files from `data/events/YYYY-CITY.yml` to `data/events/YYYY/CITY/main.yml`
- Add support to break out portions of the config file into separate files (this change is not made to any existing events; and it's optional). Currently supported are
  - `main.yml`
  - `program.yml`
  - `sponsors.yml`
  - `organizers.yml`
- Update Hugo to `0.121.2`
- Updates the "past events" page to use the same global queries as other event lists, and removes the hardcoded year range (#13692)
- moves a lot of duplicate code into "functions" in some partials, which is definitely making the whole thing a lot DRY-er
- Various internal cosmetic improvements for code readability 
- The "speaking" page now directly links to either the event's propose page or direct CFP link for the listed events 
- update utility scripts for new data file paths, including sample/template for use in scripts 
- Update documentation to refer to the new data file paths

### Steps post-merge

This is what I think is going to need to be done by anyone who has a data file "in flight" for their event, prior to them.

When the commits in this PR are applied, it fundamentally will "delete" any data file in the old format (bc it also creates it, but a commit doesn't say "move this file" it's a "delete and create"). So let's say I have a branch with a modified version of `2024-chicago.yml`. It's not going to take my version of that file and rename it to "main.yml", it's going to delete that file and create a "main.yml" with the contents from the commit in this PR.

now that said, the contents of that data file in this PR at the time of merge will be the same as whatever is currently in `main` (or at least really close to it!). we really really don't want any non-merged PRs hanging around when we merge this one. So we want to close as many PRs as possible prior to the merge. Granted, right now one of the worst offenders is me (the Chicago draft PR) which I can close without merging and I'll keep it in my local branch.

#### Local branches not in a PR

When someone rebases, etc, it might overwrite their changes in their data files. highly recommended to suggest to folks to make a copy of their data file prior to rebasing, just in case they have to recreate those changes. I will test this out myself with my own local branch right after this merges, so I can create some quick "step by step" stuff.

PR #13694, opened 2024-02-17
Update range for past events

mattstratton:
This includes "past" events in 2024


mattstratton:
I have a better fix for this in #13686 so I'll close this one in favor of hopefully merging that one soon.

PR #13707, opened 2024-02-19
Update Bengaluru and Bangalore to the same event group

### LINKED ISSUE #13367: Put Bangalore and Bengaluru into the same event group
Bangalore and Bengaluru are the same city. They switched from the anglicized name to the local name in 2018 and currently it appears as 2 different cities. Bridget has suggested that we group the two using the `event_group` setting in the event yaml.

mattstratton:
fixes #13367


glasnt:
Is the major visibility of this change to have all these events link back to each other in the "Other events" section? 
If so, at the moment this isn't working. 

2022 shows: 2013   2018   2019   2023   

2019 shows: 2011   2012   2014   2015   2016   2017   2018

PR #13712, opened 2024-02-20
Update hugo server command at hugoserver.sh

somatorio:
With the new version hugo server changed some flags, also it will fail if we don't use the image user (circleci) or root


mattstratton:
Good catch! I thought I had found all the places to update :)

mattstratton (approved):

PR #13743, opened 2024-02-27
[IST-2024] Add Speakers

developer-guy:
Added 8 speakers for DevOpsDays Istanbul


utilities/add_speakers.sh
mattstratton, line 1: Good catch!

mattstratton (approved):

PR #13751, opened 2024-02-28
[BER-2024] Adding OSISM.tech as Platinum sponsor

oseemann:


phrawzty (commented):
The sponsor appears to be named "OSISM" but the logo says "REGIO". The osism.tech website doesn't make any mention of regio either. 

Are you sure that this is the correct logo?

oseemann:
The logo is indeed the one I received. The words don't align, but the spirit does - it's about "local cloud" - i.e. promoting a national solution as opposed to the global cloud providers. @phrawzty

PR #13755, opened 2024-02-29
[site] add a null check for PR descriptions

phrawzty:
This extends the PR description check to make sure the description isn't empty.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -26,6 +26,12 @@ jobs:
             echo "PR description does not have default text removed"
             exit 1
           fi
+
+          # Check to make sure the description isn't empty
+          if [[ $PR_DESCRIPTION = "null" ]]; then
+            echo "PR description is empty"
+            exit 1
+          fi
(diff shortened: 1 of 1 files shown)


.github/workflows/pr.yml
```diff
@@ -26,6 +26,12 @@ jobs:
             echo "PR description does not have default text removed"
             exit 1
           fi
+
+          # Check to make sure the description isn't empty
+          if [[ $PR_DESCRIPTION = "null" ]]; then
```
phrawzty, line 31: This doesn't trigger if the PR description is empty, despite:

```shell
$ FOO=`jq -r ".body" out.json`        # empty PR description
$ echo $FOO
null
```

mattstratton: I’ve noticed that GitHub actions don’t always test well in a PR like this… That is the action that you are editing in the PR doesn’t always execute properly in the PR. If that makes sense, this is a small enough thing that I would say we probably just merge it and tested in prod because it’s very easy to Roll back

PR #13777, opened 2024-03-04
[RDU-2024] Speaker Bio Edits

jrn00:
Changed employment blurb for Marino Wijay

Changed title name for Ethan Thomason


jrn00:
Thanks @phrawzty!

PR #13817, opened 2024-03-13
[POR-2024] Create propose section and add new member

joaorafaelpinto:
This PR creates the propose section for 2024 and Pollyana as new member


somatorio:
I have added Pollyanna's email to the city list and sent the invite to slack

PR #13832, opened 2024-03-18
[BER-2024] Add Tigera as Platinum sponsor

oseemann:
Tigera just signed up as Platinum sponsor for the Berlin event.


phrawzty:
Hello!

If you need to change the logo of an existing sponsor—which appears to be the case here—please use the process described here: https://github.com/devopsdays/devopsdays-web/blob/main/utilities/README.md#updating-a-sponsor-image-or-data

tl;dr is that we avoid re-writing history where possible so logos aren't modified in-place.

oseemann:
Alright thanks @phrawzty , I will remove that second commit then and make a separate PR for the logo change.

oseemann:
@phrawzty Can you please reconsider this PR? I have removed the image change from it and meanwhile it got merged separately in https://github.com/devopsdays/devopsdays-web/pull/13833

PR #13845, opened 2024-03-20
[RDU-2024] Replaced Sheng Sheen with Karen Hsu as Ignite Speaker

jrn00:
Added Karen's picture and bio. Removed Sheng's picture and bio.

Karen is giving the same talk so there was no need to change that part. 

Thanks!


jrn00:
Thanks @FloorD!

PR #13886, opened 2024-03-29
[MTL-2024] Program update

sebagomez:
Updated speaker's title and company per his request.
Also updated the program and linked the Google sheet we had in previous events for faster updating. 

After the event the program will be visible and the Google sheet will be gone.


FloorD:
LGTM, the only thing I'd be worried about linking a spreadsheet is prosperity. Will you still be the owner of that spreadsheet _n_ years down the line / how will the program be preserved?

sebagomez:
It's just something to make quick changes right up to the event.
After the event the program will be the static one. We did the same for the 2022 edition

PR #13895, opened 2024-04-02
[SP-2024] Add new platinum sponsor

strund3r:
- Add `Thoughtworks` as platinum sponsor


glasnt:
For reference: Thoughtworks have sponsored in the past and do have sub-brands as existing sponsors, but I can't see anything recent, especially not with this current iteration of the logo.

gomex (approved):

PR #13903, opened 2024-04-04
Allow content reviewers to review common changes

ghoneycutt:
Create a new group by which we can add people who have done this before
and know what content is OK and what needs core to approve.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -20,3 +20,6 @@ package.json    @devopsdays/build-infra
 .gitpod.yml @devopsdays/build-infra
 /.github/workflow/ @devopsdays/build-infra
 .nvmrc  @devopsdays/build-infra
+/data/events/ @devopsdays/content-reviewers
+/data/sponsors/ @devopsdays/content-reviewers
+/content/events/ @devopsdays/content-reviewers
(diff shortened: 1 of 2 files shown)


yvovandoorn:
Seems like a good idea and able to spread the load of PRs. 

I've created the team in Github and given it the relevant permissions (but no members yet). Please review @glasnt, @jyee and @phrawzty.

phrawzty:
Interesting initiative! I'd like to know more about this:
> what content is OK and what needs core to approve.

ghoneycutt:
Adding or removing organizers requires core, though there is already a note about that in the PR template.

phrawzty:
> Adding or removing organizers requires core, though there is already a note about that in the PR template.

Yes, that is one thing. I suppose what I'd like to get from you is a more detailed description of "what content is OK and what needs core to approve" is.

ghoneycutt:
What do you think should be required by core besides adding and removing members?

timothymamo:
Interesting discussion, and would thinking something like this would both help the core team and also get people more involved (not that they're not).
The way I see how this could work, at least initially would be:
* Core team approval is needed for:
  * Adding or removing organizers
  * Initial setup of a DOD conference, even the conference was run in previous years (basically a new directory within `content/events`, `data/events` and `static/events`)
* Non core team approval:
  * Adding files within any of the above, already created, directories
  * Updating/modifying any of the files within any of the above directories, for example adding a new sponsor or updating the bio of a speaker (which I seem to do a fair amount 😆)

I'd suggest that each DOD org team suggest one or two members to be part of this group, with the proviso that they have done it before (could specify a minimum PR count or number of years). If the core team is "hesitant" (that's the word that's coming to mind currently but unsure if it's the correct one) we could always say you need at least 2 reviewers to OK such PRs initially as we test it out. 

To make it clear, the core team does a great job at merging PRs on a regular basis (seems to me this is a task which a core member is in charge of each week). As I said above I think this is more about 1. reducing the load on the core team and 2. getting more people involved.

PR #13929, opened 2024-04-10
Create redirect for Antwerp/Ghent

mattstratton:
Floor asked me to do this


yvovandoorn:
LGTM but YOLO because these are always fun.

PR #13937, opened 2024-04-11
[VLN-2024] Add new sponsor

turbut:
New sponsor


data/sponsors/bankinglab.yml
```diff
@@ -0,0 +1,2 @@
+name: "bankinglab"
+url: "http://bankinglab.com"
```
adrianmoisey, line 2:
```suggestion
url: "https://bankinglab.com/"
```

It's best to avoid the HTTP to HTTPS redirect

turbut: true. uodated now

PR #13948, opened 2024-04-12
[BERLIN-2024] Pretalx => DoD Speaker Auto Update

toshywoshy:
Speaker list updates for 2024-berlin

Auto Release from Pretalx to DoD Speakerlist

***Conference Hook System***

### DIFF
--- a/content/events/2024-berlin/speakers/ccydul.md
+++ b/content/events/2024-berlin/speakers/ccydul.md
@@ -1,0 +1,8 @@
+---
+Title: "Verena Traub"
+image: ""
+type: "speaker"
+linktitle: "ccydul"
+---
+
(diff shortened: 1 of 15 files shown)


toshywoshy:
This is a **TEST**
Please do not merge

content/events/2024-berlin/speakers/ccydul.md
```diff
@@ -0,0 +1,8 @@
+---
```
somatorio, line 1: It would be better if the filename was the speaker name

toshywoshy: Yes it would, but for now I haven't tackled that problem
It needs to be checked properly as we need to decode names to ASCII and we need to prevent duplicates
Idially this logic would also be used for talk titles, as those would also be better to use names in stead of unique identifier
This will be for a later version

PR #13957, opened 2024-04-15
[AMS-2024] Change logo fly.io

FloorD:
Fly.io wasn't happy with the way their logo showed up on the website. I pulled this one from their brand directive.

Screenshot (990)
Not good.

flyio
Good.


phrawzty:
This logo is only referenced twice—both in _future events this year_—so the sponsor logo dance isn't necessary. OK to merge.

PR #13965, opened 2024-04-17
[ATX-2024] Schedule and speakers

nimbinatus:
- feat(schedule): initial build of schedule and speaker wall


yvovandoorn:
This looks great @nimbinatus! I did a bit of testing against the deploy while observing the Pretalx instance and it seems to be quite capable at handing this off to the `gunicorn` process. 

Before I hit the big gree merge button, I also noticed this PR involved the removal of an organizer. Did you want to separate that into its own change or handle it in this. If so, mind sending a mail to info@devopsdays.org (I didn't see one) so we can remove the person from Slack, mailing list, etc.

nimbinatus:
> This looks great @nimbinatus! I did a bit of testing against the deploy while observing the Pretalx instance and it seems to be quite capable at handing this off to the `gunicorn` process. 
> 
> Before I hit the big gree merge button, I also noticed this PR involved the removal of an organizer. Did you want to separate that into its own change or handle it in this. If so, mind sending a mail to info@devopsdays.org (I didn't see one) so we can remove the person from Slack, mailing list, etc. 
> 

Thanks!

Gah, I forgot to revert that. I was experimenting with something. I'll fix it in a couple minutes...

PR #13984, opened 2024-04-21
[KC-2024] Update fly.io sponsor logo

jerdog:
Update fly.io logo at their request


phrawzty:
Logo has not been used in any historical events, so the replacement process doesn't need to be invoked in this case. OK to merge.

PR #13986, opened 2024-04-22
[SITE] add logo rendering

toshywoshy:
Add logo rendering

- use logo.png to generate logo.webp on listing page
- use logo.png to generate logo.webp on welcome page

**This adds the asset folder for hugo rendering and requires hugo-extended**
To avoid any breakage, the old style remains supported and this adds rendering if the file exists

### DIFF
--- a/themes/devopsdays-theme/reference.md
+++ b/themes/devopsdays-theme/reference.md
@@ -420,3 +420,12 @@ or
 ```
 {{< event_map width = "600" height="500" >}}
 ```
+
+### tix
+Embeds the DevOpsDays Pretix ticket and registration system
+This shortcode requires two parameters, city and year, where the city is typically the organiser city and the year the event year.
+Please note this may be differ, please check with in your Pretix for the two names.
+
+```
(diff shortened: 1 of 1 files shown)


mattstratton:
We will need to update documentation on the new Hugo version needed and ensure all scripts that load Hugo in various containers are also updated and tested.

toshywoshy:
OK, added the new logo option to the documentation
The version of hugo defined in the workflow 0.123.3 is sufficient
All scripts are working now and all containers have been updated

toshywoshy:
I tested this now several times, it doesn't break old, and only provides an new option

mattstratton:
@toshywoshy this PR seems to be the code for the tix shortcode now? I think with the force push you did to this source branch it might have lost the other changes?

toshywoshy:
Yes, that was a bad idea, still had the old files in a backup folder.
I created a new branch and enhanced the code a little, including this for all images
So I created a new PR #14341 with more options, closing
Thank you for spotting that @mattstratton

PR #13997, opened 2024-04-25
[VAN-2024] Cancel event

terraboops:
We've had to make the tough decision to cancel our event :(

This PR fixes some outstanding details and cancels the event.


adrianmoisey:
Sorry you had to cancel :/

PR #14006, opened 2024-04-27
[MTL-2024] add Marian as organizer

idcrosby:
Adding Marian as organizer for 2024, info and image


phrawzty:
Thanks for the clean email and PR—highly appreciated. 😄

PR #14011, opened 2024-04-29
[BOS-2024] Add Boston 2024 event

don-code:
Boston is organizing an event on October 21st and 22nd. We're shy on details, but want to get the homepage live.


phrawzty:
@don-code Buddy, this will be the 10-year anniversary of DOD BOS! ✨ 😲 ✨

PR #14014, opened 2024-04-29
[GE-2024] add speaker

deniger:
[GE-2024] add speaker


static/events/2024-geneva/speakers/olivier-boder.jpeg
adrianmoisey, line 1: Could I suggest a smaller JPEG?
This is 8MB, which will result in slower page loads

phrawzty: @deniger Agreed with Adrian, this doesn't need to be print quality ;)

deniger: yes sorry it's done ( png format). Sorry didn't check that before

deniger:
@phrawzty sorry for this point. it's solved now

PR #14023, opened 2024-05-02
[BER-2024] Adds more speaker pics

range:
Adds more speaker pictures


static/events/2024-berlin/speakers/anja-kunkel.png
adrianmoisey, line 1: 2.41MB is pretty large for a speaker image, do you want to resize it?

range: Yes. Also resized two others (too late for git, not too late for delivery) and fixed an issue with the program.

PR #14039, opened 2024-05-06
DEN-2024 update venue, close cfp

themightymuppet:
- updated location page with new venue details
- removed "propose" from nav since CFP is closed.
- updated welcome page with venue details


themightymuppet:
resubmitting

PR #14040, opened 2024-05-06
DEN-2024 Update venue, close cfp

themightymuppet:
- updated location page with new venue details
- removed "propose" from nav since CFP is closed.
- updated welcome page with venue details

### DIFF
--- a/data/events/2024/denver/main.yml
+++ b/data/events/2024/denver/main.yml
@@ -2,7 +2,7 @@ name: "2024-denver" # The name of the event. Four digit year with the city name 
 year: "2024" # The year of the event. Make sure it is in quotes.
 city: "Denver" # The displayed city name of the event. Capitalize it.
 event_twitter: "devopsdaysrox" # Change this to the twitter handle for your event such as devopsdayschi or devopsdaysmsp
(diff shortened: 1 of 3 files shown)


data/events/2024/denver/main.yml
```diff
@@ -40,8 +40,6 @@ nav_elements: # List of pages you want to show up in the navigation of your page
  # - name: program
  # - name: speakers
   - name: sponsor
```
stealthybox, line 42: WDYT?  maybe leave this in as a comment, so it's easy to enable for next year
```suggestion
  - name: sponsor
 # - name: propose
 #   url: https://talks.devopsdays.org/devopsdays-denver-2024/cfp
```

FloorD: I mean, next year you'll create a new event page, or is that not what you mean? Else: changes LGTM

themightymuppet: doesn't the script to generate your new event rebuild all of this? 
also we need to hold on the PR right now anyway because parking details that are included are changing.

adrianmoisey:
Any update here?

themightymuppet:
Hey @FloorD @adrianmoisey sorry for the delay - this is ready for review and merge now.

Thanks!

PR #14052, opened 2024-05-08
[KC-2024] add schedule + speakers from pretix

jerdog:
Attempt to add speakers + schedule from Pretix


jerdog:
This is ready to merge

PR #14139, opened 2024-05-28
[AMS-2024] Adding new pic for Eric D Schabell

timothymamo:
As per title


static/events/2024-amsterdam/speakers/eric-d-schabell.png
adrianmoisey, line 1: This image is 1MB.
I'd recommend a JPEG so it's a bit smaller, but it's entirely up to you

PR #14140, opened 2024-05-28
[London-2024] Add Matt Slane as an organiser

spamwell:
Add Matt Slane as a London organiser


spamwell:
Core have been emailed to let them know about this update.

PR #14153, opened 2024-05-31
Make sponsor logo script compatible with new directory structure

don-code:
The sponsor logo script can no longer find sponsor logo usages under the `data` directory, as of #13686. Running it will produce output like:
```
grep: 2009: Is a directory
grep: 2010: Is a directory
grep: 2011: Is a directory
grep: 2012: Is a directory
grep: 2013: Is a directory
grep: 2014: Is a directory
grep: 2015: Is a directory
grep: 2016: Is a directory
grep: 2017: Is a directory
grep: 2018: Is a directory
grep: 2019: Is a directory
grep: 2020: Is a directory
grep: 2021: Is a directory
grep: 2022: Is a directory
grep: 2023: Is a directory
grep: 2024: Is a directory
grep: 2025: Is a directory
```

It is now looking for all files underneath `data/events` recursively, instead of just directly underneath it.


mattstratton (approved):
This should work just fine - i believe this will also work with the capability of having a `sponsors.yml` broken out data file? (see `data/events/2023/chicago/` for example)

PR #14177, opened 2024-06-07
[AMS-2024] Updating Sysdig logo

timothymamo:
Updating the sysdig logo. I used the script provided and updated previous conferences that refereed to the previous sysdig logo to start using sysdig-before-20240607.png.


FloorD:
Super, thanks!

PR #14179, opened 2024-06-08
[HOU-2024] Added Eric Reeves, last minute replacement speaker

skryukova:
Added Eric Reeves to the program, who stepped in at the last minute as a replacement speaker


adrianmoisey:
Merging this, but just be aware that Stephen is still on the website under the speaker section.

PR #14201, opened 2024-06-14
[London-2024] Updated CoC based on Amsterdam's rather good one

spamwell:
Updating the London code of conduct based off the new Amsterdam one. Imitation, flattery, etc.


issyl0 (approved):
Thank you so much! This is great, way more bounded and less wishy-washy.

PR #14253, opened 2024-06-26
[BHM-AL-2024] Add speakers

ghoneycutt:
Add speakers for Birmingham, AL 2024.


data/events/2024/birmingham-al/main.yml
```diff
@@ -309,3 +309,169 @@ program:
     date: 2024-08-21T00:00:00-05:00
     start_time: "16:00"
     end_time: "17:30"
+program:
```
adrianmoisey, line 312: There are 2 program sections defined here. This new one and the one on line 121.
I assume one of them is incorrect?

ghoneycutt: Thanks for catching this. One of the scripts must have updated this as it was not meant to change.

PR #14282, opened 2024-07-03
[ANTWERPEN-2024] typo

npalm:
Fixing typo


toshywoshy:
We, or Olivier should fix this upstream in Pretalx, otherwise we risk this change being lost in the next release

PR #14303, opened 2024-07-07
[TLV-2024] New Sponsors + Remove Old

shar1z:
Added Coralogix, Cast.AI and Upgraded Big Data Boutique
Updated new Aerospike logo per these guidelines https://github.com/devopsdays/devopsdays-web/blob/main/utilities/README.md#updating-a-sponsor-image-or-data (I hope I did it right)

NOTE: It appears that despite *deleting the entire directory and reforking / rebasing* and starting from scratch - I still could not check at all because there is some kind of syntax error across all of the 2019 event directories (see screenshot). I removed one just to see if localhost will serve, and it didn't so I believe it's a global issue.

Screen Shot 2024-07-07 at 14 10 06

- Removed sponsors who pulled out


mattstratton:
you're getting this error because you're on an old version of hugo. You should be at  0.123.3 (newer versions than that might have issues though)

You can always see the current version we build with here - https://github.com/devopsdays/devopsdays-web/blob/5c62cb508217e5bf5ce6b8c3bf53101c778190be/netlify.toml#L21

mattstratton:
This also didn't actually update any of the other events that had this sponsor; did you run `change_sponsor_logo.sh`?

shar1z:
I did - and it output like a bunch of output so I assumed it worked and then I looked in the sponsor images and it did in fact change the logo name.  

So not sure why this didn't work globally. I will update my Hugo version. I've been avoiding doing that - because historically I've had issues with Apple silicon - but that's probably been updated.

Issue #14322, opened 2024-07-08
Dynamic Sponsor Category Display for DevOpsDays Website


SeifElbayomi:
**Problem:**
The current display of all sponsor categories on the DevOpsDays website, including those without sponsors, may deter potential sponsors due to perceived lack of participation.
Example:
![image](https://github.com/devopsdays/devopsdays-web/assets/59777501/9d6d148b-85fb-4dfa-aa88-0b9bfa2d5a19)

**Suggested Enhancement:**
Implement a backend feature that allows event organizers to choose whether to display empty sponsor categories. This will enable organizers to manage the visibility of sponsor categories dynamically based on their preference instead of commenting it out.
Example:
![image](https://github.com/devopsdays/devopsdays-web/assets/59777501/0b648fb0-8676-4cd0-81c5-dc38df1a00ed)

SeifElbayomi:
Could a maintainer please add the `enhancement` label to this issue?

mattstratton:
Fixed in #14323

PR #14323, opened 2024-07-09
[SITE] add possibility to hide sponsorlevels without sponsors #14322

toshywoshy:
The possiblitiy to add ```sponsors_showempty``` to the data yaml file
This options hides the sponsor levels with no sponsors
See #14322  for more details

### DIFF
--- a/themes/devopsdays-theme/layouts/partials/sponsors.html
+++ b/themes/devopsdays-theme/layouts/partials/sponsors.html
@@ -1,9 +1,37 @@
 {{- $e := partial "functions/get-event-data" . -}}
 
+{{- $levelcount := dict "none" 0 -}}
+{{- if $e.sponsors -}}
+  {{- range $index, $level := $e.sponsor_levels -}}
+    {{- $count := 0 -}}
+    {{- range where $e.sponsors "level" $level.id -}}
+      {{- $count = add $count 1 -}}
+    {{- end -}}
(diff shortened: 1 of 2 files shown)


SeifElbayomi:
I suggest that https://github.com/devopsdays/devopsdays-web/blob/main/utilities/README.md 
Is updated to reference this option and how to use it

mattstratton:
> I suggest that https://github.com/devopsdays/devopsdays-web/blob/main/utilities/README.md Is updated to reference this option and how to use it

I would add it to the [REFERENCE.md](https://github.com/devopsdays/devopsdays-web/blob/main/themes/devopsdays-theme/reference.md#sponsor-levels) section as that details all the fields and this is a new field.

not a bad idea to add it to the overall README but it definitely needs to be documented in REFERENCE

mattstratton (changes requested):
Please update the REFERENCE.md with this new field (specifically in this section): https://github.com/devopsdays/devopsdays-web/blob/34f89e185e9bc2b590b96bdd33ba1cfafad0c747/themes/devopsdays-theme/reference.md?plain=1#L173-L181

toshywoshy:
I do not want to add it to the README in the utilities, as there is no script for this.
This option needs to be set manually, if not specified the default mode is used, displaying all sponsor levels.
I have added the option to REFECENCE below `sponsors_accepted`

mattstratton (approved):

PR #14324, opened 2024-07-09
[BSB-2024] Add logo and background

LORDBABUINO:
Adds logo and background


static/events/2024-brasilia/logo-square.jpg
adrianmoisey, line 1: Would it be possible to make this logo a true square? (Width and Height the same)
It makes the front page look neater.

(we've accidentally let through some logos that aren't square, and the page doesn't look as neat as it could)

nimbinatus (approved):
LGTM. Only change you might want to make depends on whether you want the background image for your banner to have the ground visible so you can view the whole building. Feel free to update that in a future PR if you want.

PR #14345, opened 2024-07-15
[BHM-AL-2024] Add speaker info (#14292)

ghoneycutt:
Add speaker information for Birmingham, AL 2024.


content/events/2024-birmingham-al/speakers/oluebube-princess-egbuna.png
adrianmoisey, line 1: I'm unsure if it's intended or not, but this image is very pixilated

Issue #12515, opened 2023-03-31
Add a "talks" view which shows the talks and titles (with links) 


mattstratton:
right now there is no good way to get a list of all the talks happening; there is the Speakers view which is great, but doesn't do a lot to help show what the content of the event will be.

the Program view is good for this, but that is something you can't usually publish until you have the detailed agenda.

having a view of all the talks (probably at /events/YYYY-CITY/talks) would be really helpful!

please share ideas for designs here since I'm not good at that :)

mattstratton:
This is the "old" speakers page, which might be a bit more of what I'm thinking, but instead of the bio on the right, you have the talk title (with the speaker(s) name underneath) and maybe the description 

<img width="1359" alt="image" src="https://user-images.githubusercontent.com/1280666/229214573-81b0303c-10ee-4043-883d-34bb01e0c814.png">

have to think of how you would display this with talks with multiple speakers

bridgetkromhout:
>  instead of the bio on the right, you have the talk title (with the speaker(s) name underneath) and maybe the description

Some considerations here if you wanted to include the full text of a description would need to be the length of many talk descriptions. To prevent a ton of scrolling, would it be possible to expand/collapse the page? And would the displayed talk description be sourced from the same file as the normal talk file to avoid having two sources of truth? A few things to consider.

mattstratton:
I think we could just do a length cutoff the same we we do for speaker bios (we do this for the bios somewhere and also I do this for a certain view of my podcast Hugo theme )

tbh the title might be enough ?

mattstratton:
I think we could just do a length cutoff the same we we do for speaker bios (we do this for the bios somewhere and also I do this for a certain view of my podcast Hugo theme )

tbh the title might be enough ?

Issue #13697, opened 2024-02-18
Move JavaScript and sass processing out of gulp, and into Hugo pipes


mattstratton:
This should be relatively simple and will make it much easier to make changes to stylesheets etc. 

Biggest thing to watch is if it has any impact on build times.

mattstratton:
ok, did a little more looking; this will require eveyone to have Dart Sass installed just to build the site. That seems like...not something we are okay with.

mattstratton:
as it happens, this isn't as big a deal to do as I thought, so re-opening it as I work on it!

mattstratton:
I have some work on it here - https://github.com/devopsdays/devopsdays-web/tree/mattstratton/pipes

mattstratton:
I've update the netlify plugin config to actually work now, so the netlify builds work fine with dart sass now!

toshywoshy:
I have taking a first set and activated minification in PR #14592 as this is quite simple to do, and we have image pipeing also in PR #14341 .
It would seem that CSS/SCSS/SASS and then JS pipeing would be next to do.

PR #14346, opened 2024-07-15
[SITE] add talk page

toshywoshy:
add a talk page based upon #12515

* all talks, workshops and ignites get displayed on a single page
* this page has the presentation title, the speakers, and when it is happening with the abstract
* all speakers are also displayed, with their picture and bio

### DIFF
--- a/content/events/2024-antwerp/talks.md
+++ b/content/events/2024-antwerp/talks.md
@@ -1,0 +1,6 @@
++++
+Title = "Talks"
+Type = "talks"
+Description = "Talk List for devopsdays Antwerp 2024"
+Icons = "false"
++++
--- a/themes/devopsdays-theme/layouts/talks/single.html
+++ b/themes/devopsdays-theme/layouts/talks/single.html
@@ -1,0 +1,246 @@
+{{ define "main" }}
+
+<h2>Talks</h2>
+
+{{ .Content }}
+
+{{- $.Scratch.Set "first" "true" -}}
+{{- with .Params.icons -}}
+  {{- $.Scratch.Set "icons" (. | upper ) -}}
+{{- end -}}
+
+{{- $e := partial "functions/get-event-data" . -}}
+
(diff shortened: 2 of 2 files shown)


mattstratton:
i think this is a good start! I would want to do a little bit more with the styling on this before merging it (the background color of each item is quite harsh, etc) but i think that with some visual style improvements this is a good one.

camilla-m (approved):
lgtm

nimbinatus (changes requested):
Re styling, I'd like to see more padding in each container (probably 2em) to give more whitespace on the page. Color coding doesn't work well for accessibility reasons, so I'd try to find a different way to indicate type of talk. Italics also create problems for people who have reading issues (e.g., dyslexia). I can try to whip up some CSS for it, but I don't think reusing the styles from the program page will work since those really rely on having everything there without scrolling. If you can hang on for a few days, I'll see what I can mock up.

toshywoshy:
@nimbinatus Yes, this PR is not in a hurry, so if you can provide a mock up or any other additional feedback and information, that would be very helpful

I like the idea of adding  markers, for example some font-awesome icons, for talks, workshops and ignites, however I think we should have those consistent on the program, talk and speaker page.
I personally think having the same colors on all pages is good, this makes it easy to identify the type.
I do understand we might need different colors, but then we should change this on all pages involved and be consistent from now on.

I also think that once we do figure out the move to `hugo-extended` pipelines #13697 we can use SASS/LESS to better create multiple versions, like a high contrast version.

toshywoshy:
@nimbinatus Any news on the mockup or CSS updates?

jerdog:
> @nimbinatus Any news on the mockup or CSS updates?

I will try to take a look at mocking this up.

nimbinatus:
Sorry, folks, dayjob inhaled my time. No updates. I'll be picking this up mid next week unless someone who has some available cycles wants to.

PR #14365, opened 2024-07-17
Add sponsors to program page template 

### LINKED ISSUE #14359: Add sponsors to progam page
In another issue #14352 @mattstratton mentioned that not having the sponsors on the program page is a bug.
I think this is a feature, so I would like to get a discussion on this topic.
While adding the sponsors to the program page should be trivial, and I could implement it and create a PR.
I feel the current state of not having the sponsors and the discussion in issue #14352  iterate reasons why this bug should now be a feature.

mattstratton:
Fixes #14359 

Sponsors have been left off the program page template as an oversight. This brings the program template in line with every other event page template.

### DIFF
--- a/themes/devopsdays-theme/layouts/program/single.html
+++ b/themes/devopsdays-theme/layouts/program/single.html
@@ -222,5 +222,5 @@
 </div>
 </div>
 {{- end -}}
-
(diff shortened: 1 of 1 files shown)


nimbinatus (approved):
LGTM

PR #14368, opened 2024-07-18
[chattanooga-2024] - Update canva.com link

gjohnson:
Use view link instead of editor link for canva.


phrawzty:
This comment changes no files. Please verify.

PR #14371, opened 2024-07-19
[BOS-2024] Health and safety statement

don-code:
Add a health and safety statement for Boston 2024.

I used an awkward workaround to get the title of this page to render as "Health and Safety", rather than "Health-and-Safety". Let me know if there's a better way to do this.


phrawzty:
@don-code That is the way to do this. 👍

PR #14341, opened 2024-07-13
[SITE] Image Rendering

toshywoshy:
This PR renders all images to webp using `hugo` extended features
The codes enables images within the `assets/` folder to be rendered, while mainting the existing `static/` folders.
It converts any image format into webp using compression, and resizing the image.

The following images are rendering
- Event page : we make sure that the image is sized correctly or rescaled to be all the same and solves #14336
- Welcome page
- Sponsor logos
- Speaker page
- Program Talk Page
- Organizers Page

To prove the pages work, I have added 2024-antwerp examples, proving that a mix of `static/` and `assets` folders work, with a preference of using the `assets/` over `static/`

PR #14393, opened 2024-07-24
[ANTWERP-2024] Pretalx => DoD Schedule Auto Update

KrisBuytaert:
Schedule updates for 2024-antwerp

Auto Release from Pretalx to DoD Schedule

***Conference Hook System***


static/events/2024-antwerp/speakers/fpnjcl-img.png
nimbinatus, line 1: This file is really large. Is there some kind of compression going on with the auto-generation system that didn't kick off?

toshywoshy: Well, no, the auto generation system does not process images, it takes the images from Pretalx and uses them as is.
This means that a speaker uploading a large image, will be same image and the same size.
Once #14341 is merged, we can move to the images to the `assets` folder and then HugoIO will process them for the front end, while keeping the source

mattstratton: keep in mind that while it will make the delivered image smaller, it will still be very large in the repo and the repo is already very large and challenging to clone on slower connections, etc. So i'm not sure that fully solves the issue.

Can it be done with the remote image feature of hugo assets, wherein the image is on pretalx side and served from there and doesn't have to be in the repo?

PR #14405, opened 2024-07-29
[SITE] display time in local user time zone

### LINKED ISSUE #9819: Support for time zones for program, events, etc
Localized time zone support is pretty important for virtual events. 

The idea I have is that times/dates should be displayed in the browser's time zone.

For some of the elements, this is not terribly hard; we theoretically capture time for the dates with a TZ flag on some things (CFP start/end, etc) although those are less important.

The program page is where it matters the most. Currently, we don't really have any way that we capture this.

With the theme refactor, we can adjust how the program works, and have a proper timestamp for the elements, but we would have to go back through the previous pages/elements and update them.

We could also make it a "going forward" model, where the "old" layout for program would work, but not display localized, etc. 

The theory of implementation is that this would have to be done somehow in javascript for the frontend; for each time/date display it would be somehow wrapped in some javascript to convert it, I guess.

toshywoshy:
Fixes #9819

- add local user time zone
- add offset from UTC to convert
- convert event time to user time zone
- add several examples to demo

This solves issue #9819 and provides a way to display user local time for the program for `start_time` and `end_time`.
It uses the users browser defined time zone, and provide the timezone name and the time based.
An event needs to add `timeoffset: "+0200"` to indicate the offset of the timezone against UTC.
Please note that this value needs to be quoted as a string.
Here are some examples, which have also added into this commit
- London : +0100
- Rio de Janero : -0300
- Denver : -0600


mattstratton:
>An event needs to add timeoffset: "+0200" to indicate the offset of the timezone against UTC.

What happens with events that do not have the offset?

toshywoshy:
- So if you do not set a timezone or an offset, the old normal progam is displayed.
Check any event in this preview, for example : https://deploy-preview-14405--devopsdays-web.netlify.app/events/2024-organiser-summit/program

- If you do set `timezone` and no `offset`, it will caclulate the offset and set show the timezone after the hours
The example here is "London" : https://deploy-preview-14405--devopsdays-web.netlify.app/events/2024-london/program

- If you do not set `timezone` and only the `offset`, if will just display the local time with no other indication
The example here is "Denver" : https://deploy-preview-14405--devopsdays-web.netlify.app/events/2024-denver/program

- If you do set both `timezone` and `offset`, it will combine all features
The example here is "Rio de Janero"  : https://deploy-preview-14405--devopsdays-web.netlify.app/events/2024-rio-de-janeiro/program

phrawzty:
What happens if there's a summer/winter time shift _during_ an event?

toshywoshy:
If you use `timezone` it will display the summer/winter time correctly, as the offset gets calculated
If you use `offset`, then it will be in summer of winter time based upon the entry and the change will not be automatic

mattstratton:
This seems good! There's a merge conflict that needs resolution @toshywoshy and then I'll merge it 

We do need to update the REFERENCE.md with the new fields too.

mattstratton:
> There's a merge conflict that needs resolution

I think i fixed that :)

themes/devopsdays-theme/reference.md
```diff
@@ -65,6 +65,8 @@ All dates are in unquoted YYYY-MM-DD, like this: `variable: 2016-01-05`, or like
...
 | `enddate`                 | YYYY-MM-DD | No       | The end date of your event. Leave blank if you don't have a venue reserved yet.                                                                                                                                               | 2016-01-05                                            |
+| `timeoffset`                 | YYYY-MM-DD | No       | The offset of the timezone of your event from UTC                                                                                                                                               | "-0600"                                            |
+| `timezone`                 | YYYY-MM-DD | No       | The timezone of the event                                                                                                                                               | "Europe/London"                                            |
```
mattstratton, line 69: @toshywoshy probably good to clarify the acceptable values (point to a website that has the list, etc? not sure where!)

mattstratton (approved):

themes/devopsdays-theme/reference.md
jerdog: The list of acceptable timezone values are here (column `TZ identifier`): https://en.wikipedia.org/wiki/List_of_tz_database_time_zones#List

toshywoshy: Yes, @mattstratton while I presumed we all know the correct TZ identifiers, it needs to be specified in the documentation, as some of them are not always as straightforward
As @jerdog already pointed out Wikipedia has a list of acceptable values.
I have update the reference file accordingly

PR #14423, opened 2024-08-02
[PT-2024] added speaker daniel requena

eduardopiairo:
*  added speaker daniel requena


static/events/2024-portugal/speakers/daniel-requena.jpg
adrianmoisey, line 1: 2.94MB is very large for a speaker picture, could you make it smaller?

PR #14433, opened 2024-08-05
[RIO-2024]Changed Logo Even, Remove Organizers, Add New Sponsor

mauriciopgomes:
Changed Logo DevOps Days Rio
Remove Giovanni Martinss and Paloma Lataliza to Organization
Add New Sponsor (Grafana)

PR #14436, opened 2024-08-06
[site] Fix grafana sponsor object lunacy

phrawzty:
⚠️ PR #14433 needs to be merged _first_ as that's where the `grafana` sponsor object is introduced.

This fixes a longstanding problem with the Grafana sponsor object—specifically, that `grafanabr` should never have been merged in the first place. We're not rewriting history here because the logo and the URL are the same. This PR simplifies the situation overall.


gomex (approved):
LGTM

PR #14372, opened 2024-07-19
[LONDON-2024] Add CAST and Honeycomb as sponsors; alphabetize

issyl0:
Probably useful to review this discarding whitespace (`?w=1`) because I added the sponsors in alphabetical order.

PR #14469, opened 2024-08-13
ci: Ensure sponsors are alphabetical (at the same tier)

issyl0:
- I got bored of manually fixing the 2024 London sponsors data when I noticed it was non-alphabetical and had already been on the website for a bit with the logos not in order ([one](https://github.com/devopsdays/devopsdays-web/pull/14372), [two](https://github.com/devopsdays/devopsdays-web/pull/14434)).
- So here's a minimum viable CI check that runs whenever someone edits an event YAML file.
- It checks that the sponsors are sorted alphabetically by ID at each tier.

Example (events/2024/test/main.yml):

```yaml
sponsors:
  - id: Elephant
    level: platinum
  - id: Aardvark
    level: gold
  - id: Zebra
    level: gold
  - id: Fish
    level: gold
  - id: Cat
    level: silver
  - id: Dog
    level: silver
```

would output:

```text
Gold sponsors in events/2024/test/main.yml are not sorted alphabetically by ID.
```

and so the user would change the "gold" sponsors ordering to the following, to fix the error:

```yaml
  - id: Aardvark
    level: gold
  - id: Fish
    level: gold
  - id: Zebra
    level: gold
```

I considered fixing the ordering in-place and pushing a commit to the user's PR, but fork PRs don't have a lot of permissions and I'm too tired to think about it seriously on top of organizing an actual conference, so that can maybe be a stretch goal for later.

### DIFF
--- a/.github/workflows/alphabetical-sponsors.yml
+++ b/.github/workflows/alphabetical-sponsors.yml
@@ -1,0 +1,61 @@
+name: Check Sponsors are Alphabetical
+
+on:
+  pull_request:
+    paths:
+      - 'data/events/*.yml'
+jobs:
+  check-sorting:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v2
+
+    - name: Install yq
+      run: sudo apt-get install -y yq
+
+    - name: Check if sponsors are sorted
+      run: |
+        #!/bin/bash
+
+        for file in $(git diff --name-only | grep -E 'data/events/*/.*\.yml'); do
+          sponsors=$(yq '.sponsors' $file)
+
+          platinum=$(echo $sponsors | yq '.[] | select(.level == "platinum")')
+          gold=$(echo $sponsors | yq '.[] | select(.level == "gold")')
+          silver=$(echo $sponsors | yq '.[] | select(.level == "silver")')
+          bronze=$(echo $sponsors | yq '.[] | select(.level == "bronze")')
+          community=$(echo $sponsors | yq '.[] | select(.level == "community")')
+
+          platinum_sorted=$(echo $platinum | jq 'sort_by(.id)' 2>/dev/null)
+          gold_sorted=$(echo $gold | jq 'sort_by(.id)' 2>/dev/null)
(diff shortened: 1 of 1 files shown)


.github/workflows/alphabetical-sponsors.yml
```diff
@@ -0,0 +1,61 @@
+name: Check Sponsors are Alphabetical
+
+on:
+  pull_request:
+    paths:
+      - 'data/events/*.yml'
```
issyl0, line 6:
```suggestion
      - 'data/events/*/*.yml'
```

toshywoshy:
While this may be esthecailly correct and good for us as humans, but I do not see the need for us to enfore this.

HugoIO will sort by weight, alphabetically or a ordered, we do not need to that
This is what HugoIO does on the front end for us, as it takes the YAML file and parses that according to our programming.
We use ordered listing, which means that sponsor are listed as per the order in the YAML file.
An event has the option to do alphabetical or FIFO or LIFO.
Eache event is free to chose, we, Belgium prefer FIFO, from left to right, while London may prefer alphabetically and so you can do so.

mattstratton:
I agree… I don’t think that the code should enforce the order of the sponsors, for example, for Chicago, we tend to put them in the order that they paid/signed up as sponsors, and I don’t think that I would want the code to override the order that I wanted

issyl0:
Thanks for the feedback, both!

> We use ordered listing, which means that sponsor are listed as per the order in the YAML file.

Yep, which is why I feel that the order should be alphabetical in the YAML file so that it appears so on the website. My overall thinking for alphabetical was that it "flows" nicely within the tiers, and also that every sponsor within a tier is (theoretically) equal because they paid the same money so alphabetical enforces that in a way.

> Eache event is free to chose, we, Belgium prefer FIFO, from left to right, while London may prefer alphabetically and so you can do so.

I see. I didn't realise that the differences were per-event, I guess I thought people just added new sponsors to the end of the list. Lucky you to not be perturbed by the logos being out of order like I am. ;-)

But really it doesn't matter, it's just a small nit and each event should retain autonomy, as you say, so I'll close this. Thanks again!

PR #14479, opened 2024-08-16
[TPA-2024] Update sponsor information for Tampa 2024 event

jmgress:
[TPA-2024] Update sponsor information for Tampa 2024 event


adrianmoisey:
@jmgress there is a merge conflict here

PR #14481, opened 2024-08-16
FOR-2024: update info from event

ana-aguilar7:
## Context

- Update nutanix logo and fixing save in cloud URL.
- Add CSI as sponsor.
- Add new speaker.


data/sponsors/saveincloud.yml
```diff
@@ -1,2 +1,2 @@
 name: "SaveInCloud"
-url: "https://saveincloud.com/"
+url: "https://saveincloud.com/pt/solucoes-cloud/"
```
adrianmoisey, line 2: Other events also use this sponsor, so I don't think it makes sense to update this file.
You can set the URL of a sponsor in the main.yaml file, over here: https://github.com/devopsdays/devopsdays-web/pull/14481/files#diff-c3b496fc26b5960cb8556fde500515076c5fe030b447309e4be7c57f8d918051R204

ana-aguilar7: fixed!

PR #14488, opened 2024-08-19
[BOS-2024] Add program

don-code:
All speakers here have confirmed that they are speaking. We have not filled the last slot, but would like to put the schedule live while we try to fill it.


don-code:
Hi @mattstratton - I see you added the `do-not-merge` label to this, which is why I suspect it hasn't been merged, but I don't seem to have permission to take the `do-not-merge` label off it. Any chance you could do so, and/or merge it?

PR #14491, opened 2024-08-19
[ANTWERP-2024] Pretalx => DoD Schedule Auto Update

KrisBuytaert:
Schedule updates for 2024-antwerp

Auto Release from Pretalx to DoD Schedule

***Conference Hook System***


toshywoshy:
I think we should keep Sasha on day 1 and use the time on day 2 differently
Move the ignites before lunch and have the open spaces directly after lunch

PR #14500, opened 2024-08-25
[JKT-2018] Rename branding of Abdu and fix broken links

empeje:
I light of my new branding, I'd like to rename my past presentations under my new pseudonym Abdu Códigos. In this PR I rename my presentation so the author reflect my new branding and also fix broken links.

(maintainer edit - removed template so the bot doesn't keep commenting)


nimbinatus:
Hi there! Usually, for historical reasons, we prefer to maintain information, but we want to respect your current identity and do understand that names and psuedonyms are a personalized representation of who you are. As your wording identifies this as a branding change (a pseudonym rather than a name change), would you be open to a compromise in this situation of including your name along with your new pseudonym?

jerdog:
@empeje is this still a change you want to make? If so, can you provide an answer to Laura's question above?

empeje:
@jerdog and @nimbinatus I think I cancel my plan. I'm closing my PR.

PR #14525, opened 2024-09-01
[2024-LDN] Gold, Silver and Bronze sponsorship changes

issyl0:
- Zoho Group have sponsored us under their Site24x7 observability brand.
- Google Cloud are, once again, a Bronze sponsor.
- Chainguard have returned as a Silver sponsor.

According to screenshots on Slack, @chriscoffee had a branch for this already but I can’t find it so maybe it’s in his personal fork.


data/events/2024/london/main.yml
```diff
@@ -121,7 +121,7 @@ sponsors:
     level: gold
   - id: heroku
     level: gold
-  - id: zoho
+  - id: site24x7
```
issyl0, line 124: Let’s see what colour the logo is on the preview, because we appear to have both a green version and a black version and are not sure which one is right.

issyl0: This version is green. They sent us the black version (and we've got that on our printed banners). So we'll have to do the logo shuffle here.

PR #14589, opened 2024-09-17
[CAI-24] Adding Sponsors

aameen79:
Adding Sponsors


data/events/2024/cairo/main.yml
```diff
@@ -167,13 +171,13 @@ sponsor_levels:
     max: 1 # to prevent "join as" string from appearing
   - id: corporate
     label: "Corporate Organizers"
-    max: 1 # to prevent "join as" string from appearing
+    max: 2 # to prevent "join as" string from appearing
```
phrawzty, line 174: Strictly speaking, I don't think it's necessary to increment this—as long as it's more than 0 and less than the actual number of items in the category, the string won't appear.

Not a blocker—just information. :)

phrawzty (approved):

aameen79:
Yes I understand, I just wanted the section title to be “Sponsors” instead of “Sponsor”, I left it “1” in the Governmental section as it’s actually only one Governmental Organizer.
Thank you

PR #14592, opened 2024-09-17
[SITE] move to hugo builtin minify

toshywoshy:
* move to use hugo-extened minify function
* add `rawhtml` shortcode for non-interpreated html/js code this provides the ability to have real raw code that hugo does nto interpretate
* move several markdown content with mixed html/js to raw to prevent minify failure
    - content/events/2019-geneva/newsletter.md
    - content/events/2023-detroit/welcome.md
* remove unused file to prevent wrong interpretation
    - content/events/2024-kansas-city/program/placeholder-ignite.md
* remove nodejs and gulp from gh action workflow
* add configuration for proper minifying
* update workflow to use new hugo options syntax to enable minify


mattstratton:
one thing of note - the minify that happens in the pipeline is not actually minifying the deploys, that was basically a cheap and easy way for me to make sure that people had valid HTML in their files :)

toshywoshy:
Yes, however with the current Hugo that would not be required anymore, in the past Hugo would not check HTML, but now it checks HTML files it processes.

mattstratton (approved):

PR #14601, opened 2024-09-19
[DEN-2025] update organizer list and add 2025 placeholder

themightymuppet:
I've also sent an email to info@devopsdays.org with the updates to our denver distro - thank you!


adrianmoisey:
Don't forget to update the redirect: https://github.com/devopsdays/devopsdays-web/blob/ba322b6e6d7a0be46cb2b3aac145e0d5724260b4/static/_redirects#L36

data/events/2024/denver/main.yml
adrianmoisey, line 1: Is this 2024 file supposed to be updated?

themightymuppet: Yes 2024 should’ve been updated a while ago, it’s been out of date for a while

mattstratton:
I've updated the mailing list so this can be merged once this change has been made @themightymuppet 

> Don't forget to update the redirect:
> 
> https://github.com/devopsdays/devopsdays-web/blob/ba322b6e6d7a0be46cb2b3aac145e0d5724260b4/static/_redirects#L36

mattstratton:
I was going to fix a bunch of things in this for you, but you don't have "allow edits from maintainers" so I can't :/

themightymuppet:
sorry I missed the comment about updating redirects! I just pushed that change. 

Also I didn't see an option to allow edits from maintainers 😭 looks like that might be a per-pr option and I wonder if it needs to be off a branch other than main? I could add core to our org/repo if y'all would prefer to have access? 

thanks for the help!

mattstratton:
It should be set as the default (allow edits from maintainers). I think maybe the first time you set it, it will work on each PR after? Most PRs we get have it set and I’m pretty sure folks aren’t checking that box each time?

https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests

mattstratton:
It’s also possibly because you submitted the PR from main - probably works better on a branch 

no need to give us access and honestly it would be a pain since you can’t use a group etc :)

mattstratton (approved):

PR #14621, opened 2024-09-27
[ZRH-2025] Add sponsoring prospectus for 2025

romanoroth:
Added the sponsoring page. 

ATTENTION: You also need to approve https://github.com/devopsdays/devopsdays-assets/pull/214 these two pull requests depend on each other

### DIFF
--- a/content/events/2025-zurich/sponsor.md
+++ b/content/events/2025-zurich/sponsor.md
@@ -1,0 +1,252 @@
++++
+Title = "Sponsor"
+Type = "event"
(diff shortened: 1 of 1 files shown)


phrawzty (approved):
The sponsor page itself is totally fine, so there are no blockers to merging this PR; however, it's not _linked_ anywhere, so for now the sponsor page is an orphan. You'll want to add / activate links in a future PR, assuming you want this page to be easily accessible. :)

PR #14625, opened 2024-09-27
[SITE] Remove duplicate sponsor images and move remaining sponsors to assets

toshywoshy:
- Remove all duplicate static sponsor imsages which exists in the asset folder
- Move remaining static sponsor files to the asset folder
- The images still in the static folder give problems, and remain there for now
This would save about 200MB for the sponsor logos


mattstratton:
Moving the sponsor images out of the static directory will break all the "archieved" events (the ones with the fully generated HTML in the static dir) as they have `/img/sponsors/saucelabs.png` (for example) hardcoded (and won't process with assets)

You can see it for example at https://deploy-preview-14625--devopsdays-web.netlify.app/events/2018-chicago/welcome/ (see all the broken sponsor images)

mattstratton (changes requested):
As I mentioned in the thread, we cannot remove sponsors from static bc of the generated archived events.

toshywoshy:
OK, so closing this for now. as I do not see any other solution

don-code:
Any thoughts on having some CI step that copies images into the right place? What if we took all of the files out of `assets/img/` and copied them into `static/img/` at build time, like:
```
find assets/sponsors/ | xargs -i cp {} static/img/sponsors/
```

That would at least prevent having to keep two copies of the sponsor images. The downside is that the archived events won't be usable in local builds (I guess we could to the same in hugoserver.sh), but this doesn't seem like much of a downside, since nobody should be changing those anyway.

PR #14620, opened 2024-09-27
[CHI-2025] Adding Nvisia's Updated Logo and Gold Sponsorship

albert-cheung:
Adding Nvisia's Updated Logo and Gold Sponsorship

PR #14628, opened 2024-09-28
[CHI-2025] Adding nvisia updated logo and KubeEvents

albert-cheung:
Adding nvisia updated logo and KubeEvents


don-code:
Per prior conversation in #14620, I verified that both the current and prior logos look as expected.

don-code (approved):

PR #14634, opened 2024-09-30
[MTL-2024] Links to videos

sebagomez:
Links to recording added on each presentation's page


don-code:
I spot-checked a few of these, and confirmed that they are working.

don-code (approved):

PR #14656, opened 2024-10-04
Fix issue with organizer and speaker images

mattstratton:
Fixes a bug with the speaker and organizer images


nimbinatus (approved):
LGTM; will merge when preview build completes

PR #14662, opened 2024-10-07
[SSA-2024] add logo and location, propose and registration nav_elements

somatorio:
Add logo and location, propose and registration navigation elements to 2024 Salvador


don-code:
Hi @somatorio, were the speaker photos added in #14547 an accident? I just want to double-check that you want them removed.

toshywoshy (approved):
Yes, the speaker images are a copy from 2019, also with the new updates, we want them in assets as was done with the logo

PR #14670, opened 2024-10-08
[SITE] update google analytics

toshywoshy:
* this updates the google analytics config
* this also stop using the builtin depricated key

This is required from hugo 0.135.x and higher and should solve the deprication problem


jerdog:
Even though I am not a `code-reviewes` this gets a 👍🏻  from me

nimbinatus (approved):

PR #14672, opened 2024-10-10
[FLN-2024] Add program and speakers

jeanmorais:
Add the program and speakers profile.


toshywoshy (changes requested):
You need to move all the speaker images to the assets folder
That would be `assets/events/2024-florianopolis/speakers/`

toshywoshy (approved):

PR #14832, opened 2025-01-25
[CHI-2025] Enable 2025 volunteer page

brian-andersen:
This PR:

* Enables the 2025 volunteers page

### DIFF
--- a/content/events/2025-chicago/volunteer.md
+++ b/content/events/2025-chicago/volunteer.md
@@ -1,0 +1,22 @@
++++
+date = "2016-10-17T08:26:34-05:00"
+title = "volunteer"
+type = "event"
+description = "Sign up to volunteer at devopsdays Chicago 2025!"
+
++++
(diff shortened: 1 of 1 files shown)


toshywoshy (changes requested):
Where does it get linked from?
I do not see any navbar reference or anywhere else anything.

brian-andersen:
> Where does it get linked from? I do not see any navbar reference or anywhere else anything.

We don't link it from anywhere on the site, the URL is sent in our email soliciting volunteers so that they are routed to the DevOpsDays site rather than directly to a Google Form.  If you check previous years, it's the same.

toshywoshy (approved):
OK, I see, I feel the `iframe` should become a shortcode, but until someone, like myself creates one, this is good to go

brian-andersen:
> OK, I see, I feel the `iframe` should become a shortcode, but until someone, like myself creates one, this is good to go

No worries.  It's copy and pasted directly from the Google Form and I didn't want to risk it not working.

PR #14847, opened 2025-02-02
[Aarhus-2025] add sponsor knowit

danquah:
Updated the logo for Knowit to the latest design sent to us by representative from Knowit.

Then added them as sponsors for Aarhus 2025


phrawzty (approved):
I only checked two of the oslo-206 large diffs manually, but they seemed fine, and a quick visual inspection of that event in the preview seemed correct as well.

PR #14855, opened 2025-02-05
[CHI-2025] Remove speaker

mattstratton:
Kedasha is no longer a confirmed speaker


phrawzty:
@mattstratton The PR refers to a speaker being removed, but the entry for "ignite karaoke" is also removed. Just want to confirm that this is correct before merge.

mattstratton:
> @mattstratton The PR refers to a speaker being removed, but the entry for "ignite karaoke" is also removed. Just want to confirm that this is correct before merge.

Yep. I had been meaning to take out the ignite karaoke as well for a while :)

phrawzty (approved):

PR #14857, opened 2025-02-06
[CHI-2025] Adding Keith as a speaker

caward24:
[CHI-2025] Added Keith Willoughby as a speaker. Updated Corey Ward's picture.


somatorio:
Keith talk isn't in the program, you wanted to add them to the speaker page without a talk, is that it? (asking just to confirm b efore approving and merging)

mattstratton (approved):
Approving this - I'll fast follow with a PR to add Keith to the program

PR #14863, opened 2025-02-10
[site] Remove 2024-johannesburg

phrawzty:
This removes the 2024-ahmedabad event, which was only a stub, and the event did not happen in any case.


phrawzty:
n.b. Distribution list and Slack membership have been adjusted accordingly.

nimbinatus (approved):

PR #14864, opened 2025-02-10
[site] Remove 2024-nova_lima

phrawzty:
This removes the 2024-nova_lima event, which was only a stub, and the event did not happen in any case.


phrawzty:
n.b. Distribution list has been adjusted accordingly.

nimbinatus (approved):

PR #14875, opened 2025-02-12
[SITE] add pretix display event info

toshywoshy:
- add the variable option `info`
- if `info` is show, we display the event info in the widget


yvovandoorn:
LGTM 👍

PR #14876, opened 2025-02-12
[2025-{amsterdam,eindhoven}] Show ticket shop blurb in embeds

issyl0:
- We want this so that prospective buyers know all the details of which ticket types there are and when the event is, etc.


toshywoshy (approved):

toshywoshy:
Normally, you would need to separate PR's, but all is wll

PR #14879, opened 2025-02-14
[AMS-2025] Change Devleaps logo

FloorD:
Changing Devleaps' logo, they're a Dutch company and only ever were involved with Amsterdam (last year) and would like to change the logo for both years.


phrawzty (approved):
Normally we don't rewrite history, etc. but in this case it seems fine.

PR #14903, opened 2025-02-21
[site] Remove 2020 buenos aires

phrawzty:
This removes the 2020 Buenos Aires event, which was only a stub, and the event did not happen in any case.


phrawzty:
n.b. Modified the corresponding Slack membership and mailing list.

toshywoshy (approved):

PR #14902, opened 2025-02-20
[global] add social badges to shortcodes

themightymuppet:
- adds the following social icons as shortcodes to be used on any page:
  - Linkedin
  - Twitter
  - Listserv subscription link
  - slack invitation link
  - mastadon
  - bluesky
  - Youtube
- updates new event template to include social shortcodes

PR #14910, opened 2025-02-22
[global] migrate social badge css to scss

themightymuppet:
- moves the custom css for social badge icons from site.css to custom.scss as discussed in https://github.com/devopsdays/devopsdays-web/pull/14902#issuecomment-2675961507


mattstratton:
This should be handled in #15163

PR #14923, opened 2025-02-28
DES MOINES - 2025 - Call for speakers

jangolano:
Des Moines - 2025 Call for Speakers

### DIFF
--- a/data/events/2025/des-moines/main.yml
+++ b/data/events/2025/des-moines/main.yml
@@ -14,11 +14,11 @@ startdate: 2025-10-02T00:00:00+02:00 # The start date of your event. Leave blank
 enddate: 2025-10-03T00:00:00+02:00 # The end date of your event. Leave blank if you don't have a venue reserved yet.
 
 # Leave CFP dates blank if you don't know yet, or set all three at once.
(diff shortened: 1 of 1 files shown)


toshywoshy (approved):

data/events/2025/des-moines/main.yml
```diff
@@ -39,7 +39,7 @@ nav_elements: # List of pages you want to show up in the navigation of your page
  # - name: location
  # - name: registration
  # - name: program
- # - name: speakers
+  - name: speakers
```
toshywoshy, line 42: I think you do not want to open the speaker page yet

toshywoshy (changes requested):

data/events/2025/des-moines/main.yml
jangolano: Good catch.  I forgot that was the speaker list, not the information for the speakers. 

Thanks,

toshywoshy: Yes, if you want to have a speaker info page, you can set that up in a seperate PR when you're ready

toshywoshy (approved):

PR #14948, opened 2025-03-06
[CHI-2025] Create sponsor-layout.md

albert-cheung:
Adding a sponsor-layout page.

### DIFF
--- a/content/events/2025-dallas/propose.md
+++ b/content/events/2025-dallas/propose.md
@@ -27,9 +27,4 @@ Choosing talks is part art, part science; here are some factors we consider when
 
 <hr>
 
-<strong>How to submit a proposal:</strong> Send an email to [{{< email_organizers >}}] with the following information
-<ol>
-	<li>Type (presentation, panel discussion, ignite)</li>
-	<li>Proposal Title (can be changed later)</li>
-	<li>Description (several sentences explaining what attendees will learn)</li>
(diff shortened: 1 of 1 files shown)


mattstratton (approved):
For context of other folks looking, this is a page we will use but not link to in nav

toshywoshy:
At present the page will only show the content and all the oether events below it, as it does not have a layout file in the theme.
If this Is this the purpose, then you're fine to go.

content/events/2025-chicago/sponsor-layout.md
mattstratton, line 1: Please add this to the top of the file:
```

+++
date = "2016-10-17T08:26:34-05:00"
title = "Sponsor Layout"
type = "event"
description = "Sponsor layout for devopsdays Chicago 2025!"
+++
```

mattstratton (changes requested):

content/events/2025-chicago/sponsor-layout.md
toshywoshy: Yes, that would render a page with the content and the sponsors, which is what you're working towards, I suspect

mattstratton: Probably… This is just going to be a link that the attendees get so they can see the map of the sponsor area, but we should make it look nice anyway :)

PR #14955, opened 2025-03-08
[SITE] add asset link for url and images

toshywoshy:
shortcode to link to the devopsdays asset url
- `file` : create a href link in a new tab
- `image` : embeds an image with alt name on the page


mattstratton (dismissed):
This is all good - I would ask that we update the `REFERENCE.md` to include this shortcode so people know it exists and how to use it

toshywoshy:
Done, added to the `REFERENCE.md`

PR #14959, opened 2025-03-10
[CORE] Add Issy to active list

FloorD:
Does what it says on the tin. Welcome, Issy!


adrianmoisey (approved):
Yay!

data/core.toml
```diff
@@ -8,6 +8,7 @@ active = [
 "Camilla Martins",
 "Laura Santamaria",
 "Sharone Zitzman",
+"Issy Long",
```
issyl0, line 11: (nit) Is there a reason these lists aren't in alphabetical order?

phrawzty: No, that's an error. They should be in alphabetical order by last name, grouped by co-chair / core status.

issyl0: Can we (I?) write a GitHub Action check that parses the TOML according to those stipulations and fixes itself?

phrawzty: imho that's over-engineering the problem. This list changes once a year at most. @FloorD just needs to sort it, that's all. :)

issyl0: I fixed it manually for now because I was already there and spotted some more in the other lists. I'm used to lists that change daily in my job and I'm sick of being the human linter there. 😂

phrawzty (approved):

PR #14961, opened 2025-03-10
Fix hugo deprecation warning & bump version

valioprea:
This PR aims at fixing hugo warnings and errors locally, by updating the version and some deprecated field to the newer variant.

My colleague, @Ovijmek21 helped me with this as well.


mattstratton (changes requested):
There are a few other places the Hugo version needs to be updated:

`netlify.toml`
`hugoserver.ps1`
`hugoserver.sh`
`metadata.json`

PR #14971, opened 2025-03-13
[CHI-2025] Create sponsor-info.md

albert-cheung:
Creating sponsor-info webpage


mattstratton:
Just out of curiosity ... why do we need this when we have the google doc already and a bitly for it? I thought we had decided this year to just use the shared doc and not make a page? Just curious.

albert-cheung:
Based on Matty's comments. This page probably isn't necessary. In 2023, I believe we had both which is why I decided to create this page.

PR #14988, opened 2025-03-17
[CHI-2025] Changing peoplecert logo per their request

Shoeler:
Peoplecert requested a new logo.


toshywoshy (changes requested):
The `peoplecert-before-20250317.yml` file is missing in the `data/sponsors/` folder and the old `peoplecert.png` should be copied to `peoplecert-before-20250317.png`

PR #14999, opened 2025-03-19
[MDE-2025] Adding Event and Sponsors

mpinedag12:
[MDE-2025] Adding Event and Sponsors


assets/events/2025-medellin/logo-square.png
toshywoshy, line 1: This file is an exact copy of the `logo.png` file, so there is now no need for this file anymore

toshywoshy (changes requested):

PR #15058, opened 2025-04-04
[SITE] change to using resize to avoid logo cut

toshywoshy:
see London, Halifax and Istanbul logo on the main page it gets cut off
using resize it should resize the image better
and avoid any cutting off

For ease of understanding adding before and after screenshots
Before/Current : using the fill

After/Improvement : using resize


jerdog (approved):
LGTM

mattstratton (approved):
I think this is fine. There might be some weirdness for logos being squished but I *think* the code allows for a separate logo for the shortcode (for use on welcome page) than for the main screen as an override

PR #15067, opened 2025-04-08
[AMS-2025] v4 of the prospectus

FloorD:
We're on v4 of the prospectus due to deadlines passed and re-prioritizing. Thanks!


issyl0:
Maybe a good idea to remove the DevHops slide too from the prospectus?

FloorD:
> Maybe a good idea to remove the DevHops slide too from the prospectus?

I'll change the pdf, not the link, thanks!

toshywoshy (approved):

PR #15081, opened 2025-04-13
[DEN-2025] Update event details, init MINT sponsor

themightymuppet:
- adds new sponsor "mentorsintech"
- updates denver location


don-code:
Do you additionally want to update the `cfp_link` and `registration_link` in your main.yml, so that these URLs work with the "Register" and "Propose" buttons across the site?

themightymuppet:
> Do you additionally want to update the `cfp_link` and `registration_link` in your main.yml, so that these URLs work with the "Register" and "Propose" buttons across the site?

@don-code it seems like core may not be fully aligned on how this is done? It was asked that we embed registration last time instead of using the link to avoid redirecting offsite. We're fine with how it's configured for now but for these CTAs on the welcome page, we'd rather just have folks direct link to where they need to go so there are less clicks to complete.

content/events/2025-denver/welcome.md
```diff
@@ -4,15 +4,18 @@ Type = "welcome"
 aliases = ["/events/2025-denver/"]
 Description = "devopsdays Denver 2025"
 +++
+<b>DevOpsDays Rockies is returning to Denver in 2025!</b><br><br>
+
+  <h2><a href="https://talks.devopsdays.org/dodrox-2025/cfp" style="font-size: 1em; padding: 5 10px;text-decoration: none; border-radius: 5px;">Submit a proposal!</a></h2>
```
toshywoshy, line 9: It is better to use the `cfp_link` and then use that here using a shortcode ` {{< event_link url-key="cfp_link" text="Propose a talk!" >}}`
Then you remain consistent, but you need to add that to the `main.yml` as @don-code already suggested

toshywoshy (changes requested):

mattstratton:
I think that registration and submission forms don't *have* to be embedded; the support for offsite linking in the code was intentional. 

The desire to avoid offsite links is more for historical reasons i think, but for reg forms and cfp forms I personally think that linking offsite is totally fine. 

Usually what I do is both; just so there is a /propose and a /register page that works, but it's never exposed in nav/cta as those then to direct because that's a better experience IMHO 

That's just my opinion; final say is @nimbinatus

toshywoshy:
> > Do you additionally want to update the `cfp_link` and `registration_link` in your main.yml, so that these URLs work with the "Register" and "Propose" buttons across the site?
> 
> @don-code it seems like core may not be fully aligned on how this is done? It was asked that we embed registration last time instead of using the link to avoid redirecting offsite. We're fine with how it's configured for now but for these CTAs on the welcome page, we'd rather just have folks direct link to where they need to go so there are less clicks to complete.

I added my suggestion for the `cfp_link`.
As for `registration_link` it depends what you are using, and as you are using Pretix, we have a shortcode to embed that easily.

themightymuppet:
Thanks y'all - i've updated this to use the shortlinks - cfp nav now direct links and uses url-key in the shortcode to keep it consistent.

To say my peace - it doesn't feel right to me to block a PR depending on who happens to review and their preference. I really really appreciate everyone jumping in here to make sure this moves forward but I think something like shortlinks vs directlinks wouldn't cause a performance hit to the site so should be up to the city to decide. I understand not everyone knows all their options for managing their page but I'm simply suggesting if it works that we don't block but rather suggest so that they may consider updating it directly or in a following PR.

toshywoshy (approved):

toshywoshy:
@themightymuppet I merged your code.

However I feel the need to comment on your frustration.
At present, most PR have a turn around of less then 24 hours, unless they have organizer changes or core approval requiements or have a true blocker.
And I think we can even say that the first review happens within the first 4 hours of submitting a PR.
This is a great improvement, which has taken time and effort of a lot of people and involves a good team of reviewers working together.
As you see, we are a team, and not all members know all of the code in the same detail, we use different techniques to evaluate a PR and we all have different levels of skills of Hugo and DoD Hugo, and different methods of commenting and providing feedback, therefor you cannot have the same feedback level from each person.
As for your comment on `if it works, don't block`, I am a firm disbeliever of that, your code might work, but next time you update the CfP link in the CTA but forget to change it in the `main.yml`, while "it works", you now have inconsitencies.
If you copy your event for a next edition, and having used shortcodes as we suggest, you now have a single source of truth , this means you can now edit the `main.yml` and all other pages are up to date. Not only does this make your life easier, it makes reviewing easier, meaning we can merge even quicker.

And while I am not saying you're wrong, I am asking you to think with us to improve the situation, we want to review quickly and merge as soon as possible, but we want to be sure of the changes and that they are consistent.
I think we have a lot of improvement on that level of automation and ease-of-use and we can talk about those kind of improvements further on Slack or other mediums of conversations to make life easier, while maintaining a certain level of consistency and traceablity.
Please also remember that we have grown organically over time, and breaking changes require time, effort and enough testing.

themightymuppet:
Thank you @toshywoshy! 

To be clear, i wasn't trying to say reviews were not timely or that the process hasn't been improving. I really appreciate what all of y'all do and the effort you put into making sure each city has the support they need to push these changes forward. My frustration was really around the inconsistencies with reviews as if feels like the PRs are often blocked based on reviewer preference. I get there's a team and everyone has different experiences/techniques but some alignment/communication on review expectations would be greatly appreciated. 

And I totally understand that "if it works, don't block it" is not a great method either by any means - it was a radical suggestion, but I do think there should be a bit more autonomy within cities on how they choose to layout their site and what works best for their audience. I'm happy to help brainstorm more options that can support both sides. I don't want to increase the difficulty of reviewing. Maybe a simple config test script that runs on push to check for common patterns could help. It can auto post a comment in the PR with suggestions for changes? 

Again, very very much appreciate y'all and i'd love to help. I'll cross-post this in slack to keep the conversation going since this PR is closed. Thanks again <3

PR #15280, opened 2025-06-20
[Cluj-Napoca-2025] Starting the website for Cluj-Napoca devopsdays 2025 + adding the organizers

Ovijmek21:
In this PR the event was created and the 4 organizers were added.


phrawzty (approved):
No problem—I'm glad we could get it resolved. :)

PR #15304, opened 2025-07-01
[Detroit-2025] - Canceling

sean-sype-simmons:
Detroit 2025 - Canceling event.


phrawzty (changes requested):
😢 

Please also modify `main.yml` to remove the start, end, and CFP-related dates, and also set sponsors_accepted to no. While you're at it, you may wish to remove all of the sponsor levels as well, to keep things tidy.

And, again: 😢

PR #15306, opened 2025-07-02
Adds CastAI as a gold sponsor for denver 2025

denverdevopsdays:
Adds CastAI as a gold sponsor for denver 2025


phrawzty:
As this is from a new GH user with no history or profile information, I sent a message to the #denver channel on Slack to get confirmation from a human directly.

phrawzty (approved):

PR #15309, opened 2025-07-04
[jf-2025] Frist PR DevOpsDays JF 2025

jgt-josegeraldo:
Frist PR for inclusion of the event on the DevOpsDays JF 2025.


nimbinatus:
:wave: Hi there! There are a number of updates to the 2024 event in this PR. Did something happen that these updates need to happen?

Also, there are a few things missing from this PR. Did you use the [new event script](https://github.com/devopsdays/devopsdays-web/tree/main/utilities#events) to add this event, or did you create it by hand? The immediate thing that jumps out at me is the missing redirects.

jgt-josegeraldo:
Hi @nimbinatus 

In this case, last year's changes and the schedule, which ended up only being published on social media last year, I'm now using the PR to update the page.

I created them manually. Which redirect is missing?

nimbinatus:
Hi @jgt-josegeraldo!

As @phrawzty noted, please make this PR about one event only. We can merge things much faster if you break down the years into different PRs.

The script noted is the best way to create a new event each year. We update it as the files get moved around and change so you don't have to keep up with changes to our process. Among other things, it updates the redirect every year (e.g., if you go to https://devopsdays.org/juiz-de-fora, it goes to 2024 at the moment, but updating the redirect with the script will point it to the newest year each time so you can make things like business cards, posters, QR codes, etc. using one link that you can reuse every year). More documentation on the utilities scripts are at https://github.com/devopsdays/devopsdays-web/blob/main/utilities/README.md, including information on the redirect system.

jgt-josegeraldo:
Hi @nimbinatus @phrawzty

Ok, I created three PRs, one for each year:

- PR 2023: https://github.com/devopsdays/devopsdays-web/pull/15335

- PR 2024: https://github.com/devopsdays/devopsdays-web/pull/15336

- PR 2025: https://github.com/devopsdays/devopsdays-web/pull/15337

can I close this PR?

phrawzty:
@jgt-josegeraldo Sounds good. We'll review those instead.

PR #15344, opened 2025-07-21
[DEN-2025] update sponsors, organizer employer

themightymuppet:
- updates cast ai sponsor logo to their latest version. This used the update logo script to adjust pre-2025 events to use the existing logo.
- updates spacelift logo. This is the same exact logo we already had but it's a better resolution as it's the one directly from their presskit and not a screenshot of their logo. I did not use the update logo script here because it is not a new version of the logo.
- adds spacelift as a sponsor to denver 2025
- updates Tavis' employer from splunk to easypost.


assets/sponsors/c/castai-before-20250721.png
nimbinatus, line 1: @themightymuppet, did you use the change logo script to make this update? It looks like you did, but for some reason the data file for the changed sponsor isn't in here. Mostly wondering if the script is broken...

themightymuppet: @nimbinatus ah I missed that file when I staged it for commit - the script didn't miss it that was user error, sorry!

themightymuppet: updated

phrawzty (approved):

PR #15396, opened 2025-08-06
DSM-2025 Add Location Page

jangolano:
DSM-2025 Add Location Page


content/events/2025-des-moines/location.md
```diff
@@ -1,17 +1,30 @@
...
+Des Moines, IA 50309
+```
+
+## Parking
+
+There is a paid parking lot available near the Iowa Event Center or there are several toll parking garages attached to the skywalk system that are within a few blocks. Many hotels offer a complimentary shuttle service or there is public transit (Des Moines Area Regional Transit aka DART) or Uber/Lyft if you want to take advantage of those and avoid parking fees.
 
 <!-- Uncomment this only if you have set the coordinates for your location in the config yaml. Get Latitude and Longitude of a Point: http://itouchmap.com/latlong.html -->
-<!-- {{< event_map >}} -->
+{{< event_map >}}
```
toshywoshy: This does not really do much as you do not have the location coordinates set up

PR #15446, opened 2025-08-27
[TLV-2025] - Added New Sponsors + Edited Sponsor Page

shar1z:
- New sponsors added - Baseshift, DevOps Leaders IL, Robusta
- Edited sponsor page

### DIFF
--- a/data/events/2025/tel-aviv/main.yml
+++ b/data/events/2025/tel-aviv/main.yml
@@ -189,8 +189,6 @@ sponsors:
     level: gold
   - id: baseshift
     level: community
-  - id: devops_leaders_il
-    level: community
   - id: statscraft
     level: community
  #  url: http://mysponsor.com/?campaign=me # Use this if you need to over-ride a sponsor URL.


shar1z:
Apologies for the many edits - I still am unable to view Hugo locally since upgrading my laptop - there's some kind of error related to templates and other stuff from other events that I can't seem to resolve.
My only way to preview is through the Netlify.

data/sponsors/devops-leaders-il.yml
```diff
@@ -0,0 +1,2 @@
+name: "devops leaders il"
+url: "https://open.spotify.com/show/1D8xJvi6UMANmutW63FoNI?si=d36eb8fa4df14723"
```
mattstratton: This should be `name: devops-leaders-il`

toshywoshy: Euhm, the `name` entry isn't a real problem.
The problem is that the file name is `devops-leaders-il.yml` while the sponsor entry in main.yml is `devops_leaders_il`
You __need___ to chose the same filename as well as entry, so or use underscore `_` or dash `-`
The name is just used for image alt and in case of low bandwidth

toshywoshy: Sorry, another point is that the image file name should be the same as the data file name and entry in the sponsor section
You currently have `devops_leaders_il.jpeg`, so again be consistent in file naming

toshywoshy (changes requested):

shar1z:
Thanks @toshywoshy 
I dropped the ball on this one - I needed to add a new sponsor - and then remembered I still had this PR festering.
Can you check now - I think I aligned all the name conventions for devops_leaders_il

PR #15468, opened 2025-09-08
[DFW-2025] Add Bryan Finster to speakers

baldwinjim:
Add Bryan Finster to speakers


data/events/2025/dallas/main.yml
```diff
@@ -51,6 +51,14 @@ nav_elements: # List of pages you want to show up in the navigation of your page
 
 # These are the same people you have on the mailing list and Slack channel.
 team_members: # Name is the only required field for team members.
+  - name: "Cayla Halliday"
```
toshywoshy: You are adding an additional organizer, and this requires an email to info @ devopsday . org
For speed, I suggest you create a seperate PR, as this may take some time and this PR cannot merge untill Global Core approves this change

toshywoshy (changes requested):

data/events/2025/dallas/main.yml
baldwinjim: Thank you for pointing that out.  I will remove the additional organizer from this PR and submit that request once we are post conference.  New commit incoming!

PR #15495, opened 2025-09-22
[MTL-2025] First speakers

sebagomez:
We have the first batch or confirmed speakers


data/events/2025/montreal/main.yml
```diff
@@ -51,10 +51,10 @@ event_twitter: "devopsdaysmtl" # This will create a traditional "Follow" twitter
 
 nav_elements: # List of pages you want to show up in the navigation of your page.
  # - name: propose
- # - name: location
+  - name: location
  # - name: registration
  # - name: program
```
nimbinatus, line 56: One question: Did you want the program to be available, or are you staging that page for when you're ready to launch it? Right now, it will not appear on the site; you need to manually type in the link to get to it.

nimbinatus (commented):
One question, otherwise LGTM and ready to merge

data/events/2025/montreal/main.yml
sebagomez: I'm still working on it, so no, I don't want it published right now. Thanks for asking :)

sebagomez:
@nimbinatus if everything else is ok, could you approve it so we can have it live asap? Thanks 🙏

PR #15523, opened 2025-10-04
[ZRH-2026] Add CFP Link

romanoroth:
- Added the CFP Link


toshywoshy (approved):
Your title only states cfp link, however you're also making a sponsor change.
For our reviewing please make the title as clear as possible

PR #15528, opened 2025-10-06
[DEN] organizers update

themightymuppet:
per my ping in slack @nimbinatus


data/events/2026/denver/main.yml
```diff
@@ -71,6 +71,13 @@ team_members: # Name is the only required field for team members.
     linkedin: "https://www.linkedin.com/in/rbaccary/"
   - name: "Dale Coyner"
     linkedin: "https://www.linkedin.com/in/dale-coyner/"
+  - name: "Tavis Aitken"
```
toshywoshy, line 80: Adding a new organizers requires an email to info @ devopsdays

themightymuppet: Hey @toshywoshy, this PR is for @nimbinatus -- I pinged her about it on slack. Tavis and Carla are not actively organizing this year but we want to make sure they don't lose access. I was waiting to hear back about if this was the preferred way to go about this or if they would keep access because of their tenure anyway.

themightymuppet: Just sent a new email to info after syncing with Laura.
This is ready for review now -- thanks!

PR #15795, opened 2026-03-31
[ATL-2026] adding teleport, bitrise, and azul

ecology-chris:
adding teleport and bitrise as gold, azul as silver.


josvanschouten (approved):
lgtm

PR #15846, opened 2026-04-20
GE2026 Add linked stats

deniger:
GE2026 Add linked stats


toshywoshy (changes requested):
As this is to send analytics to LinkedIN, we need to make this clear in a privacy policy or notify users about this some where
As I do not find this back anymore, you need to add that to a privacy policy or to your page

PR #15866, opened 2026-04-28
GE2026 change speaker page css

deniger:
GE2026 change speaker page css


phrawzty (approved):
This doesn't work in the preview because of cross-domain rules, but the [live view](https://devopsdays.org/events/2026-geneva/speakers) behaves properly. As this is a sizing change, we can assume it will work as well.

PR #15910, opened 2026-05-18
Nashville-2027 - A simple landing page telling everyone to go to the new site.

technoe:
We're really frustrated with the way the parent org handles the site. So we built a new one that is backed by a database. You know, like a real application would be. That way we can update the sponsors, organizers, volunteers, & schedule in one place and have it propagate to our site and the mobile apps. 
This is a simple landing page telling everyone who visits, to go to the new site.
https://dodnashville.com
Thanks.

### DIFF
--- a/content/events/2027-nashville/welcome.md
+++ b/content/events/2027-nashville/welcome.md
@@ -1,0 +1,17 @@
++++
+Title = "DevOpsDays Nashville 2027"
+Type = "welcome"
+Description = "DevOpsDays + AI Nashville 2027"
++++
+
+<div class="row">
+  <div class="col-md-3">
+    {{< event_logo >}}
+  </div>
+  <div class="col-md-6" style="box-shadow: 0 4px 8px rgba(0,0,0,0.1); margin: 10px; padding: 10px; border-radius: 10px">
+    <h1>DevOpsDays + AI Nashville 2027</h1>
+    <p>Our event information has moved! Visit us at:</p>
+    <h2><a href="https://dodnashville.com" target="_blank">dodnashville.com</a></h2>
+    <p>Schedule, speakers, sponsors, registration, and more are all available on our new site.</p>
+  </div>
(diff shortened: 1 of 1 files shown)


toshywoshy (changes requested):
This will not really work, you may have a city welcome page, but nothing else

toshywoshy:
You're missing a lot of files, data files and some content files.
As for your other comments, I don't consider Firebase as a real database, neither is it open source
As for offsite website, I refer to @nimbinatus or Global Core

technoe:
We don't need anything more than a redirect to the new site. No logos or content. Just hey, go here instead.

jyee:
Hey folks, I've got a meeting with Don later today to discuss, but I suspect this thread is becoming representative of the overall situation. Let's press pause on this until Don and I can chat. I'll provide an update afterwards.

jyee:
@technoe Hi, just wanted to give an update on this. There's no problem with having [dodnashville.com](https://dodnashville.com/) as your main site. I think DOD Nashville has done that in the past and plenty of other events do as well. Also, it's totally ok to have your welcome page point folks over to that site.

However, this PR will need to have a `/data/events/2027/nashville/main.yml` file in order for your welcome page to show up. That's not a Core Team requirement, it's just how Hugo works (you can see in the preview build that nashville 2027 doesnt appear). You can disable the navigation elements in that data file so that it's only the welcome page. When you get dates for the 2027 event, you'll also want to add those to the data file, as this is what controls the event appearing on the homepage. Whether you add other info, like social media links, sponsors, etc. is entirely up to you and the DOD Nashville team. 

The only non-technical requirement we'd like to keep from a Core Team perspective, is the list of team members in that data file. This helps us know who is officially on the team (or not) when we're making changes to your mailing list and taking other administrative actions.

I hope that helps, and let us know if you have questions or need assistance in adding the `main.yml`.

PR #15914, opened 2026-05-20
[PT-2026] Fix images at welcome page and add kcd Porto and DevOps Porto as community sponsors

somatorio:
Note: This includes a new shortcode to the theme

### DIFF
--- a/data/events/2026/portland-or/main.yml
+++ b/data/events/2026/portland-or/main.yml
@@ -104,12 +104,15 @@ organizer_email: "portland-or@devopsdays.org" # Put your organizer email address
 
 # List all of your sponsors here along with what level of sponsorship they have.
 # Check data/sponsors/ to use sponsors already added by others.
-sponsors: []
+sponsors:
 #   - id: samplesponsorname
 #     level: gold
 #  #  url: http://mysponsor.com/?campaign=me # Use this if you need to over-ride a sponsor URL.
 #   - id: arresteddevops
 #     level: community
+  - id: microsoftazure
+    level: snack
+
 
 sponsors_accepted : "yes" # Whether you want "Become a XXX Sponsor!" link
 
@@ -134,10 +137,10 @@ sponsor_levels:
     label: Lanyard
     max: 1
   - id: coffee
-    label: Coffee Break Sponsor
+    label: Coffee Break
     max: 2
   - id: snack
-    label: Snacks Sponsor
+    label: Snacks
     max: 2
   - id: community
     label: Local Community Groups


phrawzty:
Hello @somatorio,

The parts of this PR that touch only the PT-2026 event are reasonable; however, this PR also modifies the _theme_, which is a different request altogether. It will need to go through a more rigorous review by the web team, and to be blunt: that might take a while.

My strong suggestion: make two PRs. One that gets you the visual result you want using the existing theme, and then another that proposes the theme modification for RFC.

Let me know how you want to proceed.

somatorio:
To make that change without the shortcode I would have to duplicate the kcd-porto logo, having it at the assets and a copy at static :/

Unless there is another way and @toshywoshy can point me towards it

toshywoshy (approved):

themes/devopsdays-theme/layouts/shortcodes/img.html
```diff
@@ -0,0 +1,8 @@
+{{ $image := resources.Get (.Get "src") }}
+{{ $class := .Get "class" }}
+{{ $alt := .Get "alt" }}
+{{ if $image }}
+<img src="{{ $image.RelPermalink }}" class="{{ $class }}" alt="{{ $alt }}">
+{{ else }}
```
toshywoshy: After some thought, I am not in favour of having this `else`, we shouldn't have the error end up on the front page, I prefer to fail silently, the person creating the PR or using the shortcode should have the reflex to correct that as their image would not be visible.

toshywoshy (changes requested):

themes/devopsdays-theme/layouts/shortcodes/img.html
somatorio: Oops, that was a test I did and forgot to remove :O
I'll fix it

somatorio:
Just noticed I could add "style" to the shortcode as well as this could be useful, don't know what you folks think of that though

toshywoshy:
I prefer not to add a styling to the image block, as it will inhert that depending on where you use it.
We cannot predict exactly where the shortcode will be used

content/events/2026-portugal/welcome.md
```diff
@@ -32,8 +32,8 @@ Description = "devopsdays Portugal 2026| Collab with KCD Porto 2026"
 
   <div class="row justify-content-center">
     <div class="col-sm-2">
-      <a href="https://community2.cncf.io/events/details/cncf-kcd-porto-presents-kcd-porto-2026-collab-with-devops-days-portugal"><img src="assets/sponsors/k/kcd-porto-logo.png" class="img-fluid mx-auto" /></a>
-      <a href="https://devopsporto.com/"><img src="/img/sponsors/devopsporto.png" class="img-fluid mx-auto" /></a>
+      <a href="https://community2.cncf.io/events/details/cncf-kcd-porto-presents-kcd-porto-2026-collab-with-devops-days-portugal">{{< img src="/sponsors/k/kcd-porto.png" class="img-fluid mx-auto" alt="Website KCD Porto" >}}</a>
+      <a href="https://devopsporto.com/"><img src="/img/sponsors/devopsporto.png" class="img-fluid mx-auto" alt="Website DevOps Porto"/></a>
```
nimbinatus: If you're updating the prior line to change to the shortcode, why not here as well?

themes/devopsdays-theme/reference.md
```diff
@@ -1,5 +1,7 @@
 # Reference for devopsdays-theme
+
```
nimbinatus: For the future, I get that this probably was auto-cleaned up, but adding all of the lines and changing the table widths on the markdown that doesn't provide substantive change makes PRs like this really hard to get reviewed quickly. And the next time someone else makes a change, if they have different auto-clean settings on their IDE, this will get changed again. Recommend turning those off in the future for repos where a lot of people with different IDE settings are working.

themes/devopsdays-theme/reference.md
```diff
@@ -132,9 +137,10 @@ nav_elements
   - name: party
     url: http://www.google.com
 ```
+
 The above example would create a new menu item called "Volunteer" which linked to `devopsdays/events/YYYY-CITY/volunteer`, and another menu item called "party" which would link to `http://www.google.com`
 
-The menu items also take an optional parameter of `icon` where you can set the font-awesome icon that will display on small screens. *Note: This feature is currently deprecated, but it won't break anything if you use this setting* Choose at http://fontawesome.io/icons/. Example:
+The menu items also take an optional parameter of `icon` where you can set the font-awesome icon that will display on small screens. _Note: This feature is currently deprecated, but it won't break anything if you use this setting_ Choose at http://fontawesome.io/icons/. Example:
```
nimbinatus: Why was this changed?

themes/devopsdays-theme/layouts/shortcodes/img.html
```diff
@@ -0,0 +1,6 @@
+{{ $image := resources.Get (.Get "src") }}
+{{ $class := .Get "class" }}
```
nimbinatus: I'd recommend this being optional, not required. Not everyone uses CSS classes in this repo.

themes/devopsdays-theme/reference.md
```diff
@@ -463,3 +494,10 @@ This shorcode requires the city name `city`, the year `year`, the name `name` be
 {{< asset year="2025" city="chicago" name="map layout" image="map.png" >}}
 ```
 
+### img
+
+Create a img html tag with an image from the assets
```
nimbinatus:
```suggestion
Create an `img` html tag with an image from the assets directory.
This shortcode requires the image source filepath as `src`, any CSS classes you want to add `class`, and alt text for accessibility `alt`.
```

Please add more use information. Also, please update this when you make the `class` parameter optional.

nimbinatus (changes requested):
The main big change I'd like to see here is making the `class` parameter optional and adding better documentation. The rest is primarily notes.

somatorio:
Oops, did the merge by mistake :O
I'll do the changes requested and make a new PR

Edit: Hmmm, looks like by luck the PR was emptied before merging, so no changes were actually made... I'll create 2 PRs, one adding the theme change and one using it after the first one is merged

themes/devopsdays-theme/reference.md
somatorio: Weird that I don't remember making this change 🤔

PR #15925, opened 2026-05-25
[BER-2026] Cancellation of event

range:
We are cancelling the event because of missing sponsors. I've worked through the complete `cancel-event.md` and shouldn't have missed anything. 

See you another year.


toshywoshy (approved):
I regret to see this happen

PR #15927, opened 2026-05-26
[PDX-2026] add current year

binford2k:
Adding initial website for PDX 2026


jyee:
Note, Jeremy Adams has already been added to slack and the email list (admin work was done prior, but he hadnt been added to the site)

toshywoshy (approved):

jyee:
Do not merge.
It looks like there was a typo in #15903 which means that change doesnt actually do anything. That's causing the table in this PR to fail. But this issue needs more digging as other events have markdown tables that render without the config change (e.g. see Geneva's sponsorship table in markdown).

I'm investigating why theirs works and if it's reliably reproduceable. If so, then I'll get this working and file a PR to revert the config change.

binford2k:
They added [custom css directly to the page](https://github.com/devopsdays/devopsdays-web/blob/main/content/events/2026-geneva/sponsor.md?plain=1#L6-L38) which defeats the purpose of consistency.

binford2k:
I'm not going to be pedantic about it, but the point of attaching styles to elements referencing a stylesheet is consistency and maintainability. Embedding css directly in a page is not "cleaning up," it's actually the opposite and is generally considered to be a last resort hack.

[We have a theme](https://github.com/devopsdays/devopsdays-web/tree/main/themes/devopsdays-theme/assets/scss), we should be using it instead of hacking together overrides all over.

PR #15937, opened 2026-06-02
[core] remove stale wording from rule

phrawzty:
This removes the reference to ConferenceOps from the fourth rule. It is materially the case that we can receive donations via e.g. the foundation, which is in fact laid out later in this very same document.


issyl0 (approved):

mattstratton (approved):
I think this looks good, as the caveat was true at the time but it is not true now

jyee (approved):
LGTM

PR #15949, opened 2026-06-15
[DALLAS-2026] Add Teleport as silver sponsor

sundeepbobba1985:
Add Teleport as silver sponsor for DevOps Days 2026


phrawzty (approved):
The build failed but it has nothing to do with the PR, which is straightforward.

PR #15950, opened 2026-06-15
Feat: full_label field can be used instead of the label

### LINKED ISSUE #14456: Add ability to override sponsor section titles
Connected to #14440

Today, when you set a sponsor level, that word is put in front of "sponsor(s)" and all the other places that sponsorship is referred to (i.e., if  the level if "Silver", it will be referred to as "Silver Sponsors"). A case has come up for sponsorship levels which the organizers would prefer to not include the word "sponsor".

I would suggest that this could be implemented with an optional override on the title for a sponsor level (similar to URL overrides on the navigation):

```yaml
  - name: propose
    url: https://talks.devopsdays.org/devopsdays-chicago-2025/cfp
```

so something like this:

```yaml
sponsor_levels:
  - id: platinum
    label: Platinum
    max: 2
  - id: gold
    label: Gold
  - id: academic
    display: Academic Supporters
```

It would be confusing slightly, because it's like "Label" is what is used if you want "...Sponsor(s)" appended, and "display" is the override. Maybe it would be called `display_override` to be clear?

ivellios:
# Background 

Using the `label` field implies that in the template of the event, each section of the sponsors will show up as `[label] Sponsors` (with the _Sponsors_ suffix). Some events may want/need to add different types of partners that are not explicitly _Sponsors_, and adding an automatic suffix of Sponsors makes them look awkward.

# Proposal

To avoid that, this commit introduces a new field `full_label` which - when set - will be used in place of the `[label] Sponsors` (replacing it fully).

The same rule applies to the _Join as..._ link in the section. The `label` field is not required if `full_label` is used instead. 

**Backwards compatibility**

The default still falls into the `label` field to maintain old events.

# Example

image

Closes #14456

### DIFF
--- a/themes/devopsdays-theme/layouts/partials/sponsors.html
+++ b/themes/devopsdays-theme/layouts/partials/sponsors.html
@@ -35,7 +35,9 @@
       <div class="row cta-row">
           <!--first sponsor row-->
           <div class="col-md-12">
-            {{- if eq $level.max 1 -}}
+            {{- if $level.full_label -}}
+              <h4 class="sponsor-cta">{{ $level.full_label }}</h4>
+            {{- else if eq $level.max 1 -}}
               <h4 class="sponsor-cta">{{ $level.label }} Sponsor</h4>
             {{- else -}}
               <h4 class="sponsor-cta">{{ $level.label }} Sponsors</h4>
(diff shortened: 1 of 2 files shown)


nimbinatus (approved):
Looks fine to me

nimbinatus (changes requested):
Actually, there needs to be changes to the `main.yaml` file to show how it's done: https://github.com/devopsdays/devopsdays-web/blob/main/utilities/examples/data/events/main.yml

Specifically, you need to update the docs and show an example here: https://github.com/devopsdays/devopsdays-web/blob/2d984dcb1c5d311f7c9893c13f840ada78f23241/utilities/examples/data/events/main.yml#L93-L107

ivellios:
Very good point @nimbinatus - forgot about the example file! Thank you! 
Now, it is also documented there.

PR #15959, opened 2026-06-18
Add support for modern social share platforms.

binford2k:
The upstream library was abandoned, but it still works with some
tweaking. I'm not truly convinced of the utility of share icons
anymore, but if they're there then I want to use the modern platforms.

### DIFF
--- a/themes/devopsdays-theme/layouts/partials/footer_scripts.html
+++ b/themes/devopsdays-theme/layouts/partials/footer_scripts.html
@@ -76,6 +76,12 @@
     {{- $share_str = "\"facebook\"" -}}
   {{- else if eq $current_item "linkedin" -}}
     {{- $share_str = "\"linkedin\"" -}}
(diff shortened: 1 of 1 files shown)


mattstratton:
> The upstream library was abandoned, but it still works with some
tweaking

Which is why I never got around to doing what you did, so thanks! :)

mattstratton:
I don't think these new icons or share buttons get used anywhere?

I can take a look and figure out because I think the event main page needs to know about them

mattstratton:
OK, looking at it. It should just work, but when I look in the deploy preview, I don't see the new buttons.

binford2k:
I did not change the default set. I figured the safer approach would be to make them available first so events could opt in to using them. Then at some point in the future we could rethink the default set of icons.

mattstratton:
cool, just wanted to make sure I tested it properly! lgtm

mattstratton (approved):

PR #15972, opened 2026-06-29
[contrib] add docker-compose.yml

phrawzty:
Just adding a `docker-compose.yml` because it can be useful. :)

### DIFF
--- a/makefile
+++ b/makefile
@@ -1,4 +1,7 @@
-.PHONY: deps build server clean
+.PHONY: deps build server clean docker-server
+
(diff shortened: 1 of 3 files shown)


mattstratton:
I would just do a quick pass through any of the documentation to see if this needs to be added to that so people know it exists. Also I think there's some note in a doc somewhere where we talk about all the files you have to change when you update the Hugo version. This would be one of those files

makefile
phrawzty, line 1: I also stripped the Windows-style line endings from this file, which is why every line has been modified.

mattstratton (approved):

PR #16051, opened 2026-08-01
[PDX-2026] Add New Stack as a media sponsor for DoDPDX

jberkus:
Adding The New Stack as a media sponsor.

Attn: @jyee


jyee (changes requested):
Adding a new sponsor file and logo is unneccessary. The New Stack already exists as `thenewstack.yml`

jberkus:
Ah, I didn't think to look under "T".

jyee (approved):

PR #16092, opened 2026-08-26
[HOU-2027] First commit for the next Houston event in 2027

skryukova:
Initial commit for the Houston 2027 event, to set up organizer list.


data/events/2027/houston/main.yml
```diff
@@ -0,0 +1,138 @@
...
+  - name: sponsor
+  - name: contact
+  - name: conduct
+# - name: example
+#   icon: "map-o" # This is a font-awesome icon that will display on small screens. Choose at http://fontawesome.io/icons/
+#   url: http://mycfp.com # The url setting is optional, and only if you want the navigation to link off-site
+
+
+# These are the same people you have on the mailing list and Slack channel.
+team_members: # Name is the only required field for team members.
```
nimbinatus, line 74: We seem to be missing some of these folks on the mailing list for Houston, and there are emails on the list that don't seem to map here. Would you send an email to info@ with a list of names and emails for the current organizers so we can update your mailing list, please?

skryukova: I did already, can resend again, in a minute

skryukova: Resent

nimbinatus: Ah, found it! Thank you; updating now and will merge in a moment.

nimbinatus (approved):

PR #16099, opened 2026-08-30
[Barcelona-2026] Modernize welcome page and trim social share buttons

edwardep:
Reworks the five DevOpsDays Barcelona 2026 event pages onto one visual design. All changes are scoped to this event: no theme files are touched, and the only new file is the event's own stylesheet. Rebased on `main` after #16101, so the program, speakers and registration pages that PR added are untouched here.

## One stylesheet instead of five

The welcome page had grown ~300 lines of inline `<style>`. That moves to `static/events/2026-barcelona/main.css` and every page links it, the same pattern [2023-caceres](https://github.com/devopsdays/devopsdays-web/tree/main/static/events/2023-caceres) and [2024-caceres](https://github.com/devopsdays/devopsdays-web/tree/main/static/events/2024-caceres) already use. The pages then share one set of components rather than four copies drifting apart.

## Welcome

- **Social shares.** `social_shares: ["email", "linkedin"]` — the event has no Twitter or Facebook presence, so those buttons pointed nowhere. Added an `event_social_linkedin` badge to the body: the key was configured but never rendered, so there was no actual link to the event's LinkedIn page.
- **Program, speakers and tickets get a card strip.** #16101 opened registration and published the program and speakers, and added a row for each to the facts list. Those three are now the reason to be on the page, so they get linked cards between the intro and "What's in Store?" instead of a line of small print each. The facts list goes back to facts — dates, location, contact — so no link appears twice.
- **Copy.** The closing paragraph still said to watch for news on speaking and volunteering while the CFP was open; it now points at the program and the speakers, and the call to action leads with the ticket.
- **Layout.** The fixed-width two-column `<table>` becomes a flex layout that stacks on narrow screens; the "What's in Store?" list becomes cards.
- **Colours.** Recoloured from the event logo's palette (navy, cyan, amber, crimson), replacing the theme default blue and an unrelated purple. The masthead gradient previously ended in light grey under white text, leaving the venue address hard to read; the new one holds at least 5:1 throughout.
- **Accessibility.** The theme sets `outline: none` on `:focus`, so tabbing through the new cards and buttons moved an invisible cursor — both take a `:focus-visible` ring, and the card hover lift is dropped under `prefers-reduced-motion`.

## Sponsor

Dropped its own purple palette — it was the only page not using the logo colours — and reused the shared cards, panel and call to action. Its four filled stat blocks mixed two attendance figures with the date and the venue; the figures stay as two tiles, and the date and venue move into the opening line where a reader looks for them. Copy is otherwise unchanged.

## Conduct

Was unstyled markdown. The code of conduct text is unchanged, with three changes: the reporting paragraphs are wrapped in a call-out where they already stood (the only crimson on the page, since "how do I report this" is what someone needs to find fast), the organizers' address is added to it, and the redundant "Conference Code of Conduct" `<h2>` is dropped because the theme already prints the page title as an `<h1>`. Nothing is reworded and no paragraph is reordered: a word-level diff of the prose is 373 words to 381, the difference being that heading removed, the call-out's "If You Need Help" heading, and the address line.

## Location

Was unstyled markdown. Venue, address, neighbourhood and dates now lead as a facts block, followed by the existing prose, with the map filling the column instead of sitting in a 450×250 box.

## Contact

The organizer cards come from the theme's `list_organizers` shortcode, so the markup is untouched — only the surface it lands on:

- The portrait moves to the top of the card. The shortcode prints the name first, so a two-line name pushed that card's photo below its neighbours'; the card is a flex column, so `order` fixes it without touching the shortcode.
- Portraits are cropped square. The theme pins them to `height: 300px`, which showed a different amount of each person depending on their photo.
- Three cards per row instead of four at `≥1200px`. `col-xl-3` inside the site's 1000px measure left ~220px a card, too narrow for a bio.
- "Organizer" was an `<h4>` outshouting the name it belongs to; it is a label now. The employer wears Bootstrap's `.text-muted` (`#6c757d`), too light to read at that size. The social icons come down from `fa-2x`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


edwardep:
Hi @nimbinatus , this PR is also ready for review.

content/events/2026-barcelona/welcome.md
```diff
@@ -70,14 +36,10 @@ Description = "DevOpsDays Barcelona 2026"
...
 
-<div class = "row">
-  <div class = "col-md-2">
-    <strong>Register</strong>
-  </div>
-  <div class = "col-md-8">
-    {{< event_link page="registration" text="Register to attend the conference!" >}}
-  </div>
-</div>
+<!-- Register, Program, Speakers and Sponsors were rows here. They are things
```
nimbinatus, line 39: I'm guessing Claude generated a lot of this? We really don't need comments like this to explain why you changed your welcome page in the future :)

content/events/2026-barcelona/welcome.md
```diff
@@ -88,84 +50,99 @@ Description = "DevOpsDays Barcelona 2026"
...
 
+<!---
 <div class = "row">
   <div class = "col-md-2">
-    <strong>Program</strong>
+    <strong>Newsletter</strong>
   </div>
   <div class = "col-md-8">
-    View the {{< event_link page="program" text="program." >}}
+    {{< event_link page="newsletter" text="Subscribe to our newsletter for the next news !" >}}
```
nimbinatus, line 59: There's no newsletter page in your current build, so I'd recommend removing it completely versus leaving it commented out so that someone else following you doesn't accidentally uncomment it and break the build.

static/events/2026-barcelona/main.css
```diff
@@ -0,0 +1,547 @@
+/* Palette sampled from the DevOpsDays Barcelona logo:
+   navy #111224 · cyan #00BAEB · amber #F9CD34 · crimson #F11647 · cream #FFFCF0 */
+
+/* Masthead: navy instead of the theme default blue-to-grey, so white text
```
nimbinatus, line 4: Side note: Due to the load this puts on the reviewers, we really don't recommend changing the masthead and all of the buttons like this as it slows down reviews immensely (fewer people know CSS on core's review team, so you're throttling your review). I'm merging this as I don't see anything that screams at me on a fast pass, but I probably be the only one who can help you troubleshoot at the moment if something gets wonky when we do template updates.

static/events/2026-barcelona/main.css
```diff
@@ -0,0 +1,547 @@
...
+  overflow-x: hidden;
+}
+.welcome-page-masthead,
+.welcome-page-masthead a {
+  color: #fff;
+}
+.welcome-page-date {
+  font-weight: 700;
+}
+/* Bigger title, now that the masthead carries only the name and the dates.
```
nimbinatus, line 49: Also, highly recommend stripping out Claude's overly verbose commentary, as a general note :)

nimbinatus (approved):
I'm going to ship this, but as a note that I also shared with one of your colleagues: Generally, we discourage deviating from the template as it significantly slows down or throttles reviews. I get that the current page isn't very modern looking (yes, core is working on that), and I'm aware that Claude has a tendency to make the site look exactly like this with the rounded boxes to make it modern (I think it's an assumption it makes with all Hugo sites, but maybe it's all sites?), so if you're using Claude to run updates, it's pretty common. However, perhaps if you still feel you need to do this, just split out the content updates from the style updates so those updates happen faster since more of core's review team can help you.

PR #16107, opened 2026-09-01
[LON-2026] fix puppet sponsor

andyburgin:
add puppet to gold


issyl0 (approved):

issyl0:
My bad for not noticing the banner and the PR differences!

PR #16129, opened 2026-09-08
[Philadelphia-2026] added program and speakers

hlmuludiang:
- Added conference program
- Added basic speakers
- Fixed templating issue with program concatenating speaker names when viewing website locally and updating speaker info


phrawzty (approved):
Basic but functional

PR #16136, opened 2026-09-10
[BHZ-2026] Changing to BHZ-2027

tutagomes:
Postponing the DevOpsDays BHZ-2026 to 2027.


phrawzty (approved):
It's a straight conversion with no changes to the local organiser list, so this seems fine. :)

PR #16143, opened 2026-09-14
[BOS-2026] Boston 2026 speaker lineup

don-code:
In theory, this is the finalized schedule for this year! I've left some breadcrumbs in place, in case it's not.


phrawzty (approved):
🚀

